PR #5, opened 2020-07-23
Enable running consoles in ECS Fargate

bilbof:
This provides the infrastructure necessary to get a developer console up in a container in ECS Fargate.

For example a rails console can be started for the frontend app, as demonstrated in this PR.

See the associated proposal https://github.com/alphagov/govuk-replatforming-discovery-2020/pull/7.

There's [a demo commit](https://github.com/alphagov/frontend/compare/bilbof/ecs-console-dockerfile) on the frontend app repo to show what an app console image might look like.

### DIFF
--- a/terraform/modules/fargate-console/README.md
+++ b/terraform/modules/fargate-console/README.md
@@ -1,0 +1,37 @@
+# Fargate Console
+
+This module provides a way to start a bash session in a running ECS container,
+with the required application code, secrets, and environment variables for
+developers to perform interactive operational and investigative tasks.
+
+Use cases includes:
+
+* Running a Rails console in an AWS environment
+* Running a database console
+* Running a web server without having traffic routed to that server.
+
+See proposal 003 in govuk-replatforming-discovery-2020 for further details
+of the tooling required to make this a good developer experience.
+
+## How it works
+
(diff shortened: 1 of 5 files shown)


terraform/modules/fargate-console/README.md
```diff
@@ -0,0 +1,46 @@
...
+--network-configuration '{
+  "awsvpcConfiguration": {
+    "subnets": ["subnet-6dc4370b", "subnet-463bfd0e", "subnet-bfecd0e4"],
+    "securityGroups": ["sg-0b873470482f6232d"],
+    "assignPublicIp": "DISABLED"
+  }
+}'
+```
+
+To SSH onto the container, we must run `ssh -J <username>@<jumbox-url> root@<container-ip>`.
```
issyl0, line 35: The companion PR to this has changed a bit since this was written, notably to use AWS SSM instead of raw SSH. That says:

> The tool will poll AWS ECS until the container has the status `RUNNING`, and
then will run a command such as `aws ssm start-session --target <container>`.

bilbof: Thanks, I've removed that and pointed to the proposal.

PR #7, opened 2020-08-17
CI: Use linters to check branches before merge

bilbof:
This should ensure Terraform and JSON files are valid before merging.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,16 @@
+name: linting
+on: [push, pull_request]
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: terraform fmt
+        run: |
+          brew install tfenv
+          tfenv install
+          find . -name '*.tf' | xargs tools/terraform-format.sh
+
+      - name: JSON Check
+        run: find . -name '*.json' | xargs tools/json-check.sh
--- a/tools/json-check.sh
+++ b/tools/json-check.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+set -eu
+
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,18 @@
+name: linting
+on: [push, pull_request]
+jobs:
+  test:
+    runs-on: macos-latest
```
issyl0, line 5: You can use `ubuntu-latest` here, as the Ubuntu Actions runners now ship with Homebrew on Linux.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,18 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: terraform fmt
+        run: |
+          brew install tfenv
+          tfenv install
+          tfenv install 0.13.0-beta3
+          tfenv use 0.13.0-beta3
```
issyl0, line 14:
```suggestion
```

`tfenv install` pulls from the repo's `.terraform-version` which should be in the repository root (or per-project if it needs to be different).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,18 @@
+name: linting
+on: [push, pull_request]
+jobs:
+  test:
+    runs-on: macos-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: terraform fmt
+        run: |
```
issyl0, line 10:
```suggestion
        working-directory: ./terraform/
        run: |
```

The `.terraform-version` file is in `terraform`, so if we `cd` into there.

bilbof: I'm not sure it's necessary if we have a `.terraform-version` at the root

PR #9, opened 2020-08-25
Add publisher to terraform

MahmudH:
The publisher app is a backend application, which relies on a database and Redis. 

Two load balancers are added to publisher, one internal ALB, and one public ALB.  Internal load balancers can only route requests from clients with access to the VPC for the load balancer. Public ALB's (Internet-facing load balancers) can route requests from clients over the Internet, (they have public IP addresses). 

The publisher app has been added to the security groups of Redis and DocumentDB in the inbound request section. 

The app is also added to Route53 (DNS) for both internal and public domains. 

Trello card: https://trello.com/c/SdsTMMGL/159-bring-up-publisher-in-fargate


terraform/publisher/variables.tf
```diff
@@ -0,0 +1,63 @@
...
+}
+
+variable "private_subnets" {
+  description = "The subnet ids for govuk_private_a, govuk_private_b, and govuk_private_c"
+  type        = list
+  default     = ["subnet-6dc4370b", "subnet-463bfd0e", "subnet-bfecd0e4"]
+}
+
+variable "public_subnets" {
+  description = "The subnet ids for govuk_private_a, govuk_private_b, and govuk_private_c"
```
seb-szy, line 18:
```suggestion
  description = "The subnet ids for govuk_public_a, govuk_public_b, and govuk_public_c"
```

terraform/task-definitions/publisher.json
```diff
@@ -0,0 +1,114 @@
...
+      { "name": "EMAIL_GROUP_CITIZEN", "value": "test-address@digital.cabinet-office.gov.uk" },
+      { "name": "EMAIL_GROUP_DEV", "value": "test-address@digital.cabinet-office.gov.uk" },
+      { "name": "EMAIL_GROUP_FORCE_PUBLISH_ALERTS", "value": "test-address@digital.cabinet-office.gov.uk" },
+      { "name": "FACT_CHECK_SUBJECT_PREFIX", "value": "dev"},
+      { "name": "FACT_CHECK_USERNAME", "value": "govuk-fact-check-test@digital.cabinet-office.gov.uk"},
+      { "name": "GOVUK_APP_DOMAIN", "value": "test.govuk-internal.digital" },
+      { "name": "GOVUK_APP_NAME", "value": "publisher"},
+      { "name": "GOVUK_APP_TYPE", "value": "rack"},
+      { "name": "GOVUK_STATSD_PREFIX", "value": "fargate" },
+      { "name": "GOVUK_ASSET_ROOT", "value": "https://assets.digital.cabinet-office.gov.uk" },
```
seb-szy, line 19:
```suggestion
      { "name": "GOVUK_ASSET_ROOT", "value": "https://assets.publishing.service.gov.uk" },
```

seb-szy, line 19: should this be something like: https://assets.test.publishing.service.gov.uk/

terraform/task-definitions/publisher.json
```diff
@@ -0,0 +1,114 @@
...
+      { "name": "GOVUK_GROUP", "value": "deploy"},
+      { "name": "GOVUK_USER", "value": "deploy"},
+      { "name": "GOVUK_WEBSITE_ROOT", "value": "https://www.gov.uk" },
+      { "name": "PLEK_SERVICE_CONTENT_STORE_URI", "value": "https://www.gov.uk/api" },
+      { "name": "PLEK_SERVICE_STATIC_URI", "value": "https://assets.publishing.service.gov.uk" },
```
seb-szy, line 24: should this be something like: https://assets.test.publishing.service.gov.uk/

terraform/publisher/main.tf
```diff
@@ -0,0 +1,350 @@
...
+  capacity_providers = ["FARGATE"]
+}
+
+resource "aws_ecs_task_definition" "service" {
+  family                   = var.service_name
+  requires_compatibilities = ["FARGATE"]
+  container_definitions    = file("../task-definitions/publisher.json")
+  network_mode             = "awsvpc"
+  cpu                      = 2048
+  memory                   = 4096
```
seb-szy, line 47: Should we try stick to the application sizes we have estimated within ESC Fargate cost calculator doc? It would be good to see if these estimates are sensible.

terraform/publisher/main.tf
```diff
@@ -0,0 +1,350 @@
...
+  type    = "CNAME"
+  records = ["${var.service_name}.pink.${var.internal_domain_name}"]
+  ttl     = "300"
+}
+
+
+resource "aws_route53_record" "internal_service_record" {
+  zone_id = data.aws_route53_zone.internal.zone_id
+  name    = "${var.service_name}.pink.${var.internal_domain_name}"
+  type    = "A"
```
sengi, line 330: We'll want to avoid bringing the legacy "stackname" baggage (e.g. `pink` here) into the new system. Don't worry about it for now - it's ok for experimental purposes - but we'll need to come up with a way of switching the services over without needing to re-create the stackname stuff.

terraform/publisher/variables.tf
```diff
@@ -0,0 +1,63 @@
...
+}
+
+variable "redis_security_group_id" {
+  description = "The redis security group id"
+  type        = string
+  default     = "sg-0a8a92451bfde8aa3"
+}
+
+variable "documentdb_security_group_id" {
+  description = "The documentdb security group id"
```
sengi, line 48: Please omit redundant descriptions - for example this one doesn't provide any information that isn't already in the name. Similarly the redis one above.

terraform/publisher/variables.tf
```diff
@@ -0,0 +1,63 @@
...
+  type        = string
+  default     = "sg-08a20d332cbc59c3c"
+}
+
+variable "internal_domain_name" {
+  description = "The internal root root domain name"
```
sengi, line 54:
```suggestion
  description = "The domain name suffix for internal services"
```

terraform/publisher/variables.tf
```diff
@@ -0,0 +1,63 @@
...
+  type        = string
+  default     = "test.govuk-internal.digital"
+}
+
+variable "public_domain_name" {
+  description = "The internal root root domain name"
```
sengi, line 60:
```suggestion
  description = "The domain name suffix for Internet-facing services"
```

terraform/task-definitions/publisher.json
sengi, line 19: Probably better to use a non-prod address here (maybe assets.test.publishing.service.gov.uk?), even though it's not going to modify any assets either way.

terraform/task-definitions/publisher.json
sengi, line 24: Similarly, better to use non-prod unless you're really sure.

terraform/task-definitions/publisher.json
```diff
@@ -0,0 +1,114 @@
...
+      { "name": "RAILS_ENV", "value": "production" },
+      { "name": "RAILS_SERVE_STATIC_FILES", "value": "true" },
+      { "name": "REDIS_HOST", "value": "pink-backend-redis.0f3erf.ng.0001.euw1.cache.amazonaws.com"},
```
sengi, line 27: Is there a CNAME for this that we could use so that it won't change when you re-create the Elasticache instance? If there isn't yet then don't worry.

terraform/task-definitions/publisher.json
```diff
@@ -0,0 +1,114 @@
...
+        "valueFrom": "arn:aws:secretsmanager:eu-west-1:430354129336:secret:publisher_app-FACT_CHECK_PASSWORD-A2UM6N"
+      },
+      {
+        "name": "FACT_CHECK_REPLY_TO_ADDRESS",
+        "valueFrom": "arn:aws:secretsmanager:eu-west-1:430354129336:secret:publisher_app-FACT_CHECK_REPLY_TO_ADDRESS-zZ9I9x"
+      },
+      {
+        "name": "FACT_CHECK_REPLY_TO_ID",
+        "valueFrom": "arn:aws:secretsmanager:eu-west-1:430354129336:secret:publisher_app-FACT_CHECK_REPLY_TO_ID-c725KL"
+      },
```
sengi, line 63: I don't think these two need to be secrets (FACT_CHECK_REPLY_TO_ADDRESS and FACT_CHECK_REPLY_TO_ID). They're currently [published in alphagov/govuk-puppet](https://github.com/alphagov/govuk-puppet/blob/502262ba5e81287d361a84281c3386801c45d6df/hieradata/production.yaml#L55).

terraform/task-definitions/publisher.json
```diff
@@ -0,0 +1,114 @@
...
+      {
+        "name": "GOVUK_NOTIFY_API_KEY",
+        "valueFrom": "arn:aws:secretsmanager:eu-west-1:430354129336:secret:publisher_app-GOVUK_NOTIFY_API_KEY-jYkPCM"
+      },
+      {
+        "name": "GOVUK_NOTIFY_TEMPLATE_ID",
+        "valueFrom": "arn:aws:secretsmanager:eu-west-1:430354129336:secret:publisher_app-GOVUK_NOTIFY_TEMPLATE_ID-SP4yN4"
+      },
```
sengi, line 71: This one (GOVUK_NOTIFY_TEMPLATE_ID) also isn't secret.

sengi (approved):
Left a few comments to consider but feel free to merge and iterate.

sengi:
And thanks for adding this btw! This really helps us towards being able to experiment with service discovery and so on.

terraform/modules/infra-fargate/main.tf
```diff
@@ -29,6 +29,12 @@ resource "aws_lb_target_group" "lb_tg" {
   protocol    = "HTTP"
   vpc_id      = data.aws_vpc.vpc.id
   target_type = "ip"
+
+  health_check {
+    path     = "/"
+    timeout  = 30
+    interval = 60
```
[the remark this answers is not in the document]
MahmudH: The reason for overriding the timeout and interval is due to the app starting up processes if they're both not set then the health check fails and the container is drained.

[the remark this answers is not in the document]
sengi: Ah I see, thanks. Don't worry about this for now then. Later on we can look at setting the startup grace period later (`health_check_grace_period_seconds` on the aws_ecs_service resource) which should fix this without needing to mess with all the other timeouts.

terraform/publisher/main.tf
```diff
@@ -0,0 +1,350 @@
...
+  protocol    = "HTTP"
+  vpc_id      = data.aws_vpc.vpc.id
+  target_type = "ip"
+
+  health_check {
+    path                = "/healthcheck"
+    timeout             = 120
+    interval            = 300
+    unhealthy_threshold = 10
+    healthy_threshold   = 2
```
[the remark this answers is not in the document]
sengi: (same as the other healthcheck one above - setting the grace period will take care of the startup issue but don't worry about this for now)

terraform/modules/infra-fargate/main.tf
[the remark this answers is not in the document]
bilbof: A 30 second timeout for the healtcheck request was chosen since Publisher's Redis connection timeout is ~20 seconds.

The `/healthcheck` endpoint reports Redis connectivity. If Redis is unavailable then the healthcheck endpoint will return a `200` status (with Redis errors reported in JSON) after waiting ~20 seconds for Redis.

If the timeout were <20 seconds, then Redis being unavailable to Publisher would prompt the ALB to cycle the Publisher instance, despite the application instance itself being healthy.

This is a symptom of a bigger problem: `/healthcheck` endpoints are going to be used differently in ECS. We should implement e.g. a `/v2/healthcheck` endpoint to applications during the migration.

In the Icinga/EC2 world we used `/healthcheck.json` endpoints to check for specific problems with an application and dependencies. We probably don't want the Publisher service to tell its ALBs about Redis. We should instead put Redis connectivity into a `/metrics` endpoint for Prometheus (or Statsd in the meantime) and alert on Redis connectivity from Graphite/Prometheus. We could also instead alert on symptoms like 500s when users try to enqueue publishing jobs onto the Redis-backed queue.

terraform/publisher/main.tf
```diff
@@ -0,0 +1,350 @@
...
+
+resource "aws_route53_record" "internal_service_record" {
+  zone_id = data.aws_route53_zone.internal.zone_id
+  name    = "${var.service_name}.pink.${var.internal_domain_name}"
+  type    = "A"
+
+  alias {
+    name                   = aws_lb.internal_alb.dns_name
+    zone_id                = aws_lb.internal_alb.zone_id
+    evaluate_target_health = false
```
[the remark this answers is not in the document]
bilbof: There's no reason. Initially it was for consistency, but also because we thought this might be the reason that DNS lookups were giving inconsistent results, until we realised we had to update DNS records in both AWS *and* GCP.

terraform/modules/infra-fargate/main.tf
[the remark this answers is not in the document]
sengi: Good point - I think there'll be a fair bit of work and iterating on the design of the healthchecking and app-app communication in general. The ALB (target group) healthchecks are unusual in that they seem to conflate liveness (whether to kill the container) with readiness (whether to send traffic to the container), so I think you're right about needing to ignore dependencies on the app healthcheck.

On alerting, the winning strategy is indeed to alert on symptoms (as a rule of thumb, with occasional common-sense exceptions) and set up dashboards to show up common causes easily (e.g by showing errors rates of incoming requests vs outgoing requests to dependencies).

PR #11, opened 2020-09-01
Frontend DNS record

bilbof:
This record will allow us to route traffic to the frontend machine on frontend.pink.test (including quickly switching back to the frontend app on that machine), and also route traffic to the frontend application running in ECS at frontend-ecs.test.

I propose that we use DNS changes where possible to route traffic to services in ECS. However, Frontend is a special case, since 'frontend' is also used for both the frontend machine and application. For example a request for `static.test.govuk-internal.digital/some-resource` will be routed to the 'frontend' machine and then to the static app on that machine.

To route traffic to frontend in ECS, I've created this new DNS record and updated the Frontend 'Backend' in the test account's deployment of Router API to send requests for frontend to this new host.

### DIFF
--- a/terraform/frontend/main.tf
+++ b/terraform/frontend/main.tf
@@ -25,3 +25,22 @@ module "fargate-console" {
   service_name          = "frontend_console"
   container_definitions = file("../task-definitions/frontend_console.json")
 }
+
+# Internal DNS record
+
+data "aws_route53_zone" "internal" {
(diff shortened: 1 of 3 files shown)


issyl0:
> To route traffic to frontend in ECS, I've created this new DNS record and updated the Frontend 'Backend' in Router API to send requests for frontend to this new host.

In the test account.

issyl0 (approved):

PR #13, opened 2020-09-03
Run Content Store on Elastic Container Service / Fargate

bilbof:
This brings up the content store application in ECS.

The content store app is still running in EC2 in the test environment. We can switch traffic between the two hosting environments with a DNS record change to `content-store.<env>.govuk-internal.digital`.

### DIFF
--- a/terraform/content-store/README.md
+++ b/terraform/content-store/README.md
@@ -1,0 +1,3 @@
+# Content Store application
+
+This project module manages the resources required to run content-store.
--- a/terraform/content-store/main.tf
+++ b/terraform/content-store/main.tf
@@ -1,0 +1,176 @@
+terraform {
+  backend "s3" {
+    bucket  = "govuk-terraform-test"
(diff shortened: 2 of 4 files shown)


issyl0:
How did you fix the `unicorn` errors in the container?

bilbof:
@issyl0 I didn't see any unicorn errors for this one, though there were errors related to booting up which had unicorn in the stacktrace - those were caused by my leaving out some environment variables.

I saw a MongoDB error (`Mongo::Error::NoServerAvailable`) a couple of times when bringing up the service, but annoyingly I wasn't able to reproduce that.

I had to add the following line to the Dockerfile to hack-fix the content schemas:

```
# Hack in content-schemas
RUN git clone https://github.com/alphagov/govuk-content-schemas $GOVUK_CONTENT_SCHEMAS_PATH
```

bilbof:
> The one rule of all this stuff seems to be: if it doesn't work, you've missed an envvar. 😛

Very true 😃

PR #17, opened 2020-09-17
Use a service mesh for internal networking in ECS

bilbof:
We have an opportunity to change how we do internal networking with the move from EC2 to ECS.

This adopts AWS App Mesh (service mesh) and AWS Cloud Map (service discovery) and removes the internal application load balancers (ALBs) for services running in ECS.

Benefits:

- We no longer need to manage the DNS records for services. Instead Cloud Map will create records in Route53 on our behalf, so that we can route requests e.g. to a Content Store ECS service via content-store.mesh.govuk-internal.digital. We still get some control e.g. over the TTL of the created DNS records.
- We no longer run internal ALBs for ECS services
- We get the benefits of running a service mesh managed by AWS. By running AWS-managed images of the [Envoy proxy](https://www.envoyproxy.io/) as a sidecar to our application instances in ECS, our services have consistent ingress/egress logs (similar to the nginx `access.log` we have in EC2). We can introduce gradual/canary deployment strategies using App Mesh. More benefits advertised here: https://aws.amazon.com/app-mesh.

Downsides:

- Service meshes are more complex than application load balancers, and App Mesh is relatively new product (in a few days of testing we hit bugs in the Terraform AWS provider such as https://github.com/terraform-providers/terraform-provider-aws/issues/4853)

An additional change included in this PR: we have begun to adopt a monolithic Terraform module, 'govuk', from which all Terraform commands will be run. This PR moves the Content Store, Publishing API, and Publisher apps to be managed by this module. There are pros and cons to this approach that we hope to discover by trying this out.

One additional minor change is that I've chosen to run all applications on port 80. There's no reason for all services to run on a unique port.

See the individual commit messages for specific details. Take a look at the final commit (adding content store to the mesh) which shows how this changes how we'll run an individual application in practice.

Story: https://trello.com/c/5u69DRCf/165-implement-appmesh-for-ecs-fargate-apps

### DIFF
--- a/terraform/content-store/cloudmap.tf
+++ b/terraform/content-store/cloudmap.tf
@@ -1,0 +1,14 @@
+resource "aws_service_discovery_service" "service" {
+  name = var.service_name
+
+  dns_config {
+    namespace_id = var.govuk_publishing_platform_namespace_id
+
+    dns_records {
+      ttl  = 60
+      type = "A"
+    }
+
+    routing_policy = "WEIGHTED"
+  }
+}
--- a/terraform/content-store/cloudmap.tf
+++ b/terraform/content-store/cloudmap.tf
@@ -5,7 +5,7 @@ resource "aws_service_discovery_service" "service" {
     namespace_id = var.govuk_publishing_platform_namespace_id
 
     dns_records {
-      ttl  = 60
+      ttl  = 10
       type = "A"
     }
 
--- a/terraform/content-store/main.tf
+++ b/terraform/content-store/main.tf
@@ -1,12 +1,3 @@
-terraform {
-  backend "s3" {
-    bucket  = "govuk-terraform-test"
-    key     = "projects/app-content-store.tfstate"
-    region  = "eu-west-1"
-    encrypt = true
-  }
-}
-
 provider "aws" {
   version = "~> 2.69"
   region  = "eu-west-1"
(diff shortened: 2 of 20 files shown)


terraform/content-store/cloudmap.tf
```diff
@@ -0,0 +1,14 @@
+resource "aws_service_discovery_service" "service" {
+  name = var.service_name
+
+  dns_config {
+    namespace_id = var.govuk_publishing_platform_namespace_id
+
+    dns_records {
+      ttl  = 60
```
sengi, line 8: We might want to revisit this later (e.g. Amazon [previously suggested 10](https://aws.amazon.com/blogs/aws/amazon-ecs-service-discovery/)) but no need to worry about it for now.

terraform/content-store/main.tf
```diff
@@ -67,110 +79,35 @@ resource "aws_ecs_service" "service" {
...
-  load_balancer_arn = aws_lb.internal_alb.arn
-  port              = "443"
-  protocol          = "HTTPS"
-  ssl_policy        = "ELBSecurityPolicy-2016-08"
-  certificate_arn   = data.aws_acm_certificate.internal_elb_cert.arn
-
-  default_action {
-    type             = "forward"
-    target_group_arn = aws_lb_target_group.internal_lb_tg.arn
+  egress {
```
sengi, line 92: At some point soon we'll need to switch from inline rules to separate rule resources (because inline will break extensibility when we modularise this). Also not something to worry about in this PR though.

sengi (approved):
👍 Thanks for getting all this going.

terraform/content-store/cloudmap.tf
bilbof: Thanks, I've pushed a commit for this.

PR #18, opened 2020-09-22
Move Terraform modules into TechOps-style dir structure.

sengi:
This PR only moves stuff around and fixes up paths; it doesn't change anything else.

- Non-root modules go in modules/
- Root modules go in deployments/

Fix up paths so that all the root modules are still deployable.

I also had to regenerate the detect-secrets baseline file in order for it to allow this commit. I ran `detect-secrets scan > .secrets.baseline` followed by `detect-secrets audit .secrets.baseline` (from the repo root in both cases).

Design doc: https://docs.google.com/document/d/19s4sYWn3xuPj6QivDh73jnuNMftLdz4yefPAvjT7UIA/edit# (I'm intentionally holding off on writing ADRs for a bit as this might change quite a bit over a short time.)

Tested:
- `tf plan` still works for each of the root modules (which now live under `deployments/`).

Things which are for subsequent PRs:
- There are some root modules which ought to become non-root modules (basically everything currently under `deployments/` except `govuk-test`).
- Write a README that explains the directory structure and so on.
- Make a reusable module for an app deployed on ECS; factor out most of the repetitive stuff in `modules/apps/*`.


seb-szy (approved):
LGTM

PR #19, opened 2020-09-23
Make frontend a non-root module like the other apps.

sengi:
Hook frontend into the `deployments/govuk-test` root module and make it look like the others, copying (for now) the same stuff for using AppMesh and so on.

I had to regenerate ` .secrets.baseline` yet again in order for the pre-commit to pass. I think we might have to do something about that soon because it's really quite a pain. (I can't just add `pragma: allowlist secret` comments because it's JSON.)

[Design doc](https://docs.google.com/document/d/19s4sYWn3xuPj6QivDh73jnuNMftLdz4yefPAvjT7UIA/edit#)

Tested:
- `tf apply` works. ECS task stays up.

For later:
- Factor out the common ECS task/service stuff etc. into a module so that we're not duplicating a ton of stuff in every app module.


richardTowers (approved):
Looks like a good step to me.

The secrets thing we could probably solve by templating the json files in HCL ([maybe these are the docs?](https://www.terraform.io/docs/configuration/functions/templatefile.html#generating-json-or-yaml-from-a-template)).

A couple of TODOs to deal with later, and there's a couple of hardcoded bits (like `EgressIgnoredIPs = "169.254.170.2,169.254.169.254"`) which we'll want to clean up at some point. But I think this is consistent with the rest of the codebase at the moment.

sengi:
Yeah I suspect templating the json is probably inevitable 🙈

I had my sights on that `EgressIgnoredIPs` as well 😁 I think it comes from an outdated example before [AppMesh stopped trying to intercept metadata server traffic](https://github.com/aws/amazon-ecs-agent/pull/1896). I'm hoping to do a cleanup pass soon to fix up a lot of that sort of thing.

PR #20, opened 2020-09-23
Create govuk module to factor out per env config

MahmudH:
Push all the resources and module invocations from `deployments/govuk-test` down into a shared module `modules/govuk` which creates all the GOV.UK Publishing resources for a given environment. `deployments/govuk-test` (and later integration, staging, etc.) will just call `modules/govuk` with parameters for settings which vary by environment.

Also clean up the way we define providers. Remove provider definitions from non-root modules, define a minimum provider version in each non-root module and define the provider configuration itself only in root modules — as [recommended in Terraform docs](https://www.terraform.io/docs/configuration/modules.html#providers-within-modules).

This doesn't pull out any of the per-environment settings into module variables (parameters) yet. We'll do that in a subsequent PR.

[Design doc](https://docs.google.com/document/d/19s4sYWn3xuPj6QivDh73jnuNMftLdz4yefPAvjT7UIA/edit#)

Tested: we `tf destroy`ed the previous rig and successfully `tf apply`ed this. Cluster and apps came up successfully.

Co-authored-by: Chris Banks <chris.banks@digital.cabinet-office.gov.uk>

### DIFF
--- a/terraform/deployments/govuk-test/main.tf
+++ b/terraform/deployments/govuk-test/main.tf
@@ -8,7 +8,7 @@ terraform {
 }
 
 provider "aws" {
-  region  = "eu-west-1"
+  region = "eu-west-1"
 }
 
 module "govuk" {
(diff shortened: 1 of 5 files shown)


sengi (approved):

richardTowers (approved):
Nice

PR #21, opened 2020-09-23
WIP - move container_definitions inline

richardTowers:
We can use jsonencode to define container_definitions without having to
load them from a different file.

This would allow us to use the full "power" of HCL to template out the
JSON, which could help us avoid repeating huge chunks of it.

It would probably also allow us to grab the references to secrets from
secrets manager based on their friendly name, instead of hardcoding the
ARNs like we're doing now (
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/secretsmanager_secret
)

Mostly just pushing this commit as a proof of concept - it would require
a bit more refactoring to make this nice.


sengi:
This is waaay better than what we have now and also much better than string-replacement templates. I'm definitely up for switching to this approach.

I think this should solve:
* having to hard-code Secrets Manager ARNs and their random suffixes (as you pointed out)
* ability to use comments
* the need to keep re-baselining detect-secrets; we'll be able to use inline comments to tell it that something's not a secret
* duplication of the task definition between the console and non-console versions of the task (assuming we end up still needing the two different versions of each task)
* ability to configure different environments without excessive duplication
* probably other things I haven't thought of

richardTowers:
👍  yeah. There's also an approach with `templatefile()` and `jsonencode()` which would allow us to keep the JSON in a separate file, but I actually think it's nicer inline. I'm sure we'll work out a nice way of factoring it as we go.

One thing I think we'll need to have a policy on is "how many HCL features should we use when we're generating JSON?".

In this example, I used `=` instead of `:`, which is more idiomatic HCL, but kindof makes it look like weird JSON. Most of the keys don't need quoting in HCL, and we could do things like:

```
# This:
environment { name = "APPMESH_VIRTUAL_NODE_NAME", value = "mesh/govuk/virtualNode/frontend" }
environment { name = "ENVOY_LOG_LEVEL", value = "debug" }

# Instead of this:
"environment" = [
   { "name" = "APPMESH_VIRTUAL_NODE_NAME", "value" : "mesh/govuk/virtualNode/frontend" },
   { "name" = "ENVOY_LOG_LEVEL", "value" : "debug" }
 ],
```

... but then it's quite hard to work out what the actual JSON will look like if you haven't swallowed the HCL language specification.

From https://github.com/hashicorp/hcl - "JSON can be used as a completely valid input to a system expecting HCL", so I think my recommendation would be to stick as close to JSON as possible in expressions which are being used to generate JSON (so `:`s rather than `=`s, and generally steering clear of other "syntax sugar").

sengi:
Sounds like we're gonna need a style guide sooner than I thought 😅 (or better still, a really configurable linter) Great points though. I agree sticking with JSON syntax sounds like the sensible option. We can always review that if we find it doesn't pan out the way we expect.

bilbof:
Excellent work! 👍

richardTowers:
This has been incorporated in other work now, so the PR is no longer needed.

PR #22, opened 2020-09-25
More Terraform refactoring

sengi:
* Remove fargate-security root module.
* Move the IAM resources for the ECS cluster (the task role and the execution role) to the `govuk` module since they're an essential part of the cluster, and we plan to have only one cluster per deployment for the foreseeable future.
* Avoid hardcoding some references like the ECS cluster name.
* Move the task definition JSON for `content-store` in line with task resource.
* Fetch Secrets Manager ARNs using Terraform data resources, again just for `content-store` for now. This avoids the need to hardcode the 6-character random strings from Secrets Manager (clearly a non-starter).

The `deployments/fargate-security` root module contained only a couple of security groups relating to the ECS cluster. We don't want to deploy these separately from the cluster, so move them out into the `govuk` composition module. Considered starting a separate module for the ECS cluster, but decided against this (at least for now) because we expect to have just one ECS cluster per publishing platform environment.

Inlining the task definitions (per @richardTowers's suggestion, #21) solves a lot of problems - see comments on #21. We're only doing this for `content-store` in this PR but the others will follow shortly.

Tested: applies cleanly with no diff (apart from the spurious diffs from the [aws_acm_certificate bug](https://trello.com/c/zKSqOvBJ/213), which are harmless and not relevant here)

Co-authored-by: @richardTowers 

Next steps:
* Inline the rest of the task definitions (with `jsonencode()` like we've done for `content-store`).
* Fetch the rest of the Secrets Manager ARNs via data resources (like we've done for `content-store`).


richardTowers (approved):
Looks good to me - lots more improvements we can make now that the container definitions can be dynamic 👍

PR #23, opened 2020-09-25
Bump Terraform version and configure Dependabot.

sengi:
* Upgrade Terraform from 0.13.0 to 0.13.3.
* Configure Dependabot to scan the Terraform codebase.

I couldn't find any evidence that Dependabot can upgrade .terraform-version and apparently it doesn't support updating provider versions yet either, but let's give it a try and see how it does.

If it's useful, we'll probably want to configure whom it sends reviews to by default (perhaps via a GitHub team). If I understand correctly, it'll just raise PRs for now and not add any specific reviewers. I think that's probably fine for evaluation purposes.

Tested: `terraform apply` works, no diff. [Validated](https://dependabot.com/docs/config-file/validator/) the Dependabot config but otherwise didn't test it (maybe there's a way?)


sengi:
Oh, it looks like Dependabot can't do the provider version yet. Maybe still worth trying in case it does something useful?

richardTowers:
I guess it would be useful if we decide to use any of the modules from https://registry.terraform.io/browse/modules.

I'm not sure whether we want to do that though - a lot of them are very thin wrappers around fairly simple resources...

sengi:
Yeah I'm sceptical about the value of some of those wrapper modules too.

I'm curious as to whether Dependabot will find anything to update. I suspect not yet 😞

bilbof (approved):
Nice one 👍

PR #24, opened 2020-10-02
Create an app module and use it for content-store (initially)

richardTowers:
Create a module which contains all the ECS / AppMesh / CloudMap boilerplate that's needed for a single microservice.

Apply this just to content-store for now, others to follow the same pattern shortly.

Refactor and simplify the way we're defining the security groups, in order to achieve a stable interface to the `app` module. We only need one security group per app and we don't want to have to modify the interface of the per-app modules (like content-store and publishing-api) whenever we want to change a firewall rule. Therefore move the security group rule definitions into the `govuk` composition module.

We've pulled enough variables out that neither the app module nor the content-store module refer to `test.blah` or `mesh.blah`.

Also rename a bunch of variables for consistency and clarity and improve the descriptions.

Tested: `terraform init && terraform apply` works (after having run a `terraform destroy` targeted to the old content-store module, because we're moving resources between modules)

Co-authored-by: @sengi


terraform/modules/app/main.tf
```diff
@@ -0,0 +1,87 @@
...
+  execution_role_arn       = var.execution_role_arn
+
+  proxy_configuration {
+    type           = "APPMESH"
+    container_name = "envoy"
+
+    properties = {
+      AppPorts         = var.container_ingress_port
+      EgressIgnoredIPs = "169.254.170.2,169.254.169.254" # TODO: no longer required (try omitting, might need to stay but empty?)
+      # TODO: what are these magic numbers and are they necessary?
```
bilbof, line 57: These magic numbers are sadly necessary. Envoy listens on 15000 and 15001. The IgnoredUID is required but set arbitrarily as "1337" since this is also used in the AWS documentation.

https://docs.aws.amazon.com/AmazonECS/latest/userguide/example_task_definitions.html#example_task_definition-containerdependency

https://github.com/awsdocs/amazon-ecs-developer-guide/blob/master/doc_source/appmesh-getting-started.md#-json-

bilbof (approved):
Excellent change

terraform/modules/app/main.tf
sengi: Thanks, that confirms my suspicions! :D I'll follow up with a PR to clean up some of those TODOs.

PR #26, opened 2020-10-05
Permit GOV.UK tools Concourse role to assume into test account

bilbof:
This will allow Concourse jobs to assume a role in the test AWS account and deploy ECS services.

Story: https://trello.com/c/huv5i9DJ/233


richardTowers (approved):
Looks good to me 👍

PR #27, opened 2020-10-06
Refactor publisher module to use the common app module.

sengi:
Follows on from #24.

* `apps/publisher` module now calls `modules/app` and the copy-pasted common stuff is deleted.
* Clean up related security groups and push variables up to the `govuk` composition module.
* Some app modules (the ones with load balancers) output two security groups now, so change the names of those outputs to differentiate them clearly.
* A few minor cleanups on `modules/apps/content-store` for consistency.
* Make Content Store wait for Envoy on startup. We already had this for Publisher, but we almost certainly need it everywhere.
* Look up external security groups by name using data resources instead of hardcoding the IDs.

There are still a bunch of environment variables in the task definition which need cleaning up and un-hardcoding (like we've already done for content-store). I'm intentionally not doing those as part of this PR as I don't want to delay any longer.

Tested: successfully completed a destroy/apply cycle. Also checked that factoring out the variables didn't lead to any diffs.

I haven't tested whether the apps can still talk to each other.


bilbof (approved):

sengi:
I did a bit more testing. Turns out publisher.test.publishing.service.gov.uk serves 503s from the load balancer with this PR whereas without it, it at least gets through to the app and serves the redirect to Signon. I've probably missed or messed up a security group rule or something. Might still be worth merging this and fixing forward though 🤷

PR #29, opened 2020-10-08
Update Terraform READMEs to reflect new structure.

sengi:
This is really only a temporary update so that the READMEs are at least somewhat pertinent to what's actually going on. We'll very soon want to [generate this documentation using `terraform-docs`](https://trello.com/c/jlCGZ2EB/235-set-up-terraform-docs), at which point the contents of these READMEs will move into the `main.tf`s of their respective directories.


bilbof (approved):
Excellent 👍

PR #32, opened 2020-10-13
Cleanup and fixes for Publisher module

sengi:
- Avoid hardcoding the `govuk_management_access` security group ID; fetch it by name in the root module and pass it down.
- Move a security group rule for the Publisher ALB from the publisher module into modules/govuk along with all the other security group rules (missed from #27).
- Some minor cleanup of the ALB config for Publisher (details in the commit message) and factor out the domain name.

This introduces a new top-level var `public_lb_domain_name` which looks a bit like a duplicate of `govuk_app_domain_external` as it's currently set to the same thing (`test.govuk.digital`) but I think we'll soon need the ability to vary them separately - see the 0f6951d commit message for more details.

Tested: `terraform apply` works and produces expected output.

### Tips for reviewing

This is probably best reviewed commit-by-commit rather than looking at the overall diff, because the mandatory `terraform fmt` makes it hard to see what's going on otherwise.

Please check that my var descriptions are accurate and sufficiently clear.

### DIFF
--- a/terraform/deployments/govuk-test/main.tf
+++ b/terraform/deployments/govuk-test/main.tf
@@ -28,6 +28,7 @@ module "govuk" {
   vpc_id                       = "vpc-9e62bcf8"                                            # TODO: hardcoded
   private_subnets              = ["subnet-6dc4370b", "subnet-463bfd0e", "subnet-bfecd0e4"] # TODO: hardcoded
   public_subnets               = ["subnet-6cc4370a", "subnet-ba30f6f2", "subnet-bfe6dae4"] # TODO: hardcoded
+  public_lb_domain_name        = "test.govuk.digital"
   govuk_app_domain_external    = "test.govuk.digital"
   govuk_app_domain_internal    = "test.govuk-internal.digital"
   govuk_website_root           = "test.publishing.service.gov.uk"
--- a/terraform/deployments/govuk-test/main.tf
+++ b/terraform/deployments/govuk-test/main.tf
@@ -24,19 +24,19 @@ data "aws_security_group" "redis" {
 }
 
 module "govuk" {
-  source                       = "../../modules/govuk"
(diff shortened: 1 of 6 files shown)


fredericfran-gds (approved):
Looks good (obviously, I don't know if this works since there are no end-to-end tests)

Minor suggestions/(my opinions) are that:
1. `terraform fmt` should have been applied on the separate commits. I have not seen people doing linting and theses kind of processes in 1 commit. I don't know if in a `several commits in 1 PR` scenario, you would expect each commit to be linted etc.
2. your explanation for a new top-level `var public_lb_domain_name` is not very clear in the PR description. I would have started by saying that in normal circumstances, the app needs to know about *.publishing.service.gov.uk but this is managed by another repo govuk-dns-config. This repo manages *test.govuk.digital and are allocated to resources in this repo. Hence why the 2 variables/domains. 

The description in the PR provides more detail than in commit no.3.

sengi:
Thanks for the review.

>     1. `terraform fmt` should have been applied on the separate commits.

I agree in general that each commit should pass lint checks in its own right. There's an overriding practical problem in the case of Terraform though, in that `terraform fmt` makes all the `=` signs line up, which makes it very hard to see what's actually changing and what's just being reformatted because there happens to be a new arg which is longer than the others. This PR is a particularly severe example of that, so I intentionally left the autoformatting as a separate commit - as a kindness to the reviewer!

In some previous PRs where the problem hasn't been so severe I've squashed the `terraform fmt`.

>     I would have started by saying that in normal circumstances, the app needs to know about *.publishing.service.gov.uk but this is managed by another repo govuk-dns-config. This repo manages *test.govuk.digital and are allocated to resources in this repo. Hence why the 2 variables/domains.

I think the key thing is that they're semantically different, regardless of whether they're defined in this repo or a different one. Thanks for the feedback about it being unclear though - I've reworded the explanation to be a bit more specific.
 
> The description in the PR provides more detail than in commit no.3.

Fair point - fixed!

PR #25, opened 2020-10-02
Add concourse pipeline to deploy apps to ECS

bilbof:
This is a demo example to start deploying apps to ECS through a Concourse pipeline.

(VPN access + GitHub production login required): https://cd.gds-reliability.engineering/teams/govuk-tools/pipelines/govuk-cd-frontend:

Screenshot 2020-10-06 at 18 18 21

Screenshot 2020-10-06 at 18 18 37

More work is required to:

- add working smoke tests, and other behaviour described in proposal 005
- replace ruby stuff with a concourse ECS plugin
- build and push Docker image to DockerHub in CI

These changes can come in future commits.

Story: https://trello.com/c/C2ubxCmM/175-minimal-deployment-solution
Epic: https://trello.com/c/KBHxP0tt/180-deployments

PR #33, opened 2020-10-14
Add PR template

fredericfran-gds:
Trello
------
[card](https://trello.com/c/AfiUfhEB/250-add-pr-template-to-govuk-infrastructure-repo)

Description
-----------
* A PR template will help new members know what is the GOV.UK standard for
  commits.
* This has been discussed in the comments of #30

Checks/Tests
------------
* none

Future Work
-----------
* Revise template to make it more useful by taking the feedback of team members

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,23 @@
+Succint one line description of the PR objectives
+
+Trello
+------
+* Link to Trello card if available
+
+Description
+-----------
+* Describe the objectives that this PR aims to achieve
+* Provide enough context to aid better understanding, for
+  e.g. can refer to previous related PRs 
+* See [doc](https://gds-way.cloudapps.digital/standards/pull-requests.html)
+  for further guidance
+
+Checks/Tests
+------------
+* Provide the different tests that you have performed
+
+Future Work
+-----------
+* Provide a brief description of future work to be done, for e.g.
+  fix hardcoded values 
+


sengi:
I think there a discussion to be had first about whether we really want a PR template. Writing good commit messages and PR descriptions is a cultural thing and while I think a template could potentially help nudge people in the right direction it also has the potential to be demotivating and feel like a form-filling exercise.

Personally I think if we decide to to this we might consider going with something much more minimal that serves as just a gentle reminder of what could be useful to add, rather than something that looks like a form to be filled.

fredericfran-gds:
> I think there a discussion to be had first about whether we really want a PR template. Writing good commit messages and PR descriptions is a cultural thing and while I think a template could potentially help nudge people in the right direction it also has the potential to be demotivating and feel like a form-filling exercise.
> 
> Personally I think if we decide to do this we might consider going with something much more minimal that serves as just a gentle reminder of what could be useful to add, rather than something that looks like a form to be filled.

I agree that this needs to be discussed in the team but you raised so many "requirements" in you another comment in #30 that I felt the need to formalise it for new team members (especially RE people) who may not be used to raising these types of PRs.

Ref:

> Content of the change looks good to me. Please write a meaningful PR description though.
>
> A good PR description:

> Gives enough context for someone to understand the change. (In this case, you could just say that it > follows on from #24 and #27 since those are very similar, but it's still important to give at least that > context.)
> Says how you tested it.
> Links to an issue/Trello card if there is one.
> If appropriate, mentions any follow-up work which is being left for later (for example if there's > cleanup or improvement that's intended to be done in a follow-up PR).
> We don't have a template yet but here are a couple of previous good examples: #25, #27
>
> Also bear in mind that this is a public repo so your PR descriptions are on show to the general 
> public!

sengi:
It's the formalisation that I feel is potentially problematic and needs to be dealt with sensitively. My suggestions in #30 were in the context of that PR and won't necessarily apply in all cases. I apologise if I was overly prescriptive there.

So far we've had no issues - people have been following the spirit of the GOV.UK style guide for [commit messages](https://github.com/alphagov/styleguides/blob/master/git.md) and [pull requests](https://github.com/alphagov/styleguides/blob/master/pull-requests.md#opening-a-request).

I rather regret having raised the issue of a template, as it represents a departure from how other GOV.UK teams work and I don't want to overly prescriptive. My preference would be that we just point people at the style guide as part of the onboarding process in future so that everyone's on the same page. If we do go with a template, my preference would be for something like a one-line reminder that links to style guide, or a very short summary of it.

At any rate, I'd recommend we pause this until the whole team is back in and if you're still keen on a template we can potentially revisit it then if that's ok.

fredericfran-gds:
I agree with discussing with the team after how the style should be.

For me, templates help in getting the right stuff in the first time. I saw it in `gds-cli` and `paas` repositories, it helps to see what teams expect from new committers. Obviously, the template is just guidance, sensible customizations should be encouraged.

I don't have a hard view on this but will help team members if we know what to expect.

fredericfran-gds:
the team not finding it useful, so archiving.

PR #34, opened 2020-10-15
Add Router Application

smford:
This adds the Router application.

Testing was done and the tasks spawns and runs without issue.

https://trello.com/c/BDKurf6m/246-bring-up-router-on-ecs

Pair: @sengi & @smford

### DIFF
--- a/terraform/modules/apps/router/main.tf
+++ b/terraform/modules/apps/router/main.tf
@@ -1,0 +1,85 @@
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 2.69"
+    }
+  }
+}
+
+data "aws_secretsmanager_secret" "sentry_dsn" {
+  name = "router_app-SENTRY_DSN"
+}
+
+module "app" {
+  source                           = "../../app"
+  cpu                              = "512"
+  memory                           = "1024"
+  vpc_id                           = var.vpc_id
+  cluster_id                       = var.cluster_id
+  service_name                     = var.service_name
+  subnets                          = var.private_subnets
+  mesh_name                        = var.mesh_name
+  service_discovery_namespace_id   = var.service_discovery_namespace_id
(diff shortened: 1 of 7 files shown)


terraform/modules/apps/router/main.tf
```diff
@@ -0,0 +1,77 @@
...
+      "image" : "govuk/router:deployed-to-production",
+      "essential" : true,
+      "environment" : [
+        { "name" : "GOVUK_APP_NAME", "value" : "router" },
+        { "name" : "GOVUK_APP_ROOT", "value" : "/var/apps/router" },
+        { "name" : "ROUTER_APIADDR", "value" : ":8081" },
+        { "name" : "ROUTER_BACKEND_HEADER_TIMEOUT", "value" : "20s" },
+        { "name" : "ROUTER_PUBADDR", "value" : ":8080" },
+        { "name" : "ROUTER_MONGO_DB", "value" : "router" },
+        { "name" : "ROUTER_MONGO_URL", "value" : "mongodb://${var.router_mongodb_host}" },
```
sengi, line 40: Thanks for doing the hard work to keep this simple here: eliminating unused environment variables, looking up which ones were set to defaults and making informed judgements about which of those should be included, etc. That kind of care and attention to detail is really going to help us with velocity and quality as we progress. ❤️

terraform/modules/apps/router/main.tf
```diff
@@ -0,0 +1,77 @@
...
+        { "name" : "SENTRY_ENVIRONMENT", "value" : "test" }
```
sengi, line 41: I just realised this is going to need to vary by environment, but no need to block merging - easily done as a follow-up.

sengi (approved):
Nice! LGTM

terraform/modules/apps/router/main.tf
smford: aye aye sir

PR #35, opened 2020-10-16
Update frontend app to use new modules

MahmudH:
- Apps/frontend now calls the new modules/app 
- Add new variable changes to variables.tf and main.tf files.
- Move JSON config
- Remove cloudmap and appmesh files

Trello: https://trello.com/c/pT71VNbi/244-refactor-apps-frontend-terraform-module-to-use-the-app-module


terraform/modules/apps/frontend/main.tf
```diff
@@ -7,79 +7,66 @@ terraform {
...
-  container_definitions    = file("${path.module}/frontend_console.json")
-  network_mode             = "awsvpc"
-  cpu                      = 512
-  memory                   = 1024
-  task_role_arn            = var.task_role_arn
-  execution_role_arn       = var.execution_role_arn
+module "app" {
+  source                           = "../../app"
+  cpu                              = "512"
+  memory                           = "1024"
```
sengi, line 17: These can be numbers rather than strings.
```suggestion
  cpu                              = 512
  memory                           = 1024
```

terraform/modules/apps/frontend/main.tf
```diff
@@ -7,79 +7,66 @@ terraform {
...
+        { "name" : "RAILS_ENV", "value" : "production" },
+        { "name" : "ASSET_HOST", "value" : "www.gov.uk" },
+        { "name" : "GOVUK_APP_DOMAIN", "value" : "www.gov.uk" },
+        { "name" : "GOVUK_WEBSITE_ROOT", "value" : "www.gov.uk" },
+        { "name" : "WEBSITE_ROOT", "value" : "www.gov.uk" },
+        { "name" : "PLEK_SERVICE_CONTENT_STORE_URI", "value" : "https://www.gov.uk/api" },
+        { "name" : "PLEK_SERVICE_STATIC_URI", "value" : "https://assets.publishing.service.gov.uk" },
+        { "name" : "GOVUK_ASSET_ROOT", "value" : "https://assets.digital.cabinet-office.gov.uk" },
+        { "name" : "STATSD_PROTOCOL", "value" : "tcp" },
+        { "name" : "STATSD_HOST", "value" : "statsd.test.govuk-internal.digital" },
```
sengi, line 44: As a follow-up you could consider taking some of these values from the vars which we're already passing down from `modules/govuk` (like govuk_app_domain and statsd_host for example). Don't worry about for this PR though - better to get this merged first.

sengi (approved):
Great! Thanks for taking care of this.

I ran `terraform apply` from the branch and it found no diffs, so I'm assuming your branch is already deployed - thanks for testing thoroughly. Also looks like the ECS task is happy.

Couple of minor comments, only the int/string one needs addressing for this PR.

Nice work!

PR #36, opened 2020-10-21
Attach Publisher's load balancer to its ECS service.

sengi:
Add a parameter to `modules/app` to allow attaching ALB/NLB target groups to an app's ECS service. Unfortunately we have to use Terraform's rather clunky "dynamic" construct in order to do this because the AWS provider doesn't have a separate resource type for associating an ECS service with a load balancer (unlike for EC2, where it provides aws_lb_target_group_attachment).

Define the healthcheck grace period on the ECS service if we're attaching one or more load balancers. Parameterise this as well, as we're likely to need to vary it. We have to avoid specifying health_check_grace_period_seconds on the aws_ecs_service resource if there are no load balancers, because otherwise the AWS TF provider complains; hence the ugly ternary conditional there.

[Trello card](https://trello.com/c/69WKE6kI/211)

Tested: `terraform apply` works. Publisher still doesn't serve requests (it gets killed by the healthcheck), but it's closer to working than it was before.

### DIFF
--- a/terraform/modules/app/main.tf
+++ b/terraform/modules/app/main.tf
@@ -69,6 +69,18 @@ resource "aws_ecs_service" "service" {
   desired_count   = var.desired_count
   launch_type     = "FARGATE"
 
+  health_check_grace_period_seconds = length(var.load_balancers) > 0 ? var.health_check_grace_period_seconds : null
+
+  dynamic "load_balancer" {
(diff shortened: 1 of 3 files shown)


fredericfran-gds (approved):
looks good. I guess Bill will have to refactor that in his PR if you merge in main

PR #37, opened 2020-10-22
Add draft-content-store

fredericfran-gds:
Following on from #13 and #22 which added content-store, we add the
draft version of content-store in order to build the draft stack of
GOV.UK.

We customize the existing content-store app module so that the
service_name variable can be used to differentiate between draft and non
draft app deployment.

Secrets are renamed and will have to be changed via CLI

Future work:
1. SENTRY_DSN secret is the same across all apps in same environment so
this needs to be fixed throughout the repo.

Ref:
1. [Task Trello Card](https://trello.com/c/AaBscxsV/262-spin-up-draft-content-store)
2. [Parent Trello Card](https://trello.com/c/X4jlkXyE/200-determine-a-minimal-subset-of-apps)

### DIFF
--- a/terraform/modules/apps/content-store/main.tf
+++ b/terraform/modules/apps/content-store/main.tf
@@ -8,22 +8,22 @@ terraform {
 }
 
 data "aws_secretsmanager_secret" "oauth_id" {
-  name = "content_store_app-OAUTH_ID"
+  name = "${var.service_name}_OAUTH_ID"
 }
 data "aws_secretsmanager_secret" "oauth_secret" {
-  name = "content_store_app-OAUTH_SECRET"
+  name = "${var.service_name}_OAUTH_SECRET"
 }
 data "aws_secretsmanager_secret" "publishing_api_bearer_token" {
-  name = "content_store_app-PUBLISHING_API_BEARER_TOKEN" # pragma: allowlist secret
+  name = "${var.service_name}_PUBLISHING_API_BEARER_TOKEN" # pragma: allowlist secret
 }
 data "aws_secretsmanager_secret" "router_api_bearer_token" {
-  name = "content_store_app-ROUTER_API_BEARER_TOKEN" # pragma: allowlist secret
+  name = "${var.service_name}_ROUTER_API_BEARER_TOKEN" # pragma: allowlist secret
 }
 data "aws_secretsmanager_secret" "secret_key_base" {
-  name = "content_store_app-SECRET_KEY_BASE" # pragma: allowlist secret
+  name = "${var.service_name}_SECRET_KEY_BASE" # pragma: allowlist secret
 }
 data "aws_secretsmanager_secret" "sentry_dsn" {
-  name = "content_store_app-SENTRY_DSN"
+  name = "${var.service_name}_SENTRY_DSN"
 }
 
 module "app" {
@@ -57,7 +57,7 @@ module "app" {
         { "name" : "GOVUK_STATSD_PREFIX", "value" : "fargate" },
         { "name" : "GOVUK_USER", "value" : "deploy" }, # TODO: clean up?
         { "name" : "GOVUK_WEBSITE_ROOT", "value" : var.govuk_website_root },
-        { "name" : "MONGODB_URI", "value" : "mongodb://${var.mongodb_host}/content_store_production" },
+        { "name" : "MONGODB_URI", "value" : "${var.mongodb_url}" },
         { "name" : "PLEK_SERVICE_PERFORMANCEPLATFORM_BIG_SCREEN_VIEW_URI", "value" : "" },
         { "name" : "PLEK_SERVICE_PUBLISHING_API_URI", "value" : "http://publishing-api.${var.service_discovery_namespace_name}" },
         { "name" : "PLEK_SERVICE_RUMMAGER_URI", "value" : "" },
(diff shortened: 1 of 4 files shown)


terraform/modules/apps/content-store/main.tf
```diff
@@ -8,22 +8,22 @@ terraform {
 }
 
 data "aws_secretsmanager_secret" "oauth_id" {
-  name = "content_store_app-OAUTH_ID"
+  name = "${var.draft ? "draft_content_store_app-OAUTH_ID" : "content_store_app-OAUTH_ID"}"
```
sengi, line 11: I'd recommend just substituting the name for now, like this. (I know that probably means re-inputting the secrets, sorry - shouldn't take more than a few mins though with a bit of command line and copy-paste.)

Or if you really don't fancy doing that then maybe just stick a TODO on this to say that the conditionals are a temporary hack until we get rid of these secrets. Just as long as the pattern doesn't start spreading around the codebase! 😅 

fyi - all of these secrets except for SENTRY_DSN will be going away in the near future (because using the signon-based auth for app-app traffic prevents us from bootstrapping a new environment without a ton of manual setup), and SENTRY_DSN is actually the same everywhere.

```suggestion
  name = "${var.service_name}-OAUTH_ID"
```

terraform/modules/apps/content-store/main.tf
```diff
@@ -57,7 +57,7 @@ module "app" {
         { "name" : "GOVUK_STATSD_PREFIX", "value" : "fargate" },
         { "name" : "GOVUK_USER", "value" : "deploy" }, # TODO: clean up?
         { "name" : "GOVUK_WEBSITE_ROOT", "value" : var.govuk_website_root },
-        { "name" : "MONGODB_URI", "value" : "mongodb://${var.mongodb_host}/content_store_production" },
+        { "name" : "MONGODB_URI", "value" : "${var.draft ? "mongodb://${var.mongodb_host}/draft_content_store_production" : "mongodb://${var.mongodb_host}/content_store_production"}" },
```
sengi, line 60: This should just be passed in from the caller. We should be passing in the URI rather than just the hostname.

terraform/modules/apps/content-store/main.tf
```diff
@@ -78,7 +78,7 @@ module "app" {
           "awslogs-create-group" : "true",
           "awslogs-group" : "awslogs-fargate",
           "awslogs-region" : "eu-west-1", # TODO: hardcoded region
-          "awslogs-stream-prefix" : "awslogs-content-store"
+          "awslogs-stream-prefix" : "${var.draft ? "awslogs-draft-content-store" : "awslogs-content-store"}",
```
sengi, line 81:
```suggestion
          "awslogs-stream-prefix" : "awslogs-${var.service_name}"
```

terraform/modules/apps/content-store/variables.tf
```diff
@@ -7,6 +7,12 @@ variable "cluster_id" {
   type        = string
 }
 
+variable "draft" {
+  description = "Draft mode of this application"
+  type        = bool
+  default     = false
+}
+
```
sengi, line 15:
```suggestion
```

sengi (commented):
I have to say I'm not convinced about adding a bunch of conditionals inside the module, sorry. Given that the only real differences between draft and regular content-store are its name and where it finds its database, it would be cleaner and simpler to just pass those things in and have the module be agnostic about whether it's the draft or the regular version.

I've given some more specific comments to show how I'd envisage doing that.

terraform/modules/apps/content-store/main.tf
fredericfran-gds: Hi, I don't mind using the service name (content-store) directly but I would have to create a local value (content_store) to have the right style?

fredericfran-gds: I haven't done the secrets yet, was waiting for feedback.

sengi: The style of those names isn't really something that anyone thought about btw - it's left over from the discovery / proof-of-concept stuff from a while back.

terraform/modules/apps/content-store/main.tf
```diff
@@ -7,23 +7,27 @@ terraform {
   }
 }
 
+locals {
+  service_name_snake_case = replace("${var.service_name}", "-", "_")
+}
+
```
sengi, line 13: I think it's better to just use the name as-is. This kind of thing tends to make it hard to trace where values are coming from (for example by breaking searchability of the codebase).

terraform/modules/apps/content-store/main.tf
```diff
@@ -78,7 +78,7 @@ module "app" {
           "awslogs-create-group" : "true",
           "awslogs-group" : "awslogs-fargate",
           "awslogs-region" : "eu-west-1", # TODO: hardcoded region
-          "awslogs-stream-prefix" : "awslogs-content-store"
+          "awslogs-stream-prefix" : "awslogs-draft-${var.service_name}",
```
sengi, line 81: There's a stray `draft-` here
```suggestion
          "awslogs-stream-prefix" : "awslogs-${var.service_name}",
```

fredericfran-gds: good catch

sengi (approved):

sengi:
I think this is looking pretty clean now. Thanks for tidying all this up.

PR #40, opened 2020-10-23
App deployment to ECS using Terraform

bilbof:
This PR implements "Create a new Terraform module for task definition revisions" (option D in the [deployments design doc][]).

This PR supersedes https://github.com/alphagov/govuk-infrastructure/pull/25.

Notable changes:

* Concourse pipelines are used for deploying new releases to ECS
* Minimal deployment functionality is implemented
* Concourse will call Terraform to create a task definition revision as part of a deploy, taking `application`, `env`, and `image_tag` as the only arguments. Concourse will use the AWS CLI to update the running ECS service to use the new task definition.
* Introduces task definition modules to define app task definitions
* Introduces a set of modules in `deployments/apps/<env>/<app>` (e.g. `deployments/apps/test/frontend`) for use by Concourse for deploying an app

Benefits:

* Infrastructure configuration, apart from a task definition `image_tag`, will be defined in Terraform
* The infrastructure, including apps, can be bootstrapped with one `terraform apply`
* Database migrations and other pre-deploy steps can be run prior to app deployment, which is the behaviour that developers are familiar with
* Branch deploys and other features of our current deployment system are possible (but not implemented here)

Downsides:

* Multiple entrypoints for applying Terraform might cause confusion (`deployments/apps/<env>/<app>` and `deployments/govuk-<env>`)
* The `image_tag` will not be defined in code (we default to using `deployed-to-<env>` in the bootstrapping step), therefore the task definition created during bootstrapping will not be identical to the running task definition.

There is some work to complete such as adding additional features (branch deploys, db migrations), which can come in later PRs.

Co-authored-by: @smford

[deployments design doc]: https://docs.google.com/document/d/1qmNK7TqR0_aR5i_JRITchsutTAEZavmzGs63LI1XNbA/edit


fredericfran-gds:
Looks good, I think we should just merge it as soon as possible and experiment with it to see what is missing/not working.

terraform/README.md
```diff
@@ -3,11 +3,22 @@
 ## Directory structure
 
 * `deployments`: root modules, from where you can run Terraform commands. These
-  should only call the `govuk` composition module and nothing else.
+  should only call the `govuk` composition module, and individual app deployment
+  modules.
+    * `govuk-test`: calls the `govuk` composition module to bring up the core
+      infrastructure
```
sengi, line 9: I think we should replace the phrase "core infrastructure" with something more specific. Would it be accurate enough to say that it brings up everything except for the task definitions?

terraform/README.md
```diff
@@ -3,11 +3,22 @@
...
+    * `apps`: called by the Concourse CD pipeline to create new task definition
+      revisions during a deploy.
+        * `test`
+          * `publisher`: deployment module for creating a new task definition
+            revision for the Publisher app, requires an `image_tag` variable.
```
sengi, line 14:
```suggestion
            revision for the Publisher app in the `test` environment. Requires an `image_tag` variable.
```

terraform/README.md
```diff
@@ -3,11 +3,22 @@
 ## Directory structure
 
 * `deployments`: root modules, from where you can run Terraform commands. These
-  should only call the `govuk` composition module and nothing else.
+  should only call the `govuk` composition module, and individual app deployment
+  modules.
```
sengi, line 7: The "These should only call..." sentence is probably more confusing than helpful now that we have two kinds of top-level modules - I'd consider deleting it.

terraform/README.md
```diff
@@ -3,11 +3,22 @@
...
+            revision for the Publisher app, requires an `image_tag` variable.
 * `modules`: non-root modules
     * `govuk`: composition module for an entire GOV.UK Publishing environment
     * `app`: reusable module for an app; contains the essential resources which all the apps need.
     * `apps`: composition modules for each app; calls the app module plus any
       app-specific resources.
         * `publisher`: module which creates the Publisher app
         * ...
+    * `task-definition`: reusable module for creating a task definition
+    * `task-definitions`: composition modules for application task definitions
```
sengi, line 23: I'm not sure whether "composition module" is a helpful term here as these modules really only represent one thing.

```suggestion
    * `task-definitions`: task definitions for each app
```

terraform/deployments/apps/test/content-store/common.tf
```diff
@@ -0,0 +1 @@
+../common.tf
```
sengi, line 1: Interesting idea to use symlinks for this. Let's see how that goes - seems like a win for now at least!

Looking at the bigger picture I think it's probably a smell that Terraform is not the right tool for config management of this complexity/scale, but let's not dwell on that 😅

terraform/deployments/apps/test/content-store/main.tf
```diff
@@ -0,0 +1,34 @@
...
+  govuk_app_domain_external        = local.govuk_app_domain_external
+  govuk_website_root               = local.govuk_website_root
+  image_tag                        = var.image_tag
+  mesh_name                        = local.mesh_name
+  mongodb_url                      = "mongodb://${local.mongodb_host}/content_store_production"
+  service_discovery_namespace_name = local.service_discovery_namespace_name
+  statsd_host                      = local.statsd_host
+  execution_role_arn               = data.aws_iam_role.execution.arn
+  task_role_arn                    = data.aws_iam_role.task.arn
+  sentry_environment               = "test"
```
sengi, line 33:
```suggestion
  sentry_environment               = local.sentry_environment
```

(and similarly in other occurrences below)

terraform/deployments/apps/test/content-store/variables.tf
```diff
@@ -0,0 +1,5 @@
+variable "image_tag" {
```
sengi, line 1: Is there a reason for outputs.tf being symlinked but not variables.tf? If I understand the design correctly, these modules all have the same interface and that's required in order for Concourse to be able to use them(?)

Same question for modules/task-definitions/router/outputs.tf

sengi (approved):
Go for it. (Some minor suggestions and a question below.)

richardTowers (approved):
I've had a look through the commits, and the finished file structure. I might be able to point out a few nitpicks, but generally I think this approach looks solid to me, so I suggest we merge it once you've addressed @sengi's comments.

A few things I think it would be worth considering in subsequent work:

* It's a bit confusing that we've got both `deployments/apps/blah` and `modules/apps/blah`, but the apps deployments don't call the apps modules (because the deployments only create task-definitions, and the modules create the whole ECS / AppMesh / CloudMap shenanigans). Maybe rename `deployments/apps/blah` to `deployments/task-definitions/blah`.

* Symlinking in the `common.tf` / `outputs.tf` files is clever, but it's a bit of a smell. We're still duplicating config between `deployments/govuk-test/main.tf` and `deployments/apps/test/common.tf`. I think we could make the app deployments read these settings from govuk-test's remote state, and then we wouldn't need separate app deployments for separate environments (because they could just point at the right remote state file).

* At some point it would be nice to reduce the complexity of the module structure. Right now we've got:
```
govuk-test --> govuk --> apps/blah --> app
                                   |
                                   \-> task-definition
```
we could get rid of govuk-test by using `.tfvars` files for each environment, and promoting govuk to be a deployment (as @sengi suggested). I wonder if it would be better not to create the task definitions at all in the `govuk` deployment, and just have those owned entirely by the `apps` deployments - that would make that side of the chain a bit simpler. But then we'd need to set the `deployment_controller` to `EXTERNAL` I think:
> task_definition - (Optional) The family and revision (family:revision) or full ARN of the task definition that you want to run in your service. Required unless using the EXTERNAL deployment controller. If a revision is not specified, the latest ACTIVE revision is used.

I think that can all be considered separately to this PR though - great work @bilbof and @smford !

terraform/deployments/apps/test/content-store/variables.tf
bilbof: Good spot, thanks. I'll address this in a subsequent PR since I'm going to look at removing the `common.tf` and `output.tf` files as per @richardTowers's suggestion below.

bilbof:
Thanks for your feedback, both! I've added a follow up card to address your comments at https://trello.com/c/2N5OLEHg/266.

PR #41, opened 2020-10-26
Add the signon service

rtrinque:
Add the signon service


terraform/deployments/govuk-test/main.tf
```diff
@@ -37,6 +37,8 @@ module "govuk" {
   mongodb_host                  = "mongo-1.test.govuk-internal.digital,mongo-2.test.govuk-internal.digital,mongo-3.test.govuk-internal.digital"
   redis_host                    = "pink-backend-redis.0f3erf.ng.0001.euw1.cache.amazonaws.com"
   statsd_host                   = "statsd.test.govuk-internal.digital"
+  signon_db_url                 = "mysql2://root:root@mysql-primary.test.govuk-internal.digital/signon"
```
fredericfran-gds, line 40: I wonder if this should be a secret as it encodes a username and password?

terraform/deployments/govuk-test/main.tf
```diff
@@ -37,6 +37,8 @@ module "govuk" {
...
+  signon_test_db_url            = "mysql2://root:root@mysql-primary.test.govuk-internal.digital/signon_test"
```
fredericfran-gds, line 41: same as above

terraform/modules/apps/signon/main.tf
```diff
@@ -0,0 +1,71 @@
...
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 2.69"
+    }
+  }
+}
+
+data "aws_secretsmanager_secret" "sentry_dsn" {
+  name = "signon_app-SENTRY_DSN"
```
fredericfran-gds, line 11: Based on #38, you should ideally use `name = "SENTRY_DSN"`

fredericfran-gds (commented):
I had a first pass, maybe wait for the experts for further feedback

terraform/deployments/govuk-test/main.tf
sengi: +1 Also ideally we'd make just the password a secret so that it's still clear what's connecting where. (Is it actually set to `root:root`? 🙈)

terraform/modules/apps/signon/variables.tf
```diff
@@ -0,0 +1,62 @@
...
+
+variable "private_subnets" {
+  description = "Subnet IDs to use for non-Internet-facing resources."
+  type        = list
+}
+
+variable "govuk_management_access_sg_id" {
+  description = "Group used to allow access by management systems"
+  type        = string
+  default     = "sg-0b873470482f6232d"
```
sengi, line 34: This default shouldn't be here as it's specific to the test environment. It's already being passed down from the top-level module but we haven't cleaned up some of the other occurrences of it yet (sorry about that).
```suggestion
```

sengi (commented):
Do you need an ALB to get end-user traffic into Signon? (there's a not-quite-working example in #36 😅) - feel free to do it in a separate PR if so.

terraform/deployments/variables/test/apps.tfvars
```diff
@@ -9,3 +9,5 @@ mongodb_host        = "mongo-1.test.govuk-internal.digital,mongo-2.test.govuk-in
 redis_host          = "pink-backend-redis.0f3erf.ng.0001.euw1.cache.amazonaws.com"
 sentry_environment  = "test"
 router_mongodb_url  = "mongodb://router-backend-1.test.govuk-internal.digital,router-backend-2.test.govuk-internal.digital,router-backend-3.test.govuk-internal.digital/router"
+signon_db_url       = "mysql2://root:root@mysql-primary.test.govuk-internal.digital/signon"
```
sengi, line 12: We'll need to rejig these at some point (for example to separate the hostname from the database name and probably also the credentials) but don't worry about it for now.

terraform/modules/apps/signon/variables.tf
sengi: Please remove this default.

terraform/modules/task-definitions/signon/main.tf
```diff
@@ -0,0 +1,78 @@
...
+    {
+      "name" : "signon",
+      "image" : "govuk/signon:deployed-to-production",
+      "essential" : true,
+      "environment" : [
+        { "name" : "GOVUK_APP_NAME", "value" : "signon" },
+        { "name" : "GOVUK_APP_ROOT", "value" : "/var/apps/signon" },
+        { "name" : "PORT", "value" : "8080" },
+        { "name" : "DATABASE_URL", "value" : "${var.signon_db_url}" },
+        { "name" : "RAILS_ENV", "value" : "development" },
```
sengi, line 45: I think you'll almost certainly want the rails env to be "production".

```suggestion
        { "name" : "RAILS_ENV", "value" : "production" },
```

terraform/modules/task-definitions/signon/main.tf
```diff
@@ -0,0 +1,78 @@
...
+  container_definitions = [
+    {
+      "name" : "signon",
+      "image" : "govuk/signon:deployed-to-production",
+      "essential" : true,
+      "environment" : [
+        { "name" : "GOVUK_APP_NAME", "value" : "signon" },
+        { "name" : "GOVUK_APP_ROOT", "value" : "/var/apps/signon" },
+        { "name" : "PORT", "value" : "8080" },
+        { "name" : "DATABASE_URL", "value" : "${var.signon_db_url}" },
```
sengi, line 44: Unnecessary quotes (there are a couple of other instances as well)
```suggestion
        { "name" : "DATABASE_URL", "value" : var.signon_db_url },
```

terraform/modules/task-definitions/signon/main.tf
```diff
@@ -0,0 +1,78 @@
...
+        { "name" : "RAILS_ENV", "value" : "development" },
+        { "name" : "TEST_DATABASE_URL", "value" : "${var.signon_test_db_url}" },
```
sengi, line 46: I suspect you'll find that TEST_DATABASE_URL isn't needed if you set the rails env to production instead of development.

sengi (approved):
Basically looks good, please address the remaining comments but otherwise good to merge 👍

PR #42, opened 2020-10-28
Remove disused .terraform-version from repo root.

sengi:
The `.terraform-version` file in use is the one in `/terraform`. This one is no longer used and its continued existence only causes confusion.


richardTowers (approved):

sengi:
Oh, turns out the lint hook was relying on it. Lol.

PR #46, opened 2020-10-29
Fix snags with the minimal deployment solution

richardTowers:
Fixes the concourse pipeline so that it runs cleanly on Big Concourse.

This is broken down into three commits:

- a424d7e - updates the terraform so it will try to assume a role if it's given one as a variable (needed because concourse needs to assume a role before it has permission to apply terraform)
- cee34be - grants concourse additional IAM permissions (which it needs to be able to apply terraform)
- e6c60bf - split the deploy-app task into two - one that creates a task definition by applying terraform, and one that updates the ECS service using the AWS CLI

These were enough to get all the apps deploying cleanly in the [deploy-apps-test](https://cd.gds-reliability.engineering/teams/govuk-tools/pipelines/deploy-apps-test) pipeline (note VPN required).

https://trello.com/c/w5O7J1Rb/268-fix-snags-with-the-minimal-deployment-solution


concourse/tasks/update-task-definition.yml
```diff
@@ -0,0 +1,38 @@
+platform: linux
+image_resource:
+  type: docker-image
+  source:
+    repository: governmentpaas/terraform
+    tag: terraform-0.13.3
```
sengi, line 6: It's gonna be fun finding a nice way to keep these versions in sync 😅 (not something to worry about right now though).

concourse/tasks/update-task-definition.yml
```diff
@@ -0,0 +1,38 @@
...
+inputs:
+  - name: govuk-infrastructure
+    path: src
+  - name: release
+outputs:
+  - name: terraform-outputs
+params:
+  AWS_REGION: eu-west-1
```
sengi, line 14: Might be nice to pass this into Terraform (but definitely outside the scope of this PR)

sengi (approved):
Nice!

terraform/deployments/apps/test/publisher/main.tf
```diff
@@ -16,6 +16,10 @@ terraform {
 
 provider "aws" {
   region = "eu-west-1"
+
+  assume_role {
+    role_arn = var.assume_role_arn
+  }
```
bilbof, line 22: Probably you've already checked this, but is it possible to only pass an `assume_role` block to the root module provider?

bilbof (approved):
Great one 👍

terraform/deployments/apps/test/publisher/main.tf
richardTowers: I didn't actually check - it would make it _much_ simpler if we only specified the provider's config at the deployment level (and didn't pass the assume_role_arn into the modules)

bilbof (approved):
Nice 👍 :shipit:

PR #47, opened 2020-10-30
Upgrade Terraform AWS provider to 3.13

bilbof:
This upgrade includes a nice ECS addition which should make bringing services up/down more reliable:

- adds wait_for_steady_state attribute to aws_ecs_service

This also follows Terraform's [best practices](https://www.terraform.io/docs/configuration/provider-requirements.html#best-practices-for-provider-versions) for managing provider versions. We now specify a minimum provider version for submodules, where we don't expect to `terraform apply`, and a maximum provider version at the root module.

### DIFF
--- a/terraform/deployments/govuk-test/main.tf
+++ b/terraform/deployments/govuk-test/main.tf
@@ -5,6 +5,13 @@ terraform {
     region  = "eu-west-1"
     encrypt = true
   }
+
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 3.13"
+    }
+  }
 }
 
 provider "aws" {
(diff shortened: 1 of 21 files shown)


terraform/deployments/govuk-test/main.tf
```diff
@@ -5,6 +5,13 @@ terraform {
     region  = "eu-west-1"
     encrypt = true
   }
+
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "3.13.0"
```
sengi, line 12: We might want to consider using `~>` here, at least until we have an automated solution for keeping these (per-environment) version numbers up to date.

sengi (approved):

terraform/deployments/govuk-test/main.tf
bilbof: Thanks for spotting that, I've fixed that up.

sengi:
Thanks for finding + fixing this - I hadn't spotted that we were an entire major version behind on the AWS provider 😅

PR #50, opened 2020-11-04
Add router api app

fredericfran-gds:
## Add router-api (44018038a1dead8b4c1d002bab71c27f6b082ac0)

This PR adds the router-api, which forms part of the set of minimal apps
required for the gov.uk stack.

The addition of the router-api follows new structure set in the #40 and

Refs:
1. [router api trello card](https://trello.com/c/9zgZxW9w/281-spin-up-router-api-app)
2. [minimal apps set trello card](https://trello.com/c/X4jlkXyE/200-determine-a-minimal-subset-of-apps)

## fixes to router-app (bc839ef710729e1e0899cee6e2c3794c83817ab4)

Various fixes to #34:
1. add service_name variable so as to be able to customize module for
draft version. We propagate this new variable throught the task
definitions.
2. remove unused data in terraform/modules/apps/router/main.tf
3. use same ports for router as in other environments
4. change name of mongo_host to mongo_url

## fix update task definitions (3748750a33b554eda94edf7b54fe1cdf0a59772e)

fixes based on
[doc](https://github.com/alphagov/govuk-infrastructure/blob/main/docs/applying-terraform.md#deploying-an-application):
1. fix directory of apps
2. fix terraform apply commands

## Add ability to have multipe mesh services per app (77a792a03af6df642b1aecfdf4c30f36bf2f4ab9)

Some apps may expose multiple ports/services. This commit allows the
configuration of multiple mesh services per app. The new
`custom_container_services` variable is flexible enough to change the
protocol of the mesh service too.

Note that we have to use a work around where multipe virtual
nodes/services are created per app since a virtual node can only have
one listener. This missing feature is reported
[here](aws/aws-app-mesh-roadmap#120)


terraform/modules/task-definitions/router/main.tf
```diff
@@ -57,19 +55,19 @@ module "task_definition" {
...
-          "hostPort" : 8080,
+          "containerPort" : 3054,
+          "hostPort" : 3054,
           "protocol" : "tcp"
         },
         {
-          "containerPort" : 8081,
-          "hostPort" : 8081,
+          "containerPort" : 3055,
+          "hostPort" : 3055,
```
bilbof, line 70: The Envoy proxy configuration for this task definition probably will need modifying to support these ports. At the moment (I believe) the sidecar proxy expects the app container to expose port 80 only.

No need to fix this in this PR, though!

bilbof (approved):
Nice work 👍 

One change is required, and I've added a query too.

I've noticed router's ports configuration likely mean traffic is not reaching the app instance, but you don't need to fix this here.

terraform/deployments/apps/router-api/main.tf
```diff
@@ -0,0 +1,36 @@
+terraform {
+  backend "s3" {
+    bucket  = "govuk-terraform-test"
+    key     = "projects/router.tfstate"
```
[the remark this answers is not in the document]
fredericfran-gds: Thanks, will change that

terraform/modules/apps/router-api/main.tf
```diff
@@ -0,0 +1,21 @@
...
+  source                           = "../../app"
+  execution_role_arn               = var.execution_role_arn
+  vpc_id                           = var.vpc_id
+  cluster_id                       = var.cluster_id
+  service_name                     = var.service_name
+  subnets                          = var.private_subnets
+  mesh_name                        = var.mesh_name
+  service_discovery_namespace_id   = var.service_discovery_namespace_id
+  service_discovery_namespace_name = var.service_discovery_namespace_name
+  extra_security_groups            = [var.govuk_management_access_security_group]
```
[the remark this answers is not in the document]
fredericfran-gds: I copied it from the router one, I will investigate further

terraform/modules/task-definitions/router/main.tf
fredericfran-gds: Thanks, I need to have a look how envoy is configured for apps.

terraform/deployments/apps/router-api/main.tf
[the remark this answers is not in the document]
fredericfran-gds: fixed

terraform/modules/apps/router-api/main.tf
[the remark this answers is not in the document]
fredericfran-gds: i will keep it as all the others seem to have it too.

terraform/modules/task-definitions/router/main.tf
fredericfran-gds: i fixed the envoy part i think but not the app mesh part because you can't have 2 listener per service: https://github.com/aws/aws-app-mesh-roadmap/issues/120

bilbof (approved):

terraform/modules/task-definitions/router-api/variables.tf
```diff
@@ -0,0 +1,39 @@
...
+  type        = string
+  description = "(optional) AWS IAM role to assume. Uses the role from the environment by default."
+  default     = null
+}
+
+variable "service_name" {
+  type = string
+}
+
+variable "router_urls" {
```
sengi, line 36: These don't actually seem to be URLs. Maybe call this router_nodes? (avoids potential for confusion and keeps it consistent with the environment variable)

terraform/deployments/apps/router/main.tf
```diff
@@ -24,10 +24,11 @@ provider "aws" {
 
 module "task_definition" {
   source             = "../../../modules/task-definitions/router"
+  service_name       = "router"
   image_tag          = var.image_tag
   mesh_name          = var.mesh_name
   execution_role_arn = data.aws_iam_role.execution.arn
-  mongodb_host       = var.router_mongodb_host
+  mongodb_url        = var.router_mongodb_url
```
sengi, line 31: Ideally this would be a list of hosts and we'd compose the URL at some appropriate point further down (probably inside the app module) - but don't worry about it for now. Just thought I should mention since you changed it. I'm guessing you changed it because you wanted to append the database name `/router`? If so, a cleaner way to do that would be to either introduce a variable for the database name or have it use `service_name` (which would let it handle multiple environments in the same account without any extra complexity).

terraform/deployments/apps/common.tf
```diff
@@ -10,6 +10,7 @@ locals {
   static_url                       = "https://static.${var.mesh_domain}"
   statsd_host                      = "statsd.${var.mesh_domain}" # TODO: Put Statsd in App Mesh
   website_root                     = "https://www.${var.app_domain}"
+  router_urls                      = "router.${var.mesh_domain}:3055"
```
sengi, line 13: This can't just be a load-balanced address for reaching an arbitrary router instance; it has to list all of the instances individually. Otherwise the route tree will be nearly always inconsistent because Router API will only be able to send a reload request to one arbitrarily-chosen Router instance each time there's a change, instead of reloading them all.

sengi (commented):
Thanks for adding Router API. Sorry for retrospective review but noticed a significant issue with the Router changes here - see below about Router's API port not being suitable for load balancing.

The good news is with a bit of luck we won't need that extra complexity of dealing with multiple ports in the app module. I'd recommend we remove it for now, but I've made some suggestions as to how to keep it maintainable anyway in case we need to resurrect it later.

terraform/modules/task-definitions/router-api/variables.tf
sengi: Thinking a bit more broadly, we might want to change the way Router API discovers Router instances (or the way Router is notified of changes). A static list basically prevents us from scaling Router without mucking about with Router API. This is a good example of an app issue to add to the backlog.

PR #52, opened 2020-11-05
Run db migrations during Publisher deployment

bilbof:
This will run database migrations using a new task Publisher application task definition prior to updating the ECS Service to use the new task definition.

The command 'rails db:migrate' will run in a separate task_runner cluster, and the one-off containers will not be responsible for serving user requests.

This follows the design proposal in [proposal 005].

[proposal 005]: https://github.com/alphagov/govuk-replatforming-discovery-2020/blob/1ceebbb0319ecb3b6be35e3dedea2331638cf5a8/proposals/005-ecs-fargate-deployment.md#database-migrations-will-be-performed-before-new-code-is-deployed

The run-task task is intentionally generic so that it can run arbitrary commands in a container with the same configuration as those serving users.

Additional commits are required to enable db migrations for other apps. I'll put up another PR for these.

Story: https://trello.com/c/jepU0c7G/232

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -153,6 +153,16 @@ jobs:
       params:
         APPLICATION: publisher
         GOVUK_ENVIRONMENT: test
+    - task: fetch-db-migrations-config
+      file: govuk-infrastructure/concourse/tasks/fetch-task-network-config.yml
+      params:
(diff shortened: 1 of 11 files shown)


fredericfran-gds (approved):
looks good. I guess running db:migration when sometimes there is no need to is not really an issue except for time taken to deploy.

PR #55, opened 2020-11-06
Set app branches from concourse settings file

richardTowers:
This will enable a git-based workflow for deploying branches of applications and infrastructure code to particular environments. This is useful in lower environments (like test or integration) as a way of demonstrating that changes work in a production-like setting.

GOV.UK developers currently deploy branches of applications by triggering a Jenkins job with custom parameters. Concourse does not have an equivalent user interface for providing parameters to builds.

This PR proposes the following workflow:

* Raise a PR against govuk-infrastructure, updating `concourse/settings/test.yml`
* This gives the team an opportunity to get visibility on which branches are being deployed (which is currently possible through the releases app)
* It also gives the team an opportunity to reject or approve the change to which branches are running in an environment. (Perhaps someone else is about to deploy a different branch of the same application - this gives them an opportunity to coordinate).
* Merge the PR once approved
* Concourse will pick up the change on `main` and automatically update the pipeline to deploy the branch of your app.
* Your app will be continuously deployed from your branch until you raise a PR to switch govuk-infrastructure back to `main`

It's also possible to update the branch of govuk-infrastructure in a similar way (although this might get a bit confusing).

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -15,7 +15,7 @@ resources:
     icon: github
     name: govuk-infrastructure
     source:
-      branch: main
+      branch: ((govuk-infrastructure-branch))
       uri: https://github.com/alphagov/govuk-infrastructure
     type: git
 
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -22,42 +22,42 @@ resources:
   - <<: *git-repo
     name: frontend
     source:
-      branch: master
+      branch: ((frontend-branch))
       uri: https://github.com/alphagov/frontend
       tag_filter: release_*
 
   - <<: *git-repo
     name: publisher
     source:
-      branch: master
+      branch: ((publisher-branch))
       uri: https://github.com/alphagov/publisher
       tag_filter: release_*
 
   - <<: *git-repo
     name: publishing-api
     source:
-      branch: master
+      branch: ((publishing-api-branch))
       uri: https://github.com/alphagov/publishing-api
       tag_filter: release_*
 
   - <<: *git-repo
     name: content-store
     source:
-      branch: master
+      branch: ((content-store-branch))
       uri: https://github.com/alphagov/content-store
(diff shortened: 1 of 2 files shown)


bilbof (commented):
Cool 👍

concourse/pipelines/deploy.yml
```diff
@@ -15,49 +15,49 @@ resources:
...
-      branch: master
+      branch: ((router-branch))
       uri: https://github.com/alphagov/router
       tag_filter: release_*
 
   - <<: *git-repo
     name: router-api
     source:
-      branch: master
+      branch: ((router-api-branch))
```
bilbof, line 60: One small danger to this approach is that it could require a `concourse/settings/production.yml` file, which can be modified to deploy a non-`master` branch to production. Perhaps a separate resource for integration might be in order, particularly as the `tag_filter` here is currently incompatible with branch deployments.

richardTowers: Ah, good point about the `tag_filter`. Perhaps we need to pull that into the settings too...

Something like:

`test.yml`
```
---
git:
  router-api:
    branch: master
    tag_filter: null
```

`production.yml`
```
---
git:
  router-api:
    branch: master
    tag_filter: release_*
```

Perhaps we could write a test to enforce that staging / production only had master / main branches.

richardTowers: (I think it's possible to do `((git.router-api.branch))` and `((git.router-api.tag_filter))` in concourse interpolations, but I haven't checked)

richardTowers: I think we'll need to bite the bullet and build container images in the pipeline to be able to do branch deployments at all, come to think of it.

bilbof: That looks reasonable 👍

richardTowers:
Closing this for now - we can come back to this idea a bit later (I think it makes more sense once we're building our own images, instead of relying on the ones Jenkins uses).

PR #58, opened 2020-11-09
Various improvements and fixes to allow frontend workflow to work

smford:
## Spin up frontend and fixes (b35ec6a)

Improvements:
1. add ability to set desired_count so that number of frontend tasks
2. use plain http rather than https for content-store and static
3. introduce load balancer
4. configure rails to serve assets because of no nginx

## Spin up content-store and fixes (9ab11ab)

Improvements:
1. introduce desired_count in order to enable number of content-store
tasks to be set

## Spin up Static app and fixes (038d3f6)

Improvements:
1. introduce desired_count to enable the number of Static tasks to be set
2. set Static port to 80
3. add missing secret and parameter
4. add public load balancer for static
5. configure static to serve assets

Ref:
1. [trello
   card](https://trello.com/c/mU7Gt0Bv/287-test-and-fix-frontend-app)
2. [minimal set of apps trello card](https://trello.com/c/X4jlkXyE/200-determine-a-minimal-subset-of-apps)

Co-Authored-By: Frederic Francois
<frederic.francois@digital.cabinet-office.gov.uk>

### DIFF
--- a/terraform/deployments/apps/common.tf
+++ b/terraform/deployments/apps/common.tf
@@ -1,7 +1,8 @@
 locals {
   asset_host                       = local.website_root
-  assets_url                       = "https://assets.${var.mesh_domain}"
-  content_store_url                = "https://content-store.${var.mesh_domain}"
+  assets_url                       = "https://static-ecs.${var.app_domain}"
+  frontend_assets_url              = "https://frontend.${var.app_domain}"
+  content_store_url                = "http://content-store.${var.mesh_domain}"
   redis_port                       = 6379
   service_discovery_namespace_name = var.mesh_domain
   sentry_environment               = "${var.govuk_environment}-ecs"
--- a/terraform/deployments/apps/common.tf
+++ b/terraform/deployments/apps/common.tf
@@ -6,7 +6,7 @@ locals {
   redis_port                       = 6379
   service_discovery_namespace_name = var.mesh_domain
   sentry_environment               = "${var.govuk_environment}-ecs"
-  static_url                       = "https://static.${var.mesh_domain}"
+  static_url                       = "http://static.${var.mesh_domain}"
   statsd_host                      = "statsd.${var.mesh_domain}"     # TODO: Put Statsd in App Mesh
   website_root                     = "https://www.${var.app_domain}" # TODO: Is this correct?
   router_urls                      = "router.${var.mesh_domain}:3055"
--- a/terraform/deployments/variables/test/apps.tfvars
+++ b/terraform/deployments/variables/test/apps.tfvars
@@ -2,7 +2,7 @@
 # App variables
 #--------------------------------------------------------------
 
-app_domain          = "test.publishing.service.gov.uk"
+app_domain          = "test.govuk.digital" # TODO: changed from test.publishing.service.gov.uk for easier testing.
 app_domain_internal = "test.govuk-internal.digital"
 govuk_environment   = "test"
 mongodb_host        = "mongo-1.test.govuk-internal.digital,mongo-2.test.govuk-internal.digital,mongo-3.test.govuk-internal.digital"
--- a/terraform/modules/apps/content-store/variables.tf
+++ b/terraform/modules/apps/content-store/variables.tf
@@ -40,3 +40,9 @@ variable "execution_role_arn" {
   description = "For use during bootstrapping"
   type        = string
 }
+
+variable "desired_count" {
+  description = "Desired count of Application instances"
+  type        = number
+  default     = 1
+}
--- a/terraform/modules/apps/frontend/main.tf
+++ b/terraform/modules/apps/frontend/main.tf
@@ -18,11 +18,89 @@ module "app" {
   service_discovery_namespace_id   = var.service_discovery_namespace_id
   service_discovery_namespace_name = var.service_discovery_namespace_name
   extra_security_groups            = [var.govuk_management_access_security_group]
+  desired_count                    = var.desired_count
+
+  load_balancers = [{
+    target_group_arn = aws_lb_target_group.public.arn
+    container_name   = var.service_name
+    container_port   = 80
+  }]
 }
 
-# TODO: Is this needed?
-resource "aws_security_group" "service" {
-  name        = "fargate_${var.service_name}_ingress"
+#
+# Internet-facing load balancer
+#
+
+# TODO: use a single, ACM-managed cert with both domains on. There is already
+# such a cert in integration/staging/prod (but it needs defining in Terraform).
+data "aws_acm_certificate" "public_lb_default" {
+  domain   = "*.test.govuk.digital"
+  statuses = ["ISSUED"]
+}
+
+data "aws_acm_certificate" "public_lb_alternate" {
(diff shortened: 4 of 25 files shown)


terraform/modules/task-definitions/frontend/main.tf
```diff
@@ -27,8 +27,8 @@ module "task_definition" {
   source             = "../../task-definition"
   mesh_name          = var.mesh_name
   service_name       = var.service_name
-  cpu                = 512
-  memory             = 1024
+  cpu                = 1024
+  memory             = 2048
```
bilbof, line 31: Was it necessary to increase CPU/memory?

terraform/modules/task-definition/main.tf
```diff
@@ -17,7 +17,8 @@ locals {
       "image" : "840364872350.dkr.ecr.eu-west-1.amazonaws.com/aws-appmesh-envoy:v1.15.0.0-prod",
       "user" : "1337",
       "environment" : [
-        { "name" : "APPMESH_VIRTUAL_NODE_NAME", "value" : "mesh/${var.mesh_name}/virtualNode/${var.service_name}" }
+        { "name" : "APPMESH_VIRTUAL_NODE_NAME", "value" : "mesh/${var.mesh_name}/virtualNode/${var.service_name}" },
+        { "name" : "APPMESH_RESOURCE_ARN", "value" : "mesh/${var.mesh_name}/virtualNode/${var.service_name}" },
```
bilbof, line 21: I read `APPMESH_RESOURCE_ARN` is going to replace `APPMESH_VIRTUAL_NODE_NAME`, did you see the same thing? I think we need to add `APPMESH_RESOURCE_ARN` all apps if that's the case. Not necessary for this PR though.

terraform/deployments/variables/test/apps.tfvars
```diff
@@ -2,7 +2,7 @@
 # App variables
 #--------------------------------------------------------------
 
-app_domain          = "test.publishing.service.gov.uk"
+app_domain          = "test.govuk.digital"
```
bilbof, line 5: Perhaps you could add a TODO comment here to explain why we're using this domain?

```diff
- app_domain          = "test.publishing.service.gov.uk"
+ app_domain          = "test.govuk.digital" # TODO: Changed until we have router running in ECS
```

terraform/modules/apps/content-store/variables.tf
```diff
@@ -40,3 +40,9 @@ variable "execution_role_arn" {
   description = "For use during bootstrapping"
   type        = string
 }
+
+variable "desired_count" {
+  description = "Desired count of Application instances"
+  type        = number
+  default     = 0
```
bilbof, line 47: Is this intended to default to zero?

bilbof (approved):

terraform/deployments/apps/common.tf
```diff
@@ -1,11 +1,12 @@
 locals {
   asset_host                       = local.website_root
-  assets_url                       = "https://assets.${var.mesh_domain}"
-  content_store_url                = "https://content-store.${var.mesh_domain}"
+  assets_url                       = "https://static-ecs.test.govuk.digital"
+  frontend_assets_url              = "https://frontend.test.govuk.digital"
```
sengi, line 4: If you intend to hardcode the test domain please add a TODO so that we can more easily find and fix it later. I think we already have a variable for `<env>.govuk.digital` somewhere (`public_lb_domain_name` I think - not the best name but it's there).

terraform/modules/apps/frontend/main.tf
```diff
@@ -18,11 +18,109 @@ module "app" {
...
+
+resource "aws_security_group_rule" "public_alb_egress_any_any" {
+  type      = "egress"
+  protocol  = "-1"
+  from_port = 0
+  to_port   = 0
+
+  security_group_id = aws_security_group.public_alb.id
+  cidr_blocks       = ["0.0.0.0/0"]
+}
```
sengi, line 110: Can we move these security group rules to `modules/govuk/security_group_rules.tf` please? (at least for the time being)

Same for the ones in `static/main.tf` below.

terraform/modules/apps/frontend/main.tf
```diff
@@ -18,11 +18,109 @@ module "app" {
...
+
+resource "aws_security_group_rule" "public_alb_443_https_office_ips" {
+  type      = "ingress"
+  from_port = 443
+  to_port   = 443
+  protocol  = "tcp"
+
+  security_group_id = aws_security_group.public_alb.id
+  cidr_blocks       = ["213.86.153.212/32", "213.86.153.213/32", "213.86.153.214/32", "213.86.153.235/32", "213.86.153.236/32", "213.86.153.237/32", "85.133.67.244/32"]
```
sengi, line 99: Having to duplicate this list of office networks multiple times seems messy and error prone. I suggest we instead just set a network ACL for the whole test VPC if we want to limit the test environment to the office.

terraform/modules/task-definition/main.tf
fredericfran-gds: Yes, we can do that in another PR, I was confused about whether we needed both but if we need only `APPMESH_RESOURCE_ARN`, then that's fine.

terraform/deployments/variables/test/apps.tfvars
fredericfran-gds: ok, I will add a comment.

terraform/modules/apps/content-store/variables.tf
fredericfran-gds: Originally when we put the `desired_count`, these were `0` but after the pairing session where I see you put it as default one we can change that.

terraform/deployments/apps/common.tf
fredericfran-gds: No, I was spiking on assets and found it easier. I will put the correct format.

terraform/modules/apps/frontend/main.tf
fredericfran-gds: Yes, I can do that.

terraform/modules/apps/frontend/main.tf
fredericfran-gds: Yes, I can do that.

sengi: Ok, please at least set a local in security_group_rules.tf then because the duplication is unacceptable.

fredericfran-gds: Yes, I have done that, I included a variable. Anyway, we will have to refactor in the public alb in the apps because they are duplication themselves, I was copying them straight.

terraform/modules/task-definitions/frontend/main.tf
fredericfran-gds: It was close to the limit on startup. Maybe we should look into that in more detail later.

terraform/modules/apps/frontend/main.tf
fredericfran-gds: Maybe there needs to be way an obvious way for people to know that. One would have expected the rules to be together with the group.

PR #59, opened 2020-11-13
Use govuk-aws/infra-networking remote state

richardTowers:
Hard-configuring the VPC ID and the subnets feels a bit fragile. If for some reason we were to destroy and re-create the VPC, we'd have to remember to update the values here.

Instead, we can use terraform's [terraform_remote_state](https://www.terraform.io/docs/providers/terraform/d/remote_state.html) data source to pull the outputs from the [govuk-aws/infra-networking](https://github.com/alphagov/govuk-aws/tree/test-revive/terraform/projects/infra-networking) outputs.

This effectively models the fact that the govuk-test deployment is "downstream" of the infra-networking deployment.

This reduces the number of "infrastructure" variables to the point where we could consider promoting the modules/govuk to be a deployment, with small, environment specific `.backend` files. These files should only need a few settings. Currently these are:

* `infra_networking_state_bucket`
* `public_lb_domain_name`
* `mesh_domain`
* `mesh_name`

We could use the same pattern for the various `apps` deployments - at the moment they have shared `common.tf` and `variables.tf` which mostly set things which could instead be provided as outputs of the `govuk-test` deployment.

### DIFF
--- a/terraform/deployments/govuk-test/main.tf
+++ b/terraform/deployments/govuk-test/main.tf
@@ -30,14 +30,25 @@ data "aws_security_group" "redis" {
   name = "govuk_backend-redis_access"
 }
 
+data "terraform_remote_state" "infra_networking" {
+  backend = "s3"
+  config = {
+    bucket = var.infra_networking_state_bucket
+    key    = "govuk/infra-networking.tfstate"
+    region = "eu-west-1"
+  }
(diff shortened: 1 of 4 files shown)


sengi (approved):
Cool - less hardcoded stuff ftw 👍

PR #63, opened 2020-11-18
Add wrapper script for aws ecs run-task for ease of running adhoc tasks

bilbof:
This is a tool to enable running ad-hoc tasks. It's expected to live
a few months until GDS CLI provides this feature. Building this into
GDS CLI at the moment is probably premature since the project may
change significantly in the next few months, so this tool should meet
our needs for now.

The tool prints a warning to users to let them know it's not going to
be around for long.

Expected usages:

```
# Run an ad-hoc rake task
run-task.sh -a frontend rake my:task
# Check connectivity between services in the mesh
run-task.sh -a publisher curl http://publishing-api.mesh.govuk-internal.digital
```
This allows one to perform temporary tasks easily.

### DIFF
--- a/tools/run-task.sh
+++ b/tools/run-task.sh
@@ -1,0 +1,91 @@
+#!/usr/bin/env bash
+# This temporary script runs an arbitrary task using an existing task defintion
+# in a new container, using ECS RunTask. This script will be replaced by
+# features built into the GDS CLI.
+
+set -eu
+
+OPTIND=1 # Reset in case getopts has been used previously in the shell.
+
+cluster="task_runner"
+govuk_env="test"
+
+function show_help {
+  echo "Usage: cd govuk-infrastructure && gds aws govuk-tools-internal-admin ./tools/run-rake-task -a frontend rake db:migrate"
+  exit 64
+}
+
+while getopts :a:c:e:h opt; do
+    case $opt in
+      h) show_help;;
+      a)
+        application=${OPTARG}
+        ;;
+      e)
+        govuk_env=${OPTARG:-$govuk_env}
+        ;;
+      :)
+        echo "Option -$OPTARG requires an app as the argument. e.g. -a frontend"
+        show_help
+        ;;
+    esac
+done
+
+shift $((OPTIND-1))
+[ "${1:-}" = "--" ] && shift
+command=$@
+
+if [ -z "${application}" ]; then echo "[Error] You must set the application arg" && show_help && exit 1; fi
+if [ -z "${command}" ]; then echo "[Error] You must provide a command" && show_help && exit 1; fi
+
+echo "[Warning] This is a temporary tool for use by the replatforming team."
+
+root_dir="${PWD}"
+
+app_dir="$root_dir/terraform/deployments/apps/$application"
+env_dir="$root_dir/terraform/deployments/govuk-test"
+
+echo "Fetching task_definition_arn from Terraform statefile in $app_dir"
+cd ${app_dir}
+terraform init >/dev/null
+task_definition_arn=$(terraform output task_definition_arn)
+
+echo "Fetching network_config from Terraform statefile in $env_dir"
+cd ${env_dir}
+terraform init >/dev/null
+private_subnets=$(terraform output -json private_subnets)
+security_groups=$(terraform output -json $application'_security_groups')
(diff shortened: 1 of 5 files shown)


sengi (commented):
This seems fine to me as a temporary thing, but I have a couple of concerns:

1. People might start adding all sorts of stuff to this and it could gain a life of its own. (not a reason not to do it - but a reason to manage it carefully and make sure the intent as to what it's for and what it's not for is very clear to anyone who might add to it)
2. It's not clear to me how the pipeline or the job are intended to be used, or that they're intended to be temporary, from looking at the code. Is there a way we can make this more self-documenting?

bilbof:
@sengi thanks!

I've added a comment to the pipeline file to highlight it's a temporary tool for our team.

I've renamed a couple of the keys for the jobs and groups to make it clearer what they are for. Is there anything in particular that is unclear in the file?

As an aside, we may find we want to run ad-hoc tasks in Concourse pipelines, rather than having developers send requests directly to ECS. This would let us re-used Concourse config. Alternatively, we could also have ad-hoc rake tasks go through code review, as we're planning to do with branch deploys, and not provide CLI tooling for rake tasks. Jobs in Concourse can be triggered and parameterised from the command line with `fly execute` I believe.

bilbof:
@sengi I've replaced the pipeline with a script.

This will be quicker to run and should be closer to the CLI tooling we hope to provide in future.

tools/run-task.sh
```diff
@@ -0,0 +1,90 @@
...
+shift $((OPTIND-1))
+[ "${1:-}" = "--" ] && shift
+command=$@
+
+if [ -z "${command}" ]; then echo "[Error] You must set the application arg" && show_help && exit 1; fi
+if [ -z "${command}" ]; then echo "[Error] You must set provide a command" && show_help && exit 1; fi
+
+echo "[Warning] This is a temporary tool for use by the replatforming team."
+
+root_dir=$(pwd)
```
sengi, line 42: As a rule of thumb it's a good idea to use quotes in case the variable contains whitespace. (and might as well use $PWD rather than a subshell)

```suggestion
root_dir="${PWD}"
```

tools/run-task.sh
```diff
@@ -0,0 +1,90 @@
...
+cd ${app_dir}
+terraform init >/dev/null
+task_definition_arn=$(terraform output task_definition_arn)
+
+echo "Fetching network_config from Terraform statefile in $env_dir"
+cd ${env_dir}
+terraform init >/dev/null
+private_subnets=$(terraform output -json private_subnets)
+security_groups=$(terraform output -json $application'_security_groups')
+network_config='awsvpcConfiguration={subnets='$private_subnets',securityGroups='$security_groups',assignPublicIp=DISABLED}'
```
sengi, line 57: Consider using one double quoted string here for readability rather than a series of single-quoted strings, something like:

```suggestion
network_config="awsvpcConfiguration={subnets=$private_subnets,securityGroups=$security_groups,assignPublicIp=DISABLED}"
```

sengi (approved):
LGTM - just one thing that might need fixing and a bunch of trivial suggestions.

tools/run-task.sh
```diff
@@ -0,0 +1,90 @@
...
+set -eu
+
+OPTIND=1 # Reset in case getopts has been used previously in the shell.
+
+cluster="task_runner"
+govuk_env="test"
+
+function show_help {
+  echo "Usage: cd govuk-infrastructure && gds aws govuk-tools-internal-admin ./tools/run-rake-task -a frontend rake db:migrate"
+}
```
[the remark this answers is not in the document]
bilbof: Thanks, that's awesome - I wasn't aware.

tools/run-task.sh
```diff
@@ -0,0 +1,90 @@
...
+
+while getopts :a:c:e:h opt; do
+    case $opt in
+      h) show_help && exit 0;;
+      a)
+        application=${OPTARG} >&2
```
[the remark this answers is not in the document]
bilbof: Thanks, those were unnecessary - I've removed them

tools/run-task.sh
bilbof: Thanks for the tip!

tools/run-task.sh
bilbof: Cool, I've done that 👍

PR #66, opened 2020-11-19
Set correct Slack channel for on_failure alerts

bilbof:
This channel key isn't required but it's a handy flag for where
the Slack messages are being sent.


fredericfran-gds (approved):
looks good

PR #69, opened 2020-11-19
fix frontend asset images

fredericfran-gds:
If rails is serving files rather the nginx or other web servers,
need to set config.action_dispatch.x_sendfile_header to nil

In the frontend case, this is set via setting the env variable:
`HEROKU_APP_NAME`

The other env variable change `RAILS_SERVE_STATIC_FILES` is only
to keep consistency with the frontend heroku config.

### DIFF
--- a/terraform/modules/task-definitions/frontend/main.tf
+++ b/terraform/modules/task-definitions/frontend/main.tf
@@ -57,8 +57,9 @@ module "task_definition" {
         { "name" : "STATSD_PROTOCOL", "value" : "tcp" },
         { "name" : "STATSD_HOST", "value" : var.statsd_host },
         { "name" : "GOVUK_STATSD_PREFIX", "value" : "fargate" },
-        { "name" : "RAILS_SERVE_STATIC_FILES", "value" : "enabled" },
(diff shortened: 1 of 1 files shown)


bilbof (approved):

sengi:
Just something to remember for next time: when we put temporary (or potentially temporary) workarounds like this in place it's very useful to put a `# TODO: ...` so that we can find them later.

sengi:
Also it's a bit strange that `yes` seems to be being used as a dummy value for `HEROKU_APP_NAME` here.

To give an idea of what better looks like, I would have gone for something like `{ "name" : "HEROKU_APP_NAME", "value" : "${var.service_name}" } # TODO: remove this temporary workaround once we have a production solution for serving static assets.`

fredericfran-gds:
Only the presence of env variable matters in this case.

sengi:
> Only the presence of env variable matters in this case.

That is only true with regard to how it's interpreted by the machine, not how it's interpreted by humans. Code is first and foremost a piece of communication between people. Things like this are important to maintainability.

PR #70, opened 2020-11-20
Add security group rule to access ec2 mongodb

fredericfran-gds:
This PR adds the security group rule for Content-Store to access the
ec2 mongodb.

### DIFF
--- a/terraform/deployments/govuk-test/main.tf
+++ b/terraform/deployments/govuk-test/main.tf
@@ -30,6 +30,10 @@ data "aws_security_group" "redis" {
   name = "govuk_backend-redis_access"
 }
 
+data "aws_security_group" "govuk_mongo_access" {
+  name = "govuk_mongo_access"
+}
+
 data "terraform_remote_state" "infra_networking" {
   backend = "s3"
   config = {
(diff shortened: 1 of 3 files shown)


terraform/deployments/govuk-test/main.tf
```diff
@@ -30,6 +30,10 @@ data "aws_security_group" "redis" {
   name = "govuk_backend-redis_access"
 }
 
+data "aws_security_group" "govuk_mongo_access" {
```
sengi, line 33: nit: If we follow the convention of the ones we have so far, the logical name for this data resource would be "mongodb".

```suggestion
data "aws_security_group" "mongodb" {
```

bilbof (approved):

terraform/modules/govuk/variables.tf
```diff
@@ -40,6 +40,11 @@ variable "documentdb_security_group_id" {
   type        = string
 }
 
+variable "mongodb_security_group_id" {
```
richardTowers, line 43: Is there just one mongodb instance? Just checking if this name is future proof

fredericfran-gds: there is only one, the other one is called `router_backend`, at least in gov.uk speak

richardTowers: Pretty sure it is, since the name is govuk_mongo_access.

terraform/deployments/govuk-test/main.tf
fredericfran-gds: I based it on `govuk_management_access`, damn! Nest time.

terraform/modules/govuk/variables.tf
sengi: This one's the general/shared mongodb so I think it's probably ok. We could rename it to shared_mongodb (I think we used that pattern for shared_documentdb so would make sense.)

PR #71, opened 2020-11-20
Make it possible to publish and view edition in Publisher app

bilbof:
Publisher can now make requests to https://frontend.test.govuk.digital/bank-holidays.json, which is necessary for viewing and publishing editions.

More work is required in order to make these editions viewable via the frontend app.


fredericfran-gds (approved):
great work!!!

bilbof:
Thanks! I learnt some new Terraform with this change :)

PR #73, opened 2020-11-20
Use local.service_name in signon task definition

bilbof:
var.service_name is not defined, so we can use a local here.


fredericfran-gds (approved):
Thanks for fixing this.

PR #74, opened 2020-11-20
Read security groups from remote state

richardTowers:
Similar to https://github.com/alphagov/govuk-infrastructure/pull/59

The terraform module in govuk-aws that manages the security groups
already exports all the ones we need:

https://github.com/alphagov/govuk-aws/blob/master/terraform/projects/infra-security-groups/outputs.tf

This isn't a major win, as looking them up by their names ought to be
reasonably stable. But it feels a bit cleaner to me to read from the
upstream terraform statefile instead of from AWS.

### DIFF
--- a/terraform/deployments/govuk-test/main.tf
+++ b/terraform/deployments/govuk-test/main.tf
@@ -18,27 +18,20 @@ provider "aws" {
   region = "eu-west-1"
 }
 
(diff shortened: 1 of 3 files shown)


bilbof (approved):
Nice 👍

sengi:
Thanks @richardTowers - this is much better!

PR #75, opened 2020-11-20
Move concourse IAM config into its own file

richardTowers:
The concourse IAM stuff hasn't got anything to do with the ECS
cluster, so it shouldn't be in that file.

This should be a noop, as I'm just moving config from one file to
another. The only thing to review is the comments.


sengi (approved):
Thanks!

PR #77, opened 2020-11-23
Bump Envoy version

bilbof:
This is the latest version of Envoy available from AWS:

https://docs.aws.amazon.com/app-mesh/latest/userguide/envoy.html


seb-szy (approved):
All good, please go ahead.

PR #79, opened 2020-11-23
Various minor changes pending from previous work

fredericfran-gds:
### Replace `APPMESH_VIRTUAL_NODE_NAME` by `APPMESH_RESOURCE_ARN` (2022e19)

With envoy image: `1.15.0 or later`, we should use `APPMESH_RESOURCE_ARN`
configuration parameter rather than the deprecated
`APPMESH_VIRTUAL_NODE_NAME`.

Ref:
1. [AWS Envoy Docs](https://docs.aws.amazon.com/app-mesh/latest/userguide/envoy.html)

### Add `APPMESH_RESOURCE_ARN` parameter (8f93f49)

Add `APPMESH_RESOURCE_ARN` parameter to missing:
1. router
2. router-api
3. static

### Add TODO for assets workaround (9dac450)

We defined some extra parameters for the Frontend and Static App to serve  static assets directly rather than through nginx. We need to record that these need to be remove once we have a production ready solution to this issue.

Ref:
1. #58

### DIFF
--- a/terraform/modules/task-definitions/content-store/main.tf
+++ b/terraform/modules/task-definitions/content-store/main.tf
@@ -49,7 +49,7 @@ module "task_definition" {
       "image" : "govuk/content-store:${var.image_tag}", # TODO: replace temporary label
       "essential" : true,
       "environment" : [
-        { "name" : "APPMESH_VIRTUAL_NODE_NAME", "value" : "mesh/${var.mesh_name}/virtualNode/${var.service_name}" },
+        { "name" : "APPMESH_RESOURCE_ARN", "value" : "mesh/${var.mesh_name}/virtualNode/${var.service_name}" },
         { "name" : "DEFAULT_TTL", "value" : "1800" },
         { "name" : "GOVUK_APP_DOMAIN", "value" : var.service_discovery_namespace_name },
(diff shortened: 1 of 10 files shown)


bilbof (approved):

terraform/modules/task-definitions/content-store/main.tf
```diff
@@ -49,7 +49,7 @@ module "task_definition" {
       "image" : "govuk/content-store:${var.image_tag}", # TODO: replace temporary label
       "essential" : true,
       "environment" : [
-        { "name" : "APPMESH_VIRTUAL_NODE_NAME", "value" : "mesh/${var.mesh_name}/virtualNode/${var.service_name}" },
+        { "name" : "APPMESH_RESOURCE_ARN", "value" : "mesh/${var.mesh_name}/virtualNode/${var.service_name}" },
```
sengi, line 52: Does this actually need to be on the app containers in the first place? I'm fairly sure it's only used by Envoy.

fredericfran-gds: I replaced it everywhere based on another review, I can delete them if not needed.

PR #80, opened 2020-11-23
remove superfluous APPMESH_RESOURCE_ARN

fredericfran-gds:
We remove superflous APPMESH_RESOURCE_ARN from the application
containers themselves as they are needed only for the envoy container
inside the same task.

Ref:
1. [AWS Envoy Docs](https://docs.aws.amazon.com/app-mesh/latest/userguide/envoy.html)
2. #79


sengi (approved):
Thanks for cleaning this up.

PR #81, opened 2020-11-23
Add deployment for draft content store

bilbof:
This makes it possible to deploy the draft content store.

Further work is required to deploy this application from Concourse.

### DIFF
--- a/terraform/deployments/apps/content-store/main.tf
+++ b/terraform/deployments/apps/content-store/main.tf
@@ -27,7 +27,7 @@ module "task_definition" {
   service_name                     = "content-store"
   govuk_app_domain_external        = var.app_domain
   govuk_website_root               = local.website_root
-  image_tag                        = var.image_tag
+  image_tag                        = "bill-content-schemas" # TODO: Change back once content schemas are available
   mesh_name                        = var.mesh_name
   mongodb_url                      = "mongodb://${var.mongodb_host}/content_store_production"
   service_discovery_namespace_name = local.service_discovery_namespace_name
--- a/terraform/modules/task-definitions/content-store/variables.tf
+++ b/terraform/modules/task-definitions/content-store/variables.tf
@@ -6,6 +6,12 @@ variable "govuk_website_root" {
   type = string
 }
 
+variable "router_api_hostname_prefix" {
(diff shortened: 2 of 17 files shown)


terraform/deployments/apps/content-store/main.tf
```diff
@@ -30,6 +30,7 @@ module "task_definition" {
   image_tag                        = var.image_tag
   mesh_name                        = var.mesh_name
   mongodb_url                      = "mongodb://${var.mongodb_host}/content_store_production"
+  plek_hostname_prefix             = ""
```
fredericfran-gds, line 33: is this intended to be empty?

bilbof: Yep!

fredericfran-gds (approved):
Thanks

terraform/deployments/apps/content-store/main.tf
sengi: Might be slightly nicer to have it default to empty in variables.tf? Would avoid the surprise that way. This looks slightly like it's overriding something to empty that might not otherwise have been empty.

terraform/modules/task-definitions/content-store/variables.tf
```diff
@@ -19,6 +19,10 @@ variable "mongodb_url" {
   type = string
 }
 
+variable "plek_hostname_prefix" {
```
sengi, line 22: This would probably benefit from a description mentioning that it's primarily for prepending the "draft-" to the names of its dependencies.

In the long run we might be better off not using this feature of Plek and just overriding the specific dependencies which need to change, rather than having Plek prepend "draft-" to everything and then having to specify exceptions to that.

terraform/deployments/apps/content-store/main.tf
bilbof: Thanks, I've done that 👍

terraform/modules/task-definitions/content-store/variables.tf
bilbof: Thanks, I've made this name more specific.

sengi (approved):

PR #85, opened 2020-11-27
Fix word splitting if APPLICATION contains spaces

richardTowers:
This is unlikely to happen in practice, but it's good practice to quote
variables so that word splitting doesn't happen in unexpected places.

At some point we should set up some CI to run shellcheck on these
snippets (e.g. using
[pipecleaner](https://github.com/alphagov/paas-cf/tree/master/tools/pipecleaner))


sengi (approved):
Thanks!

PR #86, opened 2020-11-27
Add Publishing API Sidekiq as an ECS Service

bilbof:
Paired with @MahmudH.

This adds Publishing API's worker process as an additional ECS Service.

The approach follows the [design for running workers in ECS](https://docs.google.com/document/d/1XyzUHH-M1xxH7SMhRYyoJqxjZHOHeD2x-ZSX11EHh_k/edit#).

We also rename the publishing-api service to publishing-api-web, which we may want to do for all services.

Card: https://trello.com/c/dsIE7CEs/300-add-worker-service-for-publishing-api


terraform/modules/task-definitions/publishing-api/main.tf
```diff
@@ -16,7 +16,7 @@ provider "aws" {
 }
 
 locals {
-  service_name = "publishing-api"
+  app_name = "publishing-api"
```
sengi, line 19: Not something to worry about in this PR but I feel like the semantics of service_name vs app_name vs mesh_subdomain etc. are getting pretty confusing, especially now that we've subtly changed the meaning of service_name by adding the -web suffix to the main service.

Maybe we could refactor later to pass in app_name and app_name_suffix (or something) and compute service_name - I don't know.

sengi (approved):
👍

terraform/modules/task-definitions/publishing-api/main.tf
bilbof: This config is a result of the publishing-api-web service tasks being reachable via `publishing-api.mesh.govuk-internal.digital` rather than `publishing-api-web...` which doesn't seem valuable, I'll remove that logic in a subsequent PR.

PR #87, opened 2020-11-27
Add Redis Cluster

fredericfran-gds:
We add a dedicated a Redis cluster for our ECS in addition to the
existing one in the EC2 GOV.UK platform. This will avoid crosstalk
between some apps running in both ECS and EC2 and writing to the same
databases.

Ref:
1. [task trello card](https://trello.com/c/yzWwXCtK/302-create-a-redis-cluster-for-apps)


terraform/modules/redis/variables.tf
```diff
@@ -0,0 +1,27 @@
+variable "cluster_name" {
+  description = "name of the cluster"
+  type        = string
+  default     = "shared"
+}
+
+variable "internal_domain_name" {
+  description = "The internal root root domain name"
```
sengi, line 8: "root" is repeated here. Might as well just replace it the better description from above:

```suggestion
description = "Domain in which to create DNS records for private resources. For example, test.govuk-internal.digital"
```

terraform/modules/redis/variables.tf
```diff
@@ -0,0 +1,27 @@
+variable "cluster_name" {
+  description = "name of the cluster"
```
sengi, line 2: This description doesn't add anything (and isn't correctly punctuated to boot). I'd suggest either deleting it or at least something like

```suggestion
  description = "Name of Redis cluster to be created."
```

terraform/modules/redis/main.tf
```diff
@@ -0,0 +1,54 @@
+terraform {
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = ">= 3.0"
+    }
+  }
+}
+
+/// elasticache cluster
```
sengi, line 10: This comment doesn't really add anything - please consider removing.

terraform/modules/redis/main.tf
```diff
@@ -0,0 +1,54 @@
...
+
+resource "aws_elasticache_subnet_group" "redis_cluster_subnet_group" {
+  name       = var.cluster_name
+  subnet_ids = var.subnet_ids
+}
+
+resource "aws_security_group" "redis" {
+  name        = "${var.cluster_name}_elasticache"
```
sengi, line 18: Do you have a plan for how this will work with multiple test environments?

terraform/modules/redis/main.tf
```diff
@@ -0,0 +1,54 @@
...
+  vpc_id      = var.vpc_id
+  description = "${var.cluster_name} elasticache cluster"
```
sengi, line 20: This description doesn't add anything - consider removing.

sengi (commented):
Looks basically ok, just a couple of questions and some trivial suggestions.

The PR description seems to be blank (which is odd because there's only one commit and the commit has a description which would do just fine). Could you copy that in please?

terraform/modules/redis/variables.tf
fredericfran-gds: will do

terraform/modules/redis/variables.tf
fredericfran-gds: ok

terraform/modules/redis/main.tf
fredericfran-gds: ok

terraform/modules/redis/main.tf
fredericfran-gds: are you referring to the workspaces work? This is not merge yet.

terraform/modules/redis/main.tf
fredericfran-gds: I will remove

terraform/modules/redis/main.tf
```diff
@@ -0,0 +1,54 @@
...
+  name         = var.internal_domain_name
+  private_zone = true
+}
+
+resource "aws_route53_record" "internal_service_record" {
+  zone_id = data.aws_route53_zone.internal.zone_id
+  name    = "${var.cluster_name}-redis.${var.internal_domain_name}"
+  type    = "CNAME"
+  ttl     = 300
+  records = [join("", aws_elasticache_replication_group.redis_cluster.*.configuration_endpoint_address)]
```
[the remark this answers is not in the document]
fredericfran-gds: Ummm... I copied the code from govuk-aws, I modified that now.

terraform/modules/redis/main.tf
sengi: It doesn't matter whether that work is merged yet. The reason I'm asking is because your colleagues are working on enabling multiple test environments per account, so it would be considerate to think about whether your PR is compatible with that.

I think all you need to do to ensure that this plays nicely with multiple copies of the infrastructure is to remove the default from `cluster_name` in the Redis module and pass it in from the govuk module. Someone can then easily add the environment name to it later once there's a variable for that.

terraform/modules/redis/main.tf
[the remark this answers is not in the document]
sengi: Ah I see, aws_elasticache_replication_group.redis_cluster is singular after all. Presumably the old code was some obfuscated TF 0.11 workaround then?

bilbof (approved):
LGTM! Thank you for this. 👍  

Chris's point about making sure this'll work with multiple clusters makes sense to me. I think you're right that it's worth having this within the `govuk` module even though it's a datastore.

terraform/modules/redis/main.tf
fredericfran-gds: I will discuss with Roch, I guess I need to pull the he cluster name variable up to deployment level.

PR #89, opened 2020-12-01
Fix Redis cluster

fredericfran-gds:
This fixes #87 by turning off clustering mode for the Redis cluster
because some apps are not compatible/configured with that feature.

We also add a security group to allow traffic to exit the Redis cluster.

Ref:
1. [aws docs](https://aws.amazon.com/blogs/database/work-with-cluster-mode-on-amazon-elasticache-for-redis/)


sengi (approved):
Out of interest, what does the Redis cluster need initiate outward connections to? Is it just for the replication between the two nodes?

fredericfran-gds:
maybe we don't need it? I was getting some connections errors, might be just the DNS changeover which was the cause.

sengi:
Ha ok. Thanks for clarifying. Please do remove it if you find it's not needed after all.

PR #90, opened 2020-12-01
Refactor mesh subdomain

MahmudH:
The ECS services; publishing-api and publisher have DNS subdomains and mesh_subdomains. To make these variables easier to understand we have changed the internal DNS record to the service name (for example, `publishing-api-web.mesh`) 

Trello: https://trello.com/c/hFvyUMeC/312-change-web-process-subdomains-to-include-web-suffix

### DIFF
--- a/terraform/deployments/apps/publishing-api-web/main.tf
+++ b/terraform/deployments/apps/publishing-api-web/main.tf
@@ -29,7 +29,6 @@ module "task_definition" {
   govuk_website_root               = local.website_root
   image_tag                        = "bilbof_dockerfile-fix" # TODO: Pinned due to content-schemas issue
   mesh_name                        = var.mesh_name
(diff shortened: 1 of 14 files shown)


richardTowers (approved):
I think this is a slight improvement - it was confusing having both `service_name` and `mesh_subdomain`, and then coalescing them.

On the other hand, I think that having to put `-web` in our DNS records is a bit of a smell - the workers shouldn't need DNS entries at all, so there shouldn't be any need to disambiguate `blah-web.mesh.govuk-internal.digital` from `blah-worker.mesh.govuk-internal.digital`.

My suggestion is that we merge this for now, but we should consider dividing up `modules/app` into `modules/app-web` and `modules/app-worker` so that it's possible to create workers without creating a bunch of appmesh / cloudmap config which is only necessary for the web instances.

PR #95, opened 2020-12-03
Configure Signon in ECS

bilbof:
This makes two changes:

- configures Signon in ECS so users can sign in with it at https://signon-ecs.test.govuk.digital
- configures other apps to use Signon for app-to-app authorization

I've also introduced a public ALB module to reduce duplication.

https://trello.com/c/RpigpQpb/303-get-signon-in-ecs-working


terraform/modules/public-load-balancer/main.tf
```diff
@@ -0,0 +1,79 @@
...
+  port        = 80
+  protocol    = "HTTP"
+  vpc_id      = var.vpc_id
+  target_type = "ip"
+
+  health_check {
+    path = "/healthcheck"
+  }
+
+  depends_on = [aws_lb.public]
```
sengi, line 45: I have a feeling this might still needed but I can't remember for sure. Is there a Terraform bug we can link to? I did a quick search and couldn't find anything that fits this scenario. (Don't worry if not, just would be good to avoid mysterious-looking workarounds as they tend to then proliferate without being questioned.)

terraform/modules/public-load-balancer/variables.tf
```diff
@@ -0,0 +1,38 @@
...
+}
+
+variable "app_name" {
+  type        = string
+  description = "A GOV.UK application name. E.g. publisher, content-publisher"
+}
+
+variable "cluster_name" {
+  type    = string
+  default = "govuk" # TODO: Is this the default value?
```
sengi, line 13: I think calling this `cluster_name` could be confusing, because this module shouldn't have anything to do with the ECS cluster. Since this is really a suffix for the name of the ALB that the module creates, we could call it that and add a description saying what it does. I don't know if we settled on terminology for this yet but `workspace_suffix` maybe? (or `name_suffix`?)

I appreciate this whole environment naming thing is a bit of a can of worms though - no worries if we can't solve this neatly in this PR. Might need some more thought about the broader issue of how we pass these names around in a way which is sufficiently clear as to what's what.

terraform/modules/public-load-balancer/security_groups.tf
```diff
@@ -0,0 +1,43 @@
...
+  name        = "fargate_${var.app_name}_${var.cluster_name}_public_alb"
+  vpc_id      = var.vpc_id
+  description = "${var.app_name} Internet-facing ALB in ${var.cluster_name} cluster"
+}
+
+#
+# Security group rules
+#
+
+resource "aws_security_group_rule" "service_from_alb_http" {
```
sengi, line 15: Cool this should cut out some repetitive nonsense from `govuk/security_group_rules.tf`. Thanks!

terraform/modules/task-definitions/content-store/main.tf
```diff
@@ -64,6 +64,7 @@ module "task_definition" {
         { "name" : "PLEK_SERVICE_PUBLISHING_API_URI", "value" : "http://publishing-api-web.${var.service_discovery_namespace_name}" },
         { "name" : "PLEK_SERVICE_ROUTER_API_URI", "value" : "http://${var.router_api_hostname_prefix}router-api.${var.service_discovery_namespace_name}" },
         { "name" : "PLEK_SERVICE_RUMMAGER_URI", "value" : "" },
+        { "name" : "PLEK_SERVICE_SIGNON_URI", "value" : "https://signon-ecs.${var.govuk_app_domain_external}" },
```
sengi, line 67: I'm guessing the change of user-facing name is so that the old and new can coexist? This is probably something we'll need to think about a bit more in terms of how we'd switch things over. Ok for now though.

terraform/modules/task-definitions/signon/main.tf
```diff
@@ -71,9 +89,25 @@ module "task_definition" {
...
+        {
+          "name" : "SECRET_KEY_BASE",
+          "valueFrom" : data.aws_secretsmanager_secret.secret_key_base.arn
+        },
         {
           "name" : "SENTRY_DSN",
           "valueFrom" : data.aws_secretsmanager_secret.sentry_dsn.arn
+        },
+        {
+          "name" : "DATABASE_URL",
```
sengi, line 101: No need to fix this here but maybe add a TODO or a card - database_url shouldn't be a secret (only the password should be).

sengi (approved):
Cool, thanks for sorting out the LB module.

terraform/modules/task-definitions/signon/main.tf
bilbof: I agree - this is something we'll need to fix in the app.

terraform/modules/public-load-balancer/main.tf
bilbof: I copied it over from Publisher - I've removed it since I also don't know the reason for it.

terraform/modules/public-load-balancer/main.tf
```diff
@@ -0,0 +1,79 @@
...
+  listener_arn    = aws_lb_listener.public.arn
+  certificate_arn = data.aws_acm_certificate.public_lb_alternate.arn
+}
+
+#
+# Internet-facing load balancer
+#
+
+resource "aws_lb" "public" {
+  name               = "fargate-public-${var.app_name}-${var.cluster_name}"
```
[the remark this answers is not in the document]
bilbof: Thanks for the heads up, I've removed `fargate-`.

terraform/modules/public-load-balancer/variables.tf
bilbof: Thanks, I've changed it to `workspace_suffix` for now. I agree, we may want to revisit this once Roch finishes the multi-cluster work.

terraform/modules/task-definitions/content-store/main.tf
bilbof: Yes, I don't think we've worked this out yet.

terraform/modules/task-definitions/signon/main.tf
bilbof: I added a card for this https://trello.com/c/ahHLtryi/119

PR #96, opened 2020-12-03
CD terraform

fredericfran-gds:
This PR attempts to terraform apply the `govuk` base infrastructure and conditionally update each service deployments if their respective task definition has changed compared to what is currently running in the ECS cluster.

Ref:
1. [trello card](https://trello.com/c/jNYME2fN/305-continuously-deploy-the-govuk-terraform-deployment)

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -206,6 +206,9 @@ jobs:
   - name: deploy-frontend
     plan:
     - get: govuk-infrastructure
+      passed:
+      - run-terraform
+      trigger: true
     - get: release
       resource: frontend
       trigger: true
@@ -245,6 +248,9 @@ jobs:
   - name: deploy-draft-frontend
     plan:
     - get: govuk-infrastructure
+      passed:
+      - run-terraform
+      trigger: true
     - get: release
       resource: frontend
       trigger: true
@@ -273,6 +279,9 @@ jobs:
   - name: deploy-publisher
     plan:
     - get: govuk-infrastructure
+      passed:
+      - run-terraform
+      trigger: true
     - get: release
       resource: publisher
       trigger: true
@@ -313,6 +322,9 @@ jobs:
   - name: deploy-publishing-api
     plan:
     - get: govuk-infrastructure
+      passed:
+      - run-terraform
+      trigger: true
     - get: release
       resource: publishing-api
       trigger: true
@@ -343,6 +355,9 @@ jobs:
   - name: deploy-content-store
     plan:
     - get: govuk-infrastructure
+      passed:
+      - run-terraform
+      trigger: true
     - get: release
       resource: content-store
       trigger: true
@@ -363,6 +378,9 @@ jobs:
   - name: deploy-draft-content-store
     plan:
     - get: govuk-infrastructure
+      passed:
+      - run-terraform
+      trigger: true
     - get: release
       resource: content-store
       trigger: true
@@ -383,6 +401,9 @@ jobs:
   - name: deploy-router
     plan:
     - get: govuk-infrastructure
+      passed:
+      - run-terraform
+      trigger: true
     - get: release
       resource: router
       trigger: true
@@ -403,6 +424,9 @@ jobs:
   - name: deploy-draft-router
     plan:
     - get: govuk-infrastructure
+      passed:
+      - run-terraform
+      trigger: true
     - get: release
       resource: router
       trigger: true
@@ -423,6 +447,9 @@ jobs:
   - name: deploy-router-api
     plan:
     - get: govuk-infrastructure
+      passed:
+      - run-terraform
+      trigger: true
     - get: release
       resource: router-api
       trigger: true
@@ -443,6 +470,9 @@ jobs:
   - name: deploy-draft-router-api
     plan:
     - get: govuk-infrastructure
+      passed:
+      - run-terraform
+      trigger: true
     - get: release
       resource: router-api
       trigger: true
@@ -463,6 +493,9 @@ jobs:
   - name: deploy-signon
     plan:
     - get: govuk-infrastructure
+      passed:
+      - run-terraform
+      trigger: true
     - get: release
       resource: signon
       trigger: true
@@ -483,6 +516,9 @@ jobs:
   - name: deploy-static
     plan:
     - get: govuk-infrastructure
+      passed:
+      - run-terraform
+      trigger: true
     - get: release
       resource: static
       trigger: true
@@ -503,6 +539,9 @@ jobs:
   - name: deploy-draft-static
     plan:
     - get: govuk-infrastructure
+      passed:
+      - run-terraform
+      trigger: true
     - get: release
       resource: static
       trigger: true
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -15,7 +15,7 @@ resources:
     icon: github
     name: govuk-infrastructure
     source:
-      branch: main
+      branch: cd-terraform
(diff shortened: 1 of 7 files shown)


richardTowers:
This looks pretty good to me - only thing I'd suggest is removing the `trigger: true` bits - I don't think we actually need to trigger the app deployments when terraform succeeds, they just need the `passed: [ run-terraform]` bit.

fredericfran-gds:
cool, let me test that...

fredericfran-gds:
I needed without the trigger, doesn't work

sengi (commented):
Cool. Am I right in thinking this will trigger a redeploy of all apps on any Terraform change?

Which scenarios did you test with this?

Would be good to document what runs when (feel free to do that in another PR if you want). Probably in docs/applying-terraform.md?

sengi:
Have I got the right place for testing? https://cd.gds-reliability.engineering/teams/govuk-tools/pipelines/deploy-apps-test Is it deployed there at the moment?

concourse/tasks/update-ecs-service.yml
```diff
@@ -28,14 +28,30 @@ run:
...
+      new_task_definition_arn="$(cat "terraform-outputs/${APPLICATION}_task_definition_arn")"
+      if [ -z "${new_task_definition_arn}"]; then
+        echo "failed to retrieve new task definition for ${APPLICATION}, exiting..."
+        exit 1
+      fi
+
+      if [ "${current_task_definition_arn}" = "${new_task_definition_arn}" ]; then
+        echo "No need to update ${APPLICATION} service since its task definition was not updated"
+        exit 0
+      fi
```
bilbof, line 46: I don't think this bit is necessary, since if the task definition is the same, then `aws ecs update-service` is a no-op.

fredericfran-gds: Not what we experienced yesterday...

bilbof: For example, the ECS Service `content-store` is currently running task definition `content-store:61`.

```
aws ecs update-service --cluster govuk --service content-store --task-definition content-store:61
ecs wait services-stable --cluster govuk --services content-store # will exit immediately.
```

bilbof: If the service was not stable when `ecs update-service` was run, this might cause `ecs awit services-stable` to wait, even if the task definition had not changed.

concourse/pipelines/deploy.yml
```diff
@@ -164,9 +169,46 @@ jobs:
...
+          args:
+          - '-c'
+          - |
+            set -eu
+
+            terraform init -backend-config "role_arn=$ASSUME_ROLE_ARN"
+            terraform apply \
+              -var "assume_role_arn=$ASSUME_ROLE_ARN" \
+              -var-file "../variables/test/common.tfvars" \
+              -var-file "../variables/test/infrastructure.tfvars" \
```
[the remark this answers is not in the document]
fredericfran-gds: I can do that.

fredericfran-gds:
> Have I got the right place for testing? https://cd.gds-reliability.engineering/teams/govuk-tools/pipelines/deploy-apps-test Is it deployed there at the moment?

Yes, but now it is not working because the govuk-infrastructure branch is set to `main` again in the concourse pipeline config, if you set it to `cd-terraform`, this will work.

concourse/tasks/update-ecs-service.yml
fredericfran-gds: after further digging, I agree with you that it should be a no-op under normal circumstances if the task definition is the same. I think we got confused by the command printing the logs of the events which are older ones and not triggered by this change.

I think it is still could that we include the conditional above if you don't have an objection.

fredericfran-gds:
> Cool. Am I right in thinking this will trigger a redeploy of all apps on any Terraform change?
> 
> Which scenarios did you test with this?
> 
> Would be good to document what runs when (feel free to do that in another PR if you want). Probably in docs/applying-terraform.md?

I updated the docs here in this commit:
https://github.com/alphagov/govuk-infrastructure/pull/96/commits/73e201191c7c194ee4766b5ac4b45406edc66fc0

The scenario under which this was tested was just modifying the govuk-infrastructure and letting Concourse detect the change. 

It will first run:
- run-terraform job
- after that, it will run the other deploy-${application} in parallel.
- as Bill mentions, it is a no-op if the task definition is not updated even though it is a bit confusing the output of the `update-ecs-service` task. That's why we added a conditional at the beginning of that task so that `service update` section is not run if the task definition is not changed.

concourse/tasks/update-ecs-service.yml
sengi: If this additional complexity is indeed unnecessary then I have an objection to keeping it, unless there's a clear rationale for doing so.

docs/applying-terraform.md
```diff
@@ -10,19 +10,30 @@ We have attempted - where possible - to separate infrastructure concerns
...
+After the base infrastructure has been terraformed, the Concourse pipeline will create new task definitions
+for any services/applications that have changed. If a new task definition has been created, the respective
+ECS Service will be updated/deployed.
+
+## Local Deployment
+
+### Infrastructure
+
+You can update the base infrastructure from your machine to test things.
+For example, we run the following commands to updating the test environment:
```
sengi, line 28: typo
```suggestion
For example, run the following commands to update the test environment:
```

concourse/tasks/update-ecs-service.yml
sengi: Ah, just saw your explanation in another thread below. So this basically just gets us slightly nicer-looking logs. I think I'm ok with either way then, but if you want to leave it in then please add a comment explaining what it's really for - because otherwise there's no way that somebody reading that shell script later would be able to know that that check/comparison logic isn't essential.

sengi (commented):
Ok so sounds like it's been tested for at least the trivial case of updating something and it running successfully. I guess that's probably ok for now - would have been nice to be able to review the actual working thing but no worries.

Thanks for updating the docs!

Happy to 👍 this once the update-service thing is resolved one way or the other.

concourse/tasks/update-ecs-service.yml
fredericfran-gds: I can do that.

docs/applying-terraform.md
fredericfran-gds: I can do that.

fredericfran-gds:
all done in terms of review fixes.

terraform/deployments/govuk-test/main.tf
```diff
@@ -65,4 +71,4 @@ module "govuk" {
   signon_desired_count              = var.signon_desired_count
   static_desired_count              = var.static_desired_count
   draft_static_desired_count        = var.draft_static_desired_count
-}
+}
```
sengi, line 74: Somehow missing a line ending here apparently - might want to check your editor config.

concourse/pipelines/deploy.yml
```diff
@@ -196,7 +238,6 @@ jobs:
     - get: govuk-infrastructure
       passed:
       - deploy-frontend
-      trigger: true
```
sengi, line 199: I hadn't spotted this before - is it intentional?

sengi (approved):

concourse/pipelines/deploy.yml
fredericfran-gds: ah it was what Richard suggested but I will re-add that because it did not work for the others.

terraform/deployments/govuk-test/main.tf
fredericfran-gds: Wierd. Fixed.

PR #98, opened 2020-12-09
Add a simple AWS billing dashboard

richardTowers:
This is possible now that we've signed in to the root account and
enabled IAM users to view billing information :tada:.

This will hopefully give us a slightly easier-to-access view of our
billing (compared to having to sign in to AWS and look in
cost-explorer).

It's a bit naïve - CloudWatch's EstimatedCharges metric just increases
throughout the month and then resets. Might be nice to try to measure
the gradient, so we can see the rate we're spending at. But this
is good enough for now.

One interesting thing the graphs show is that the cost of the `test`
environment is still dominated by EC2 and RDS, even though we have quite
a bit of ECS stuff in there now. Not very scientific, but still.

Tested locally with docker-compose:

image

### DIFF
--- a/grafana/terraform/billing.tf
+++ b/grafana/terraform/billing.tf
@@ -1,0 +1,264 @@
+locals {
+  billing_dashboard_config = {
+    "annotations" : {
+      "list" : [
+        {
+          "builtIn" : 1,
+          "datasource" : "-- Grafana --",
+          "enable" : true,
+          "hide" : true,
+          "iconColor" : "rgba(0, 211, 255, 1)",
+          "name" : "Annotations & Alerts",
+          "type" : "dashboard"
+        }
+      ]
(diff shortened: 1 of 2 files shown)


sengi (approved):
Personally I think the [AWS billing dashboard](https://console.aws.amazon.com/billing/home?#/) presents this info in a much friendlier way and is about the same effort to open. No harm in this being there for now though.

richardTowers:
> Personally I think the [AWS billing dashboard](https://console.aws.amazon.com/billing/home?#/) presents this info in a much friendlier way and is about the same effort to open. No harm in this being there for now though.

True enough. I like having everything in Grafana, but at least it's available in AWS now (prior to @seb-szy's changes this morning we couldn't see those dashboards either).

PR #99, opened 2020-12-09
Use Fargate spot instances in the test environment.

sengi:
Add FARGATE_SPOT as an allowable capacity provider in our ECS cluster and make the default provider configurable.

The default is deliberately kept as on-demand and only the test environment specifies spot, so that prod will have the right setting by default even if this gets forgotten.

Tasks can still override the default capacity provider, so this doesn't affect our ability to do things like have one-off user tasks run as spot instances if we want (though we're not making use of that anywhere yet).

Motivation for this:

- Saves money.
- Will help to surface any potential issues when instances go away (and
  it's important for reliability that we find and fix any such issues). It's free [chaosmonkey](https://netflix.github.io/chaosmonkey/), essentially.
- Unlikely to hinder development (and if it does, it's easy to switch
  off temporarily or even permanently).

Tested: `terraform plan` looks reasonable. Haven't applied it; planning on doing that after merge and reverting if it doesn't work, so as to minimise disruption to other work.

```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  ~ update in-place

Terraform will perform the following actions:

  # module.govuk.aws_ecs_cluster.cluster will be updated in-place
  ~ resource "aws_ecs_cluster" "cluster" {
        arn                = "arn:aws:ecs:eu-west-1:430354129336:cluster/govuk"
      ~ capacity_providers = [
            "FARGATE",
          + "FARGATE_SPOT",
        ]
        id                 = "arn:aws:ecs:eu-west-1:430354129336:cluster/govuk"
        name               = "govuk"
        tags               = {}

      + default_capacity_provider_strategy {
          + capacity_provider = "FARGATE_SPOT"
        }

        setting {
            name  = "containerInsights"
            value = "enabled"
        }
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```


richardTowers (approved):
Nice, looks like it's worth a try, for sure!

PR #101, opened 2020-12-11
Fix shell if statement in update-ecs-service

richardTowers:
I spotted in Frederic's demo that when concourse runs these scripts, it
shows some errors:

```
sh: missing ]

sh: missing ]
```

This is because there's no space between the string we're checking in
the if statement and the closing `]` character.

These don't cause the script to fail, even though it's got `set -e` at
the top because they're part of an if statement.


rtrinque (approved):
Looks good

PR #102, opened 2020-12-11
Create modules for configuring task definitions

richardTowers:
These modules will be used instead of the task-definition and task-definitions/* modules to create ECS Task Definitions for the GOV.UK applications (the so called "small" terraform deployments which are run by concourse).

We've kept these as simple as possible, by only having them take inputs and produce outputs - they don't create any resources in AWS directly. Instead, the deployment will create the resource using the outputs from the modules.

### DIFF
--- a/terraform/modules/app-container-definition/main.tf
+++ b/terraform/modules/app-container-definition/main.tf
@@ -1,0 +1,49 @@
+variable "name" {
+  type = string
+}
+variable "image" {
+  type = string
+}
+variable "environment_variables" {
+  type = map
+}
+variable "log_group" {
+  type = string
+}
+variable "secrets_from_arns" {
+  type = map
+}
+variable "aws_region" {
+  type = string
+}
+output "value" {
+  value = {
+    "name" : var.name,
+    "image" : var.image,
+    "essential" : true,
+    "environment" : [for key, value in var.environment_variables : { name : key, value : value }],
+    "dependsOn" : [{
+      "containerName" : "envoy",
+      "condition" : "START"
+    }],
+    "logConfiguration" : {
+      "logDriver" : "awslogs",
+      "options" : {
+        "awslogs-create-group" : "true",
+        "awslogs-group" : var.log_group,
+        "awslogs-region" : var.aws_region,
+        "awslogs-stream-prefix" : var.name,
+      }
+    },
+    "mountPoints" : [],
+    "portMappings" : [
+      {
+        "containerPort" : 80,
+        "hostPort" : 80,
+        "protocol" : "tcp"
+      }
+    ],
+    "secrets" : [for key, value in var.secrets_from_arns : { name : key, valueFrom : value }]
+  }
+}
+
--- a/terraform/modules/app-container-definition/main.tf
+++ b/terraform/modules/app-container-definition/main.tf
@@ -16,16 +16,19 @@ variable "secrets_from_arns" {
 variable "aws_region" {
   type = string
 }
+variable "depends_on" {
+  type        = map
+  default     = {}
+  description = "Other containers which this depends on (e.g. for envoy, set this to {envoy: \"START\"})"
+}
+
 output "value" {
   value = {
     "name" : var.name,
     "image" : var.image,
     "essential" : true,
     "environment" : [for key, value in var.environment_variables : { name : key, value : value }],
-    "dependsOn" : [{
-      "containerName" : "envoy",
-      "condition" : "START"
-    }],
+    "dependsOn" : [for key, value in var.depends_on : { containerName : key, condition : value }],
     "logConfiguration" : {
       "logDriver" : "awslogs",
       "options" : {
--- a/terraform/modules/envoy-configuration/main.tf
+++ b/terraform/modules/envoy-configuration/main.tf
@@ -1,0 +1,77 @@
+# See the user guide at
+# https://docs.aws.amazon.com/app-mesh/latest/userguide/app-mesh-ug.pdf
+# for more details on configuring Envoy in AppMesh
+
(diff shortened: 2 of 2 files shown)


terraform/modules/app-container-definition/main.tf
```diff
@@ -0,0 +1,49 @@
...
+output "value" {
+  value = {
+    "name" : var.name,
+    "image" : var.image,
+    "essential" : true,
+    "environment" : [for key, value in var.environment_variables : { name : key, value : value }],
+    "dependsOn" : [{
+      "containerName" : "envoy",
+      "condition" : "START"
+    }],
```
sengi, line 28: I think we'll need to parameterise this, otherwise this module is hard-coupled to the envoy one (and we'll shortly have a use case for a task that doesn't use Envoy, when we set up Grafana). Perhaps a map `depends_on` of containerName to condition, something like this?
```suggestion
    "dependsOn" : [for key, value in var.depends_on : { containerName : key, condition : value }],
```

terraform/modules/app-container-definition/main.tf
```diff
@@ -0,0 +1,49 @@
...
+      }
+    },
+    "mountPoints" : [],
+    "portMappings" : [
+      {
+        "containerPort" : 80,
+        "hostPort" : 80,
+        "protocol" : "tcp"
+      }
+    ],
```
sengi, line 45: Suspect we might end up needing the ability to pass these in a well, but I think hardcoding 80 for now is ok — let's see how it goes. Maybe it'll discourage people from doing whacky things with the port mappings this time 😛

terraform/modules/app-container-definition/main.tf
```diff
@@ -0,0 +1,49 @@
...
+}
+variable "aws_region" {
+  type = string
+}
+output "value" {
+  value = {
+    "name" : var.name,
+    "image" : var.image,
+    "essential" : true,
+    "environment" : [for key, value in var.environment_variables : { name : key, value : value }],
```
sengi, line 24: This is a nice improvement over the terraform-provider-aws API! Good stuff 👍

terraform/modules/envoy-configuration/main.tf
```diff
@@ -0,0 +1,77 @@
...
+variable "log_group" {
+  type = string
+}
+
+variable "aws_region" {
+  type = string
+}
+
+locals {
+  # From the user guide: Our examples use 1337 for historical purposes.
```
sengi, line 23: I see this is indeed a direct quote from the App Mesh user guide, but the malapropism _historical purposes_ is a bit confusing (at least to my mind). They presumably meant historical _reasons_. How about
```suggestion
  # 1337 is an arbitrary choice copied from the examples in the user guide.
```

sengi (approved):

terraform/modules/app-container-definition/main.tf
richardTowers: I'm inclined to leave this hardcoded for now, on the basis that (at the moment) the intended use case for this module is for the gov.uk apps.

There are things like Grafana which will also need task definitions (as you say), but perhaps they're different / unique / simple enough that we could inline their container definitions when we create their resources, instead of trying to make this module totally generic.

richardTowers: I do agree that the coupling between this and the envoy module is unpleasant though 🤔

richardTowers: How do you feel about this?

```
    "dependsOn" : (var.depends_on_envoy
      ? [{ "containerName" : "envoy", "condition" : "START" }]
      : []
    ),
```

Where `depends_on_envoy` is a bool with no default, so apps would have to say `depends_on_envoy = true` when calling the module.

The alternatives I can think of have drawbacks:

* `depends_on` variable with a default of `{envoy: "START"}` - having the default makes this module _mainly_ for apps in the mesh, and other things (not in the mesh) would have to say `depends_on: {}` to opt out of envoy, which feels weird.
* `depends_on` variable with no default - means every app in the mesh has to pass the exact same `depends_on: {envoy: "START"}` param, which is an annoying bit of duplication, and feels a bit like an abstraction leak (although to be fair, they have to call the envoy settings module, so perhaps that leak has already leaked).

One other thing we could consider is leaning in to the coupling and combining the app-container-definition module and the envoy-settings module. I think we discussed that and didn't like it, because it's less flexible. I'm not 100% sure the flexibility is a good thing though.

terraform/modules/app-container-definition/main.tf
richardTowers: Yeah, I think these also have to match the envoy configuration:

https://github.com/alphagov/govuk-infrastructure/pull/102/files#diff-21c98290833886dad6304137790d6ef2775d8ed580f91b4fe77690c380d762f2R52

... so there's a bit more coupling there. Hopefully we won't have to make this configurable for the GOV.UK apps. 

Router is a possible exception, since that wants to listen on two ports at once at the moment - maybe we can fix that when we change how it loads its config. For now, we can always use `merge()` on the output of this module if we really want to customise something in a one-off cast.

terraform/modules/app-container-definition/main.tf
sengi: That's a good point about the default being a bit awkward either way. I think the way to do it would be the second option (empty default and explicit dependency on envoy). Combining the two modules into an opinionated module for the GOV.UK use case is starting to sound more attractive. I'd be fine with either of those tbh.

richardTowers: Implemented as an empty default (so apps will need to be explicit about their dependency on envoy) in e4ee6de.

PR #103, opened 2020-12-11
Simplify tools

richardTowers:
I think we can get rid of a couple of wrapper shell scripts which don't do very much.

My preference is to call tools as directly as possible, and only wrap them in scripts if absolutely necessary.


bilbof (approved):
Nice ⭐

PR #104, opened 2020-12-11
Run terraform validate in a GitHub Action

richardTowers:
https://www.terraform.io/docs/commands/validate.html

> The terraform validate command validates the configuration files in a directory, referring only to the configuration and not accessing any remote services such as remote state, provider APIs, etc.

> Validate runs checks that verify whether a configuration is syntactically valid and internally consistent, regardless of any provided variables or existing state. It is thus primarily useful for general verification of reusable modules, including correctness of attribute names and value types.

> It is safe to run this command automatically, for example as a post-save check in a text editor or as a test step for a re-usable module in a CI system.

I think this is the most terraform validation we can do in GitHub
without giving it actual access to our AWS account to read statefiles
and run plans.

We could also set up a concourse job to run plans against our test AWS
account though, since concourse has already got the AWS access it would
need.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
+      - name: get date
+        id: get-date
+        run: |
+          echo "::set-output name=date::$(/bin/date -u "+%Y%m%d")"
+
       - name: cache terraform plugins
         uses: actions/cache@v2
         env:
@@ -13,10 +18,7 @@ jobs:
         with:
           path: ~/.terraform/plugin-cache
           key:
-            ${{ runner.os }}-${{ env.cache-name }}-v1-${{ hashFiles('terraform/deployments/**/main.tf') }}
-          restore-keys: |
-            ${{ runner.os }}-${{ env.cache-name }}-v1-${{ hashFiles('terraform/deployments/**/main.tf') }}
-            ${{ runner.os }}-${{ env.cache-name }}-v1-
+            ${{ env.cache-name }}-${{ steps.get-date.outputs.date }}-${{ hashFiles('terraform/deployments/**/main.tf', 'terraform/deployments/**/versions.tf') }}
 
       - name: install terraform
         working-directory: terraform
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,13 +6,50 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: terraform fmt
+      - name: cache terraform plugins
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-terraform-plugins
+        with:
+          path: ~/.terraform/plugin-cache
+          key:
+            ${{ runner.os }}-${{ env.cache-name }}-v1-${{ hashFiles('terraform/deployments/**/main.tf') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ env.cache-name }}-v1-${{ hashFiles('terraform/deployments/**/main.tf') }}
+            ${{ runner.os }}-${{ env.cache-name }}-v1-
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -6,13 +6,50 @@ jobs:
...
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-terraform-plugins
+        with:
+          path: ~/.terraform/plugin-cache
+          key:
+            ${{ runner.os }}-${{ env.cache-name }}-v1-${{ hashFiles('terraform/deployments/**/main.tf') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ env.cache-name }}-v1-${{ hashFiles('terraform/deployments/**/main.tf') }}
+            ${{ runner.os }}-${{ env.cache-name }}-v1-
```
sengi, line 19: If I've understood the docs right, this means the cache keys will contain a hash of the contents of the main.tf's, but on lookup it'll still use a cache from a previous commit if there isn't one with a matching hash. That sounds fine except that I'm not seeing how it'll ever update. The caches get GCed after a week of not being accessed, but that's unlikely to be the case except maybe over the holidays. They don't get GCed based on age, as far as I can tell from the docs.

So in ordinary use, I think this will work as intended until new plugin versions start appearing, at which point the cache will become increasingly ineffective as jobs will continue to use an old cache which won't contain newer plugins, until almost nothing in the cache is relevant any more or until we go a whole week without pushing a commit or raising a PR. That's assuming that Terraform is smart enough to download things which it finds missing from its cache directory (and I have a sneaky suspicion that it's not(!) so maybe it'll even just break once new versions start appearing).

If I've got all that right (probably a big if!), we might want to consider using a datestamp instead of the hash (like in [this example from the docs](https://github.com/actions/cache#creating-a-cache-key)).

sengi (approved):

.github/workflows/ci.yml
richardTowers: Yes, I think that's accurate.

At the moment, our [main.tf](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/govuk-test/main.tf#L12) files specify a version constraint on the provider (like `version = "~> 3.13"`), so the theory was that we'd keep using the cached one until we chose to upgrade to a newer constraint.

It's not really an issue if it's using a slightly out-of-date provider from its cache when it runs `terraform init` or `terraform validate` - neither of these commands really exercise the plugins.

That said, the provider only takes a couple of seconds to download from GitHub's network, so caching them here is maybe unnecessary complexity. The thing that's important is that they share a cache between deployments, so we don't have to download them _n_ times.

I'd be happy to remove this extra cache if you'd prefer?

sengi: It seems a bit of a shame not to preserve the cache between builds, since it'd mean a fair bit of unnecessary traffic and wasted resources. We could just give this a try and see if it breaks. If it were me I'd probably use a datestamp instead of the hash though, because I don't think the hash is actually doing anything useful and that way if it goes wrong at least it'll sort itself out after a day.

The [TF docs](https://www.terraform.io/docs/commands/cli-config.html#provider-plugin-cache) say that it'll download the latest version if it doesn't find it in TF_PLUGIN_CACHE_DIR — it won't use an outdated version.

richardTowers: Ah, I see. It picks a version before it looks at the cache. I thought it would always use the cached version if that one met the constraints. But you're right, the docs say otherwise.

richardTowers: Thanks for spotting this, top reviewing!

I've addressed it in 5053e5c

sengi: So I think we now effectively have separate TF_PLUGIN_CACHE_DIR per changeset (and also per day). That seems fine. Thanks!

PR #105, opened 2020-12-11
Actions: Install terraform with setup-terraform

richardTowers:
Instead of `brew` and `tfenv`.

This should have much better caching than the brew / tfenv combination,
which should speed up our actions a fair bit (from a quick look, it should
save about 50 seconds on each run).

We already have the terraform version in multiple places (concourse,
tfenv, now github actions too). I don't think there's an elegant way to
have a single-source-of-truth for them, so I think the best bet is
probably to write a test to check that they're in sync.


rtrinque (approved):
Looks good

PR #106, opened 2020-12-11
Add rspec tests to check versions are in sync

richardTowers:
We've got a few places popping up where versions of things will need to
be kept in sync. For example, we're using terraform 0.13.3 locally, and
in concourse, and in GitHub Actions. There's no simple way of making all
these tools read from a canonical source, so instead let's just add some
tests that things are in-sync.

I'm using ruby / rspec for these tests, because this is
govuk-infrastructure and govuk developers are familiar with ruby /
rspec. This adds a new set of versions to keep in sync 🙈 (_edit: no
it doesn't, because ruby/setup-ruby is .ruby-version aware_).

I hope that these tests will expand to cover more areas (not just
version syncing) in future.


sengi (approved):
The YAML parsing stuff looks a bit brittle, but it's better than just letting the versions go out of sync so let's go for it 👍 

Ideally I think we wouldn't care about the job or task names and just look for all `image_resource` with a tag that starts `terraform-` (and the version value from all `hashicorp/setup-terraform` steps). That'd be way too much logic to put directly in a test though (I think it'd need a little library with tests of its own, even using something like jsonpath), so I can see why you did it this way.

richardTowers:
Yeah, it's definitely brittle, but at least it's (reasonably) simple.

When we get annoyed by the tests constantly breaking when we rename steps we can think of a nicer way of doing it.

PR #108, opened 2020-12-16
Add `terraform fmt` as a pre-commit script

richardTowers:
This should help reduce the cycle time if you forget to run terraform
fmt before committing something. At the moment, you'd have to wait for
GitHub Actions to catch your mistake. This way, it will remind you
automatically.

Output looks like this:

```
Terraform fmt.................................................Failed
- hook id: terraform_fmt
- files were modified by this hook

main.tf
```

... after which there will be some unstaged changes to main.tf which you
can stage if you're happy with them.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,11 @@
+---
 repos:
--   repo: git@github.com:Yelp/detect-secrets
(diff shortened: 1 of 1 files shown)


sengi (approved):
Thanks!

PR #110, opened 2020-12-21
Add workspace awareness

rtrinque:
Terraform workspace allow us to run terraform on separated workspaces, this functionality is very basic in that it simply creates a new state for the new workspace. To run the same infrastructure code from separate workspaces then you need to add some segregation logic to your code that will interpolate naming variables from the current workspace.

### DIFF
--- a/terraform/deployments/concourse-iam/main.tf
+++ b/terraform/deployments/concourse-iam/main.tf
@@ -27,7 +27,7 @@ provider "aws" {
 }
 
 resource "aws_iam_role" "govuk_concourse_deployer" {
-  name        = "${terraform.workspace == "default" ? "govuk-concourse-deployer" : "govuk-concourse-deployer-${terraform.workspace}"}"
+  name        = "govuk-concourse-deployer"
   description = "Deploys applications to ECS from Concourse"
 
   assume_role_policy = <<EOF
--- a/terraform/deployments/concourse-iam/main.tf
+++ b/terraform/deployments/concourse-iam/main.tf
@@ -27,7 +27,7 @@ provider "aws" {
 }
 
 resource "aws_iam_role" "govuk_concourse_deployer" {
-  name        = "govuk-concourse-deployer"
+  name        = "govuk-concourse-deployer-${terraform.workspace}"
   description = "Deploys applications to ECS from Concourse"
 
   assume_role_policy = <<EOF
--- a/terraform/deployments/concourse-iam/main.tf
+++ b/terraform/deployments/concourse-iam/main.tf
@@ -27,7 +27,7 @@ provider "aws" {
 }
 
 resource "aws_iam_role" "govuk_concourse_deployer" {
-  name        = "govuk-concourse-deployer-${terraform.workspace}"
+  name        = "${terraform.workspace == "default" ? "govuk-concourse-deployer" : "govuk-concourse-deployer-${terraform.workspace}"}"
   description = "Deploys applications to ECS from Concourse"
 
   assume_role_policy = <<EOF
--- a/terraform/deployments/govuk-test/certificates.tf
+++ b/terraform/deployments/govuk-test/certificates.tf
@@ -12,12 +12,6 @@ resource "aws_route53_record" "workspace_public_zone_ns" {
   ttl     = "30"
 
   records = aws_route53_zone.workspace_public.name_servers
-  # [
-  #   aws_route53_zone.external_zone.name_servers.0,
-  #   aws_route53_zone.external_zone.name_servers.1,
-  #   aws_route53_zone.external_zone.name_servers.2,
-  #   aws_route53_zone.external_zone.name_servers.3,
-  # ]
 }
 
 resource "aws_acm_certificate" "workspace_public" {
--- a/terraform/deployments/govuk-test/certificates.tf
+++ b/terraform/deployments/govuk-test/certificates.tf
@@ -1,0 +1,69 @@
+# TODO: Use ecs as default workspace
+# TODO: DNS Delegation from default zone
+data "aws_route53_zone" "root_public" {
+  # Created in govuk-aws
+  name = "${var.govuk_environment}.${var.public_domain}" # test.govuk.digital
+}
+
+resource "aws_route53_record" "workspace_public_zone_ns" {
+  zone_id = data.aws_route53_zone.root_public.zone_id
+  name    = local.public_lb_domain_name
+  type    = "NS"
+  ttl     = "30"
+
+  records = aws_route53_zone.workspace_public.name_servers
+  # [
+  #   aws_route53_zone.external_zone.name_servers.0,
+  #   aws_route53_zone.external_zone.name_servers.1,
+  #   aws_route53_zone.external_zone.name_servers.2,
+  #   aws_route53_zone.external_zone.name_servers.3,
+  # ]
+}
+
+resource "aws_acm_certificate" "workspace_public" {
+  domain_name       = "*.${local.public_lb_domain_name}"
+  validation_method = "DNS"
+
+  lifecycle {
+    create_before_destroy = true
+  }
+}
+
+resource "aws_route53_zone" "workspace_public" {
+  name = local.public_lb_domain_name # plouf.test.govuk.digital
+}
+
+resource "aws_route53_zone" "internal_public" {
+  name = local.internal_domain_name # plouf.test.govuk-internal.digital
+}
+
+resource "aws_route53_zone" "internal_private" {
+  name = local.internal_domain_name # plouf.test.govuk-internal.digital
+
+  vpc {
+    vpc_id = data.terraform_remote_state.infra_networking.outputs.vpc_id
+  }
+
+}
+
+resource "aws_route53_record" "workspace_public" {
+  for_each = {
+    for dvo in aws_acm_certificate.workspace_public.domain_validation_options : dvo.domain_name => {
+      name   = dvo.resource_record_name
+      record = dvo.resource_record_value
+      type   = dvo.resource_record_type
+    }
+  }
+
+  allow_overwrite = true
+  name            = each.value.name
+  records         = [each.value.record]
+  ttl             = 60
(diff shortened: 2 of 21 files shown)


terraform/deployments/concourse-iam/main.tf
```diff
@@ -27,7 +27,7 @@ provider "aws" {
 }
 
 resource "aws_iam_role" "govuk_concourse_deployer" {
-  name        = "govuk-concourse-deployer"
+  name        = "govuk-concourse-deployer-${terraform.workspace}"
```
bilbof, line 30: 👍 we'll also need to change `govuk-concourse-deployer` in the Concourse pipelines to point at the default workspace.

terraform/deployments/govuk-test/certificates.tf
```diff
@@ -0,0 +1,69 @@
...
+  ttl     = "30"
+
+  records = aws_route53_zone.workspace_public.name_servers
+  # [
+  #   aws_route53_zone.external_zone.name_servers.0,
+  #   aws_route53_zone.external_zone.name_servers.1,
+  #   aws_route53_zone.external_zone.name_servers.2,
+  #   aws_route53_zone.external_zone.name_servers.3,
+  # ]
+}
```
bilbof, line 21: Can we delete these lines?

terraform/deployments/govuk-test/main.tf
```diff
@@ -42,18 +42,29 @@ data "terraform_remote_state" "infra_security_groups" {
   }
 }
 
+locals {
+  internal_domain_name  = "${terraform.workspace == "default" ? var.internal_domain_name : "${terraform.workspace}.${var.internal_domain_name}"}" #plouf.test.govuk-internal.digital
+  public_lb_subdomain   = "${terraform.workspace == "default" ? var.govuk_environment : "${terraform.workspace}.${var.govuk_environment}"}"       #plouf.test
+  public_lb_domain_name = "${local.public_lb_subdomain}.${var.public_domain}"                                                                     #plouf.test.govuk.digital
+  mesh_domain           = "${var.mesh_subdomain}.${local.internal_domain_name}"                                                                   #mesh.plouf.govuk-internal.digital
+  mesh_name             = "${terraform.workspace == "default" ? var.mesh_name : "${var.mesh_name}-${terraform.workspace}"}"                       #govuk-plouf
+  ecs_cluster_name      = "${terraform.workspace == "default" ? var.ecs_cluster_name : "${var.ecs_cluster_name}-${terraform.workspace}"}"         #govuk-plouf
```
bilbof, line 51: I think we plan to use `ecs` as the default workspace name. Did you want to make that change in a subsequent PR?

terraform/modules/app/main.tf
```diff
@@ -15,6 +15,7 @@ locals {
   subdomain               = var.service_name
   container_services      = "${length(var.custom_container_services) == 0 ? [{ container_service = "${local.subdomain}", port = 80, protocol = "http" }] : var.custom_container_services}"
   service_security_groups = concat([aws_security_group.service.id], var.extra_security_groups)
+  security_group_name     = "${terraform.workspace == "default" ? var.service_name : "${var.service_name}-${terraform.workspace}"}"
```
bilbof, line 18: I think it would be clearer to pass in the security group `name` to the module, then this module doesn't need to know about the workspaces.

```
security_group_name = aws_security_group.publisher.name
```

terraform/modules/apps/publisher/main.tf
```diff
@@ -61,15 +61,15 @@ data "aws_acm_certificate" "public_lb_alternate" {
 }
 
 resource "aws_lb" "public" {
-  name               = "fargate-public-${var.service_name}"
+  name               = "${var.service_name}-${terraform.workspace}"
```
bilbof, line 64: I wonder if perhaps in the `govuk` module we could pass in an argument called `workspace_service_name`. It might be nicer to use this variable rather than creating `"${var.service_name}-${terraform.workspace}"` in a few different places.

bilbof (commented):
Nice work! I've added a few comments in line which would be good to address

terraform/deployments/concourse-iam/main.tf
rtrinque: Or we can use an interpolation to avoid changing the name of the role for default?

terraform/deployments/govuk-test/certificates.tf
rtrinque: Commented out lines removed

terraform/deployments/govuk-test/main.tf
rtrinque: Ok, named changed to ecs

terraform/modules/app/main.tf
rtrinque: I'm not sure about this one, the good thing with having the interpolation in the module is that this way we don't have to put it in every new apps we will write in the future.

terraform/deployments/concourse-iam/main.tf
```diff
@@ -27,7 +27,7 @@ provider "aws" {
 }
 
 resource "aws_iam_role" "govuk_concourse_deployer" {
-  name        = "govuk-concourse-deployer"
+  name        = "${terraform.workspace == "default" ? "govuk-concourse-deployer" : "govuk-concourse-deployer-${terraform.workspace}"}"
```
richardTowers, line 30: I don't think this one is needed - the `concourse-iam` deployment is separate to the rest of govuk. It's just here to give the concourse worker permission to run the terraform. I don't think that needs to be workspace aware.

terraform/deployments/govuk-test/main.tf
```diff
@@ -42,18 +42,29 @@ data "terraform_remote_state" "infra_security_groups" {
   }
 }
 
+locals {
+  internal_domain_name  = "${terraform.workspace == "default" ? var.internal_domain_name : "${terraform.workspace}.${var.internal_domain_name}"}" #plouf.test.govuk-internal.digital
+  public_lb_subdomain   = "${terraform.workspace == "default" ? var.govuk_environment : "${terraform.workspace}.${var.govuk_environment}"}"       #plouf.test
+  public_lb_domain_name = "${local.public_lb_subdomain}.${var.public_domain}"                                                                     #plouf.test.govuk.digital
+  mesh_domain           = "${var.mesh_subdomain}.${local.internal_domain_name}"                                                                   #mesh.plouf.govuk-internal.digital
```
richardTowers, line 49:
```suggestion
  mesh_domain           = "${var.mesh_subdomain}.${local.internal_domain_name}"                                                                   #mesh.plouf.test.govuk-internal.digital
```

terraform/deployments/concourse-iam/main.tf
rtrinque: Fixed

terraform/deployments/variables/test/infrastructure.tfvars
```diff
@@ -5,6 +5,8 @@ ecs_default_capacity_provider = "FARGATE_SPOT"
 #--------------------------------------------------------------
 
 # TODO: Could this be named more clearly?
```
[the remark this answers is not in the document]
rtrinque: Fixed

terraform/deployments/concourse-iam/main.tf
rtrinque: as Richard pointed out, this probably doesn't need to be workspace aware, reverted to its original value

terraform/deployments/govuk-test/main.tf
```diff
@@ -42,18 +42,29 @@ data "terraform_remote_state" "infra_security_groups" {
   }
 }
 
+locals {
+  internal_domain_name  = "${terraform.workspace == "default" ? var.internal_domain_name : "${terraform.workspace}.${var.internal_domain_name}"}" #plouf.test.govuk-internal.digital
+  public_lb_subdomain   = "${terraform.workspace == "default" ? var.govuk_environment : "${terraform.workspace}.${var.govuk_environment}"}"       #plouf.test
+  public_lb_domain_name = "${local.public_lb_subdomain}.${var.public_domain}"                                                                     #plouf.test.govuk.digital
+  mesh_domain           = "${var.mesh_subdomain}.${local.internal_domain_name}"                                                                   #mesh.plouf.test.govuk-internal.digital
+  mesh_name             = "${terraform.workspace == "default" ? var.mesh_name : "${var.mesh_name}-${terraform.workspace}"}"                       #govuk-plouf
+  ecs_cluster_name      = "${terraform.workspace == "default" ? var.ecs_cluster_name : "${var.ecs_cluster_name}-${terraform.workspace}"}"         #ecs-plouf
```
bilbof, line 51: The comments are a bit confusing here since they use "govuk", "plouf", and "ecs".

The "plouf" references throughout might be confusing too, but we can change those to `ecs` or remove them in another commit.

terraform/deployments/variables/test/common.tfvars
```diff
@@ -3,5 +3,7 @@
 #--------------------------------------------------------------
 
 # TODO: When it becomes possible, pull these from Terraform "data" type
-mesh_domain = "mesh.govuk-internal.digital"
-mesh_name   = "govuk"
+mesh_subdomain   = "mesh"
+mesh_domain      = "mesh.govuk-internal.digital"
+mesh_name        = "govuk"
+ecs_cluster_name = "ecs"
```
bilbof, line 9: What was the intent behind giving the cluster a prefix? Is it so we can define other clusters without having name clashes?

terraform/modules/apps/signon/main.tf
```diff
@@ -7,6 +7,10 @@ terraform {
   }
 }
 
+locals {
+  workspace_suffix = "${terraform.workspace == "default" ? "govuk" : "${terraform.workspace}"}"
```
bilbof, line 11: I think we should change the default workspace to `ecs` here

bilbof (approved):
Nice work 👏 

There are some small parts to change, but it would be great to merge this and follow up on those later.

PR #115, opened 2021-01-07
Grafana

fredericfran-gds:
As part of the [Observability epic](https://trello.com/c/tCpSuGiD/306-epic-observability), we spin up a Grafana instance in ECS to monitor GOV.UK ECS apps
and other components of the new platform e.g. billing.

## Features

1. Grafana spinned up in new cluster Monitoring to seperate concerns
2. Grafana is accessed via GitHub auth
3. Configuration of Grafana dashboards is via Terraform

## Testing

1. Grafana accessible via VPN at https://grafana-ecs.test.govuk.digital

## Deployment

Process is documented in docs/applying-terraform.md#monitoring

## Future Work

1. Subsequent apply of the terraform config for internal Grafana
configuration fails, workaround is documented in the deployment process.
2. Add teams to GitHub authorised users rather than the whole org
Alphagov
3. Deployment of the monitoring stack via Concourse

Ref:
1. [trello card](https://trello.com/c/RzaURJi5/323-set-up-a-grafana-in-ecs)


docs/applying-terraform.md
```diff
@@ -66,3 +66,57 @@ aws ecs wait services-stable \
...
+```sh
+cd terraform/deployments/monitoring-test/grafana/app-config
+
+gds aws govuk-test-admin -- terraform init
+
+gds aws govuk-test-admin -- terraform apply
+```
+
+If you get an error: `Error: status: 404, body: {"message":"Data source not found"}`,
+you should run: `gds aws govuk-test-admin -- terraform state rm module.grafana-app-config.grafana_data_source.cloudwatch` before re-applying the terraform.
```
bilbof, line 122: Would this error occur frequently? What is the cause of it?

terraform/deployments/monitoring-test/base/main.tf
```diff
@@ -0,0 +1,71 @@
+terraform {
```
bilbof, line 1: Could this file be moved to `terraform/deployments/monitoring-test/main.tf`, or is the `base` directory required?

terraform/deployments/monitoring-test/base/main.tf
```diff
@@ -0,0 +1,71 @@
...
+  backend "s3" {
+    bucket  = "govuk-terraform-test"
```
bilbof, line 3:
```suggestion
    bucket  = var.govuk_aws_state_bucket
```

terraform/deployments/monitoring-test/base/main.tf
```diff
@@ -0,0 +1,71 @@
...
+    key      = "govuk/infra-security-groups.tfstate"
+    region   = "eu-west-1"
+    role_arn = var.assume_role_arn
+  }
+}
+
+data "terraform_remote_state" "govuk" {
+  backend = "s3"
+  config = {
+    bucket   = "govuk-terraform-test"
```
bilbof, line 48:
```suggestion
    bucket   = var.govuk_aws_state_bucket
```

If `var.govuk_aws_state_bucket` is permitted in both `data.config` and `terraform.backend` blocks, this would make the module reusable by other environments.

We could move the module and pass in the environment specific variables:

```diff
+ terraform/deployments/monitoring-test/base/main.tf
- terraform/deployments/monitoring/main.tf
```

What do you think?

terraform/deployments/monitoring-test/base/main.tf
```diff
@@ -0,0 +1,71 @@
...
+    key      = "projects/govuk.tfstate"
+    region   = "eu-west-1"
+    role_arn = var.assume_role_arn
+  }
+}
+
+data "aws_iam_role" "execution" {
+  name = "fargate_execution_role"
```
bilbof, line 56: Should we pull this out of remote state too? Same for terraform/deployments/monitoring-test/grafana/app-config/main.tf - would be nice to pull data out consistently. Feel free to add TODOs for these and we could change them later.

terraform/deployments/monitoring-test/grafana/main.tf
```diff
@@ -0,0 +1,135 @@
...
+data "aws_secretsmanager_secret" "github_client_id" {
+  name = "grafana_github_client_id"
+}
+
+data "aws_secretsmanager_secret" "github_client_secret" {
+  name = "grafana_github_client_secret"
+}
+
+
+resource "random_string" "grafana_password" {
```
bilbof, line 62: Cool! I think we should use `random_password` here

https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/password

docs/applying-terraform.md
```diff
@@ -66,3 +66,57 @@ aws ecs wait services-stable \
  --services "$APPLICATION" \
  --region "$AWS_REGION"
 ```
+
+### Monitoring
+
+The monitoring stack of GOV.UK is in a separate cluster compared to the one where
+the GOV.UK apps run. Below are the details how to run deploy the monitoring stack
+which includes only Grafana for now.
+
```
bilbof, line 75: It might also be worth mentioning in the docs that if we are applying to a new environment, we'll need to setup a GitHub app (and manually add `github_client_secret` to SecretsManager).

terraform/deployments/monitoring-test/grafana/main.tf
```diff
@@ -0,0 +1,135 @@
...
+
+  environment_variables = {
+    GF_SECURITY_ADMIN_USER               = "admin",
+    GF_AUTH_GITHUB_ENABLED               = true,
+    GF_AUTH_GITHUB_SCOPES                = "user:email,read:org",
+    GF_AUTH_GITHUB_AUTH_URL              = "https://github.com/login/oauth/authorize",
+    GF_AUTH_GITHUB_TOKEN_URL             = "https://github.com/login/oauth/access_token",
+    GF_AUTH_GITHUB_API_URL               = "https://api.github.com/user",
+    GF_AUTH_GITHUB_ALLOW_SIGN_UP         = true,
+    GF_AUTH_GITHUB_ALLOWED_ORGANIZATIONS = "alphagov",
```
bilbof, line 88: Unnecessary now, but we can also restrict login access to people on GOV.UK.

```
GF_AUTH_GITHUB_TEAM_IDS  = "123" # gov-uk team
```

terraform/modules/monitoring/grafana.tf
```diff
@@ -0,0 +1,134 @@
...
+  domain   = "*.test.govuk.digital"
+  statuses = ["ISSUED"]
+}
+
+data "aws_acm_certificate" "public_lb_alternate" {
+  domain   = "*.test.publishing.service.gov.uk"
+  statuses = ["ISSUED"]
+}
+
+resource "aws_lb" "grafana_public" {
```
bilbof, line 74: There's a `public-load-balancer` module you can make use of, which might enable you to remove the ALB resources here.

terraform/modules/monitoring/variables.tf
```diff
@@ -0,0 +1,58 @@
...
+variable "desired_count" {
+  description = "Desired count of Application instances"
+  type        = number
+  default     = 1
+}
+
+variable "office_cidrs_list" {
+  description = "List of GDS office CIDRs"
+  type        = list
+  default     = ["213.86.153.212/32", "213.86.153.213/32", "213.86.153.214/32", "213.86.153.235/32", "213.86.153.236/32", "213.86.153.237/32", "85.133.67.244/32"]
```
bilbof, line 57: I wonder if there's a way to avoid duplicating these here? Could we remove the default and put them in `tfvars`?

bilbof (commented):
Nice work 👏 Thanks for adding documentation, too!

I've added some comments inline - let me know if you have any questions.

We don't need to separate the ECS Task Definition and ECS Service deployments for Grafana. We separate out the deployments for apps so we can hook into deploys and run steps like database migrations, but I'm not sure we need to do that in this case - it's just the app module pushes us to do it.

We could merge the deployments (`monitoring-test/grafana/base` and `monitoring-test/grafana/app-config`) for Grafana into a single deployment (perhaps `monitoring/grafana`) and remove a few files, [as with Statsd](https://github.com/alphagov/govuk-infrastructure/pull/116). What do you think?

fredericfran-gds:
> Nice work 👏 Thanks for adding documentation, too!
> 
> I've added some comments inline - let me know if you have any questions.
> 
> We don't need to separate the ECS Task Definition and ECS Service deployments for Grafana. We separate out the deployments for apps so we can hook into deploys and run steps like database migrations, but I'm not sure we need to do that in this case - it's just the app module pushes us to do it.
> 
> We could merge the deployments (`monitoring-test/grafana/base` and `monitoring-test/grafana/app-config`) for Grafana into a single deployment (perhaps `monitoring/grafana`) and remove a few files, [as with Statsd](https://github.com/alphagov/govuk-infrastructure/pull/116). What do you think?

You can't run these 2 together because the `app-config` make HTTP calls the grafana app after it has been up (i.e. after ecs service stable).

docs/applying-terraform.md
fredericfran-gds: This occurs when a new Grafana container is spun and it does not have the Cloudwatch config anymore. It is weird that it didn't just re-create it.

terraform/deployments/monitoring-test/base/main.tf
fredericfran-gds: we could move the files in base one step up. It was just me structuring it this way to look nicer from my point of view.

terraform/deployments/monitoring-test/base/main.tf
fredericfran-gds: I can do that.

terraform/deployments/monitoring-test/base/main.tf
fredericfran-gds: yeah, I was using govuk-test as the example but I can do that because I think this will be the new style. I don't know if I want to keep the base folder though.

terraform/deployments/monitoring-test/base/main.tf
fredericfran-gds: yep, I can try to do that if easier.

terraform/deployments/monitoring-test/grafana/main.tf
fredericfran-gds: good catch, I was going for that but I don't know what happened.

docs/applying-terraform.md
fredericfran-gds: good point. I will do that.

terraform/deployments/monitoring-test/grafana/main.tf
fredericfran-gds: Yeah, it is a future thing to do. Might try to do it today, I have to use curl to get the list as you can't get it via GUI directly.

terraform/modules/monitoring/grafana.tf
fredericfran-gds: I will have a look. thanks.

terraform/modules/monitoring/variables.tf
fredericfran-gds: yeah, now that it is in 2 places, maybe time to pull it up.

terraform/modules/monitoring/grafana.tf
```diff
@@ -0,0 +1,134 @@
+module "grafana" {
+  source                           = "../app"
+  execution_role_arn               = var.execution_role_arn
+  vpc_id                           = var.vpc_id
+  cluster_id                       = aws_ecs_cluster.cluster.id
+  service_name                     = "grafana"
+  subnets                          = var.private_subnets
+  mesh_name                        = var.mesh_name
+  service_discovery_namespace_id   = var.service_discovery_namespace_id
+  service_discovery_namespace_name = var.service_discovery_namespace_name
```
bilbof, line 10: One problem with having the module reuse the `service_discovery_namespace_name` created by the `govuk` deployment, is that we cannot destroy and recreate the `govuk` deployment without first destroying the `grafana` deployment.

If the `monitoring/grafana` module has been applied, we'll see the following error when trying `cd govuk-test && terraform destroy`

```
Error: error deleting Service Discovery Private DNS Namespace: ResourceInUse: Namespace has associated services; delete the services before deleting the namespace
```

```
Error: error deleting App Mesh service mesh: ResourceInUseException: Mesh govuk has active VirtualServices, please delete VirtualServices before deleting the Mesh
```

I wonder if we should accept that rebuilding the infrastructure first requires destroying grafana.

Alternatively we could see if we can extricate the grafana deployment completely, or merge it into the govuk deployment. What do you think?

terraform/deployments/monitoring-test/base/main.tf
bilbof: Moving the resources into one state file would make sense to me. It would mean only a single `terraform apply` would be required.

bilbof:
> You can't run these 2 together because the app-config make HTTP calls the grafana app after it has been up (i.e. after ecs service stable).

What HTTP calls does the app-config make, that it needs to be a separate module? Could you explain this a bit more - sorry I'm a bit confused.

terraform/deployments/monitoring-test/grafana/main.tf
bilbof: No worries about doing this right now, a TODO would be enough.

terraform/deployments/monitoring-test/base/main.tf
bilbof: Your call. Feel free to use TODOs.

terraform/deployments/monitoring-test/base/main.tf
bilbof: Thanks. Yes I think the `base` nesting is unnecessary.

docs/applying-terraform.md
bilbof: Interesting, is that a bug in the grafana provider? 

I wonder if we can prevent it somehow... perhaps with a `depends_on = [grafana_data_source.cloudwatch]`?

fredericfran-gds: Yes it is a good point. If for any reason, the container is respun, we lose the config data.

terraform/modules/monitoring/grafana.tf
fredericfran-gds: Umm.. we wanted to keep it separate from the discussion with Richard. Maybe what we can do it not include an app into the service mesh if the namespace is null. This will solve this issue.

fredericfran-gds:
> > You can't run these 2 together because the app-config make HTTP calls the grafana app after it has been up (i.e. after ecs service stable).
> 
> What HTTP calls does the app-config make, that it needs to be a separate module? It sounds like we create the ECS Service and then the grafana provider makes the HTTP calls to set the config for grafana?
> 
> I wonder, could grafana pull config from an S3 bucket, or could we build a grafana image with our config built in? This might be a way to solve the problem, given we need to apply terraform every time a new container is cycled.

Yes, you are right: "The ECS Grafana Service is first created and then the Grafana Terraform provider makes the HTTP calls to set the config for Grafana". 

We can persist this change by backing Grafana with a database as mentioned above: https://grafana.com/docs/grafana/latest/administration/set-up-for-high-availability/

terraform/modules/monitoring/grafana.tf
bilbof: 🤔 what was the reason for separating it? Perhaps each workspace doesn't need its own grafana? But maybe it does :)

I ask, since it introduces hidden dependencies - e.g. we can't apply the main branch now (https://cd.gds-reliability.engineering/teams/govuk-tools/pipelines/deploy-apps-test/jobs/run-terraform/builds/71). If we put the monitoring stuff back into the govuk deployment it'll mean extra grafanas but I think that's OK - we shouldn't be designing the terraform for production around a tool for the test environment.

bilbof: I chatted with Richard about the motivation for splitting them up. The main reason is that we can apply grafana to production without having to bring up a whole cluster, and get immediate benefit from the new version of grafana (early gains from replatforming 😄).

We could create a `monitoring` ECS Cluster, and not use App Mesh for this app, since it just needs a public ALB. That would remove some dependencies on the govuk module. What do you think?

docs/applying-terraform.md
bilbof: That sounds good, sorry it's extra faf - feel free to merge this in and write a card to do that. I'm guessing we could still use the Terraform provider in this case, since it'll just persist whatever changes the resource makes via HTTP calls.

terraform/deployments/monitoring-test/base/main.tf
fredericfran-gds: we would still require the same number of Terraform applies.

bilbof: 🤔 is there a way we could have a single TF apply, or is the grafana provider not clever enough to wait for the container to be ready?

terraform/modules/monitoring/variables.tf
bilbof: A TODO is fine here too 👍

sengi:
> > > You can't run these 2 together because the app-config make HTTP calls the grafana app after it has been up (i.e. after ecs service stable).
> > 
> > 
> > What HTTP calls does the app-config make, that it needs to be a separate module? It sounds like we create the ECS Service and then the grafana provider makes the HTTP calls to set the config for grafana?
> > I wonder, could grafana pull config from an S3 bucket, or could we build a grafana image with our config built in? This might be a way to solve the problem, given we need to apply terraform every time a new container is cycled.
> 
> Yes, you are right: "The ECS Grafana Service is first created and then the Grafana Terraform provider makes the HTTP calls to set the config for Grafana".
> 
> We can persist this change by backing Grafana with a database as mentioned above: https://grafana.com/docs/grafana/latest/administration/set-up-for-high-availability/

Let's not introduce a database if we can help it. That's a big leap in complexity and it shouldn't be necessary for what should be a stateless app which reads some config at startup.

terraform/deployments/monitoring-test/base/main.tf
fredericfran-gds: not possible, see issue: https://github.com/hashicorp/terraform/issues/13022

terraform/deployments/monitoring-test/base/main.tf
fredericfran-gds: since the backend is not parameterizable, we will have to stick to a fixed value.

terraform/deployments/monitoring-test/grafana/main.tf
fredericfran-gds: done

fredericfran-gds: done

terraform/deployments/monitoring-test/base/main.tf
fredericfran-gds: I created some iams directly for monitoring so that it is not dependent on govuk.

terraform/modules/monitoring/grafana.tf
fredericfran-gds: done, I had to add some options to the `public-load-balancer` module

terraform/modules/monitoring/variables.tf
fredericfran-gds: Done

terraform/modules/monitoring/grafana.tf
fredericfran-gds: Now the monitoring cluster is independent of govuk one.

terraform/modules/app/appmesh.tf
```diff
@@ -13,7 +13,7 @@ resource "aws_appmesh_virtual_service" "service" {
 }
 
 resource "aws_appmesh_virtual_node" "service" {
-  count     = length(local.container_services)
+  count     = var.service_mesh ? length(local.container_services) : 0
```
bilbof, line 16: It might be worth not using the app module for grafana, since it has a fair few differences. It might make the config for both apps and grafana a bit easier to reason about. What do you think?

terraform/modules/govuk/variables.tf
```diff
@@ -129,7 +129,6 @@ variable "signon_desired_count" {
 variable "office_cidrs_list" {
   description = "List of GDS office CIDRs"
   type        = list
-  default     = ["213.86.153.212/32", "213.86.153.213/32", "213.86.153.214/32", "213.86.153.235/32", "213.86.153.236/32", "213.86.153.237/32", "85.133.67.244/32"]
```
bilbof, line 132: Thanks for this 👍

bilbof (approved):
Fantastic 👍 You just have some conflicts.

I'm wondering if we could make the app module simpler by avoiding having grafana use it, but feel free to ignore that comment.

It might be nice to write followup cards to look into resolving some of issues we didn't get to (e.g. cycling grafana instances leads to grafana not being configured).

terraform/modules/app/appmesh.tf
fredericfran-gds: with you change of introducing the service mesh module, this is not a bit complexity.

fredericfran-gds:
> Fantastic 👍 You just have some conflicts.
> 
> I'm wondering if we could make the app module simpler by avoiding having grafana use it, but feel free to ignore that comment.
> 
> It might be nice to write followup cards to look into resolving some of issues we didn't get to (e.g. cycling grafana instances leads to grafana not being configured).

Yes, I will.

PR #116, opened 2021-01-07
Run Statsd as an ECS Service

bilbof:
This runs Statsd as an ECS service within the 'govuk' ECS cluster.

I've left the workspace out of the statsd namespaces for now, but we could add the workspace at the top level e.g. `govuk-<workspace>.app.<app_name>.ecs`.

Paired with @smford on this.


terraform/modules/app/main.tf
```diff
@@ -67,3 +67,13 @@ resource "aws_security_group" "service" {
   vpc_id      = var.vpc_id
   description = "${var.service_name} app ECS tasks"
 }
+
+resource "aws_security_group_rule" "statsd_from_apps_tcp" {
```
richardTowers, line 71: Not necessarily a blocker for this PR, but I wonder if we should have a separate security group for "all the apps in the mesh", so we can represent the kinds of statsd / prometheus situations a bit more cleanly.

So the security group rule would be "any app in this security group can talk to statsd", rather than having a separate rule for every app.

Later on we're going to need to say "all apps can talk to prometheus", and probably other similar things, which would be easier if we had a single group that represented all the apps.

bilbof: That sounds reasonable to me. Perhaps we can do this when we create Prometheus?

I had an impression from working with govuk-aws that shared security groups (like `govuk_management_access`) were an anti pattern, but I'm not speaking from much experience. I wonder if this could lead to folks being tempted to use the "app_in_mesh" group inappropriately?

One thing to watch out for is that ECS Services are permitted a maximum of 5 security groups. This isn't a blocker for doing this, just something to be aware of given some apps would have 3 security groups if we went down this route.

terraform/modules/task-definitions/signon/main.tf
```diff
@@ -15,6 +15,10 @@ provider "aws" {
   }
 }
 
+locals {
+  app_name = "signon"
+}
+
```
sengi, line 21: I'd merge this with the existing locals block further down (similarly for all the other occurrences). Could even consider reusing local.service_name as well, if it fits in most cases - no big deal either way on that one though because we'll need to take a holistic look at these various app-name-related variables at some point soon (but not now).

terraform/modules/statsd/appmesh.tf
```diff
@@ -0,0 +1,49 @@
+resource "aws_appmesh_virtual_service" "statsd" {
```
sengi, line 1: What was the thing that prevented using the app module for this boilerplate stuff? (sorry for being lazy - just easier to ask rather than play spot-the-difference!)

terraform/modules/app/main.tf
sengi: Ultimately we don't want to be passing security groups down into every module if we can help it. Having an "all services" group sounds like a good way to do that. You're right that we do have to use security groups sparingly but this is a pretty compelling reason. One for https://trello.com/c/QnM38plJ/31-plan-architecture-of-vpc-network-acls-and-security-groups

sengi: (I'm also ok with this for now btw, as long as everyone knows it's not something to be copy/pasted!)

terraform/modules/statsd/appmesh.tf
bilbof: The app module sadly uses a bootstrap image, and relies on the app deployments to update the service. Statsd doesn't need an app deployment, so I put it in here for now. I agree the duplication isn't nice - perhaps a mini-module for cloudmap/app mesh might be useful.

terraform/modules/app/main.tf
bilbof: Thanks both, this seems like a good approach so I'll fix it up in this commit.

terraform/modules/statsd/appmesh.tf
sengi: Oh of course, thanks. I wonder if a small generalisation to the app module might be the way forward - i.e. allow passing in task_definition but have it default to module.bootstrap_task_definition.arn.

bilbof: I considered that, though another problem is that while this service is unlikely to require complex deployment steps or be continuously deployed - so doesn't warrant a deployment - we have [a lifecycle rule](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/modules/app/main.tf#L55) on the task definition which would require a deployment module. Though perhaps there is a terraform way to wrap this in a conditional?

sengi: Hm yeah that's a good point, but as you've suggested I think it'd be solved neatly with a straightforward conditional when setting the lifecycle rule.

bilbof: Sadly I have found a conditional isn't possible in a lifecycle rule (https://github.com/hashicorp/terraform/issues/3116#issuecomment-147382638). I'll see if I can find a neat way around this.

sengi: Oh no 🙈 sorry, didn't mean to send you down a rabbithole there.

bilbof: No it's OK, we were going to hit this problem eventually. A couple of options I'm exploring: a) treat Statsd an app, with an unnecessary deployment module, b) treat Statsd as something different and try to share config with the app module. I've pushed up a change for the second option.

terraform/modules/app/main.tf
bilbof: I've made this change

sengi (approved):
Let's go with this and see how it pans out. Just a couple of trivial suggestions.

PR #118, opened 2021-01-08
Output redis host and port

richardTowers:
We'll need these for any app task definitions which need REDIS_HOST and REDIS_PORT as environment variables.

Note that REDIS_URL shouldn't be needed unless the URL gets more complicated (e.g. needs basic auth creds or something). See [govuk_sidekiq's docs](https://github.com/alphagov/govuk_sidekiq/blob/2c30a9b02c5b1741b6069b87583ffb722b9542a0/README.md#4-configure-puppet).

Terraform plan output:

```
Plan: 0 to add, 0 to change, 0 to destroy.


Changes to Outputs:

  + redis_host = "shared-redis.test.govuk-internal.digital"

  + redis_port = 6379
```


MahmudH (approved):
👍

PR #120, opened 2021-01-12
Fixup content store deployments

bilbof:
This will enable us to apply content-store after its refactoring.

The task definition Concourse script now supports an app deployment module outputting multiple task definition arns.

content-store's outputs.tf before:

```terraform
output "task_definition_arn" {
  value       = module.task_definition.arn
  description = "ARN of the task definition revision"
}
```

content-store's outputs.tf after:

```terraform
output "draft_task_definition" {
  description = "ARNs of the draft task definition revision"
  value       = aws_ecs_task_definition.draft.arn
}

output "live_task_definition" {
  description = "ARNs of the live task definition revision"
  value       = aws_ecs_task_definition.live.arn
}
```

Following this change, the deploy pipeline must now select the correct task definition to use on the ECS Service.

We could dynamically select a task definition by specifying 'live' or 'web' or 'draft-worker' in the pipeline job params. I found it simpler to make the dependency on the output of modules explicit (`live_task_definition`).

This has further implications for other things dependent on the output of the app deployments. For example, `tools/run-task.sh` needs to be rewritten (it won't work for a few apps, but that's OK).

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -205,7 +205,7 @@ jobs:
     - task: update-ecs-service
       file: govuk-infrastructure/concourse/tasks/update-ecs-service.yml
       params:
-        APPLICATION: frontend
+        ECS_SERVICE: frontend
         GOVUK_ENVIRONMENT: test
     serial: true
     on_failure: &notify-slack-failure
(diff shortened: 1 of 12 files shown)


fredericfran-gds (approved):
looks good

PR #122, opened 2021-01-13
Run deploy tasks in parallel

bilbof:
We can make use of the [in_parallel step](https://concourse-ci.org/jobs.html#in-parallel-step) in Concourse jobs where tasks don't depend on each other.

This makes app deployments complete up to 2x faster. Publisher deploys (which update multiple ECS services) take 4mins, down from 8mins. All apps see a 1min reduction in build time due to fetching git repos in parallel.

Parallel builds

Here we can get repos, register new task definitions, and apply the new task definitions to ECS Services, with each step running multiple tasks in parallel, which will speed up deploy times for these services.


richardTowers (approved):

sengi (approved):
Nice

PR #124, opened 2021-01-13
fix refactor conflicts #115 and #116

fredericfran-gds:
In #116, there is a re-factoring of the app mesh and cloud map into a new service mesh module. This was not taken into account in #115. This fixes this. Applies works locally.


rtrinque (approved):
Looks good to me

PR #125, opened 2021-01-14
Refactor app deployments in concourse

richardTowers:
Previously there was a race condition, where both content-store and
draft-content-store could apply the same terraform at the same time.
Even if we had locking on our terraform statefile (which we don't at the
moment), this doesn't seem like great design.

Instead, I think we should just have one job for content-store, which
updates both the draft and live version.

The process is now:
* `update-task-definition` (which applies the content-store terraform
  deployment, creating a task definition for both draft and live stacks)
* In parallel, run `update-ecs-service` for both the draft and live
  stacks

To make this work, I've refactored how the terraform outputs get passed
around. Instead of creating one file per output, the
update-task-definition task now puts all the outputs in one big JSON
file. update-ecs-service can then use `jq` (which is conveniently
already in the awscli container) to pluck out the output it wants.

I think this should work for old and new-style apps.

NOTE: draft because I haven't tested this on concourse yet.

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -15,7 +15,7 @@ resources:
     icon: github
     name: govuk-infrastructure
     source:
-      branch: main
+      branch: refactor-app-deployment-concourse
       uri: https://github.com/alphagov/govuk-infrastructure
     type: git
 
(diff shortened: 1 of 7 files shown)


richardTowers:
Going to re-open this from a different branch...

PR #126, opened 2021-01-14
Refactor app deployments in concourse

richardTowers:
Previously there was a race condition, where both content-store and
draft-content-store could apply the same terraform at the same time.
Even if we had locking on our terraform statefile (which we don't at the
moment), this doesn't seem like great design.

Instead, I think we should just have one job for content-store, which
updates both the draft and live version.

The process is now:
* `update-task-definition` (which applies the content-store terraform
  deployment, creating a task definition for both draft and live stacks)
* In parallel, run `update-ecs-service` for both the draft and live
  stacks

To make this work, I've refactored how the terraform outputs get passed
around. Instead of creating one file per output, the
update-task-definition task now puts all the outputs in one big JSON
file. update-ecs-service can then use `jq` (which is conveniently
already in the awscli container) to pluck out the output it wants.

Annoyingly the terraforom outputs need to go in application-specific
paths, because publisher still has separate web and worker
deployments. Once the publisher refactoring is finished, this can be
further simplified.

I think this should work for old and new-style apps.

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -86,7 +86,6 @@ groups:
       - update-pipeline
       - run-terraform
       - deploy-content-store
-      - deploy-draft-content-store
       - deploy-frontend
       - smoke-test-frontend
       - deploy-draft-frontend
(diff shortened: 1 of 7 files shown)


richardTowers:
This is currently deployed to concourse here: https://cd.gds-reliability.engineering/teams/govuk-tools/pipelines/deploy-apps-test

... and it looks like everything's green. I should be able to clean this up quite a bit once the publisher refactoring lands :)

bilbof (approved):
LGTM, thanks 👍

PR #127, opened 2021-01-14
Fix Statsd security group rules

bilbof:
This enables Statsd tasks to pull the Envoy image from ECR, and fixes an issue that prevents apps from sending requests to Statsd.

It also adds rules to the `mesh_ecs_service` security group, which will help us to remove the security group `govuk_management_access` from services.

### DIFF
--- a/terraform/modules/govuk/security_group_rules.tf
+++ b/terraform/modules/govuk/security_group_rules.tf
@@ -506,7 +506,7 @@ resource "aws_security_group_rule" "signon_to_any_any" {
 
 resource "aws_security_group_rule" "statsd_from_apps_tcp" {
   description              = "Allow services in the App Mesh to send metrics to the mesh Statsd via TCP"
-  type                     = "egress"
+  type                     = "ingress"
   from_port                = "8125"
   to_port                  = "8125"
   protocol                 = "tcp"
--- a/terraform/modules/govuk/security_group_rules.tf
+++ b/terraform/modules/govuk/security_group_rules.tf
@@ -514,5 +514,45 @@ resource "aws_security_group_rule" "statsd_from_apps_tcp" {
   source_security_group_id = aws_security_group.mesh_ecs_service.id
 }
 
+resource "aws_security_group_rule" "mesh_service_to_any_https" {
+  description       = "Mesh services send requests to anywhere over HTTPS"
+  type              = "egress"
+  from_port         = 443
+  to_port           = 443
+  protocol          = "tcp"
+  cidr_blocks       = ["0.0.0.0/0"]
+  security_group_id = aws_security_group.mesh_ecs_service.id
+}
+
+resource "aws_security_group_rule" "mesh_service_to_any_dns_tcp" {
+  description       = "Mesh services send DNS queries anywhere over TCP"
+  type              = "egress"
+  from_port         = 53
+  to_port           = 53
+  protocol          = "tcp"
+  cidr_blocks       = ["0.0.0.0/0"]
+  security_group_id = aws_security_group.mesh_ecs_service.id
+}
+
+resource "aws_security_group_rule" "mesh_service_to_any_dns_udp" {
+  description       = "Mesh services send DNS queries anywhere over UDP"
+  type              = "egress"
+  from_port         = 53
+  to_port           = 53
+  protocol          = "udp"
+  cidr_blocks       = ["0.0.0.0/0"]
+  security_group_id = aws_security_group.mesh_ecs_service.id
+}
+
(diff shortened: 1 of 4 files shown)


terraform/modules/govuk/security_group_rules.tf
```diff
@@ -506,13 +506,23 @@ resource "aws_security_group_rule" "signon_to_any_any" {
...
 
+resource "aws_security_group_rule" "mesh_service_to_any_any" {
+  description       = "Mesh services send requests to anywhere over any protocol"
+  type              = "egress"
+  from_port         = 0
+  to_port           = 0
+  protocol          = -1
+  cidr_blocks       = ["0.0.0.0/0"]
+  security_group_id = aws_security_group.mesh_ecs_service.id
+}
```
richardTowers, line 525: So, this is consistent with the existing infrastructure, because `govuk_management_access` has a similar rule attached. But do we actually want to permit full egress from every service?

bilbof: You're right, it shouldn't be necessary.

To give some context, we're seeing `CannotPullContainerError` when trying to pull the Envoy image for Statsd tasks ([example](https://eu-west-1.console.aws.amazon.com/ecs/home?region=eu-west-1#/clusters/govuk/tasks/b52999b10e544bb894469695152d5d55/details)), but not for other any other task:

```
CannotPullContainerError: Error response from daemon: Get https://<ecr-url>/v2/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

I had thought that permission to pull from ECR was covered by the execution role that each Fargate task has. Since the `fargate_execution_role`  has the managed policy `AmazonECSTaskExecutionRolePolicy` attached, as recommended [in the docs](https://docs.aws.amazon.com/app-mesh/latest/userguide/envoy.html):

> Access to this container image in Amazon ECR is controlled by AWS Identity and Access Management

I asked AWS support about what was going on, and they wrote back

> Basically your security group rules for the Task should allow outbound access. The default here is typically All Traffic to 0.0.0.0/0. 
> When using the Amazon provided DNS in VPC, the Fargate task needs to have outbound traffic allowed for HTTPS (port 443) [1]. Therefore, the task you launched in the ECS service "statsd" was not allowed any outbound traffic and could not pull image from your ECR repository.

They also recommended we look at [this troubleshooting doc](https://aws.amazon.com/premiumsupport/knowledge-center/ecs-pull-container-api-error-ecr/) which wants us to allow outbound access:

> Confirm that the security group attached to your instance and Fargate task allows outbound access for the following: HTTPS on port 443, DNS (UDP and TCP) on port 53, and your subnet's network access control list (network ACL). from [docs on pulling image from ECR](https://aws.amazon.com/premiumsupport/knowledge-center/ecs-pull-container-api-error-ecr/)

My understanding is that we've not noticed this requirement before, because of the `govuk_management_access` security group being attached to every service.

I considered restricting the outbound traffic to the subset required (HTTPS etc), but I thought that'll be a source of confusing errors in the future.

I can't see a way to cleanly restrict the outbound traffic to just a particular ECR repository, though [others have looked into it](https://stackoverflow.com/questions/53190853/security-group-egress-rule-to-only-permit-ecr-requests). Is there a simple way to do this that I've missed?

richardTowers: Oh crikey.

I think it would be a little nice to have `mesh_service_to_any_https` and `mesh_service_to_any_dns` rules, and then have a ticket to purge `govuk_management_access` from this repo. It's extremely badly named.

bilbof: Thanks, I've added those, and created [a ticket](https://trello.com/c/v7S1lVMg/171-purge-govukmanagementaccess-from-govuk-infrastructure).

terraform/modules/govuk/security_group_rules.tf
```diff
@@ -506,13 +506,43 @@ resource "aws_security_group_rule" "signon_to_any_any" {
...
+
+resource "aws_security_group_rule" "mesh_service_to_any_dns" {
+  description       = "Mesh services send DNS queries anywhere"
+  type              = "egress"
+  from_port         = 53
+  to_port           = 53
+  protocol          = -1
```
richardTowers, line 532: I think you need to set this to:
```suggestion
  protocol          = "udp"
```

There's a warning at the top of the terraform docs about this: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group_rule

bilbof: Thanks, that should be fixed now.

richardTowers (approved):

PR #128, opened 2021-01-14
Reduce frequency of app deployments

richardTowers:
Previously, concourse would run every app deployment after every run of
the run-terraform job (which deploys the govuk-terraform).

This leads to the app deployments running far more frequently than they
really need to.

Ideally, app deployments would only happen when one of:

* A new version of the container image is published
* Their terraform deployment has changed
* One of the modules used by their terraform deployment has changed
* The outputs of the govuk-test deployment have changed
* One of the concourse tasks involved in their deployment has changed

To realise this dream, I've created:

* An S3 resource that recieves terraform outputs from run-terraform -
  every app deployment is triggered by this.
* One extra git resource for every app, which uses the `paths` property
  so that only changes to files relevant to the app deployment create
  new versions

This should significantly reduce the situations where concourse decides
to redeploy every app simultaneously (although that will still happen in
some situations, like a shared module being changed, or a terraform
output changing).

| Before | After |
|---|---|
|  image | image |

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -6,15 +6,21 @@ resource_types:
     type: docker-image
     source:
       repository: cfcommunity/slack-notification-resource
-      tag: latest
+      tag: latest # TODO - don't use latest (once we've worked out a policy for third party images)
       username: ((docker_hub_username))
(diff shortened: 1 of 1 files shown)


bilbof (approved):
Nice 👏

PR #129, opened 2021-01-14
add monitoring concourse pipeline

fredericfran-gds:
In #115, we introduced the new monitoring cluster with Grafana as a
Fargate ECS service in it.

In this PR, we create the associated concourse pipeline.
This can be accessed here: https://cd.gds-reliability.engineering/teams/govuk-tools/pipelines/monitoring

The concourse pipeline works as follows:
1. deploy the base monitoring cluster
2. deploy the Grafana ECS task, i.e container.
3. configure Grafana

The task to configure Grafana is designed to:
1. wait for Grafana to be up and running, as determined by HTTP status
code 200
2. apply the Terraform Grafana config
3. handle bug https://github.com/grafana/terraform-provider-grafana/issues/129
where terraform does not recreate some grafana resources if not found

In addition, we need to add CDIO Big Concourse Egress IPs to the list of
IPs allowed to reach Grafana so that Concourse can apply against Grafana.

Future work:
1. refactor shell scripts into separate files once agreed style.

Ref:
1. [trello card](https://trello.com/c/ZJuJoV3H/350-create-monitoring-concourse-pipeline)

### DIFF
--- a/concourse/pipelines/monitoring.yml
+++ b/concourse/pipelines/monitoring.yml
@@ -1,0 +1,235 @@
+---
+definitions:
+
+resources:
+  - icon: github
+    name: govuk-infrastructure
+    source:
+      branch: main
+      uri: https://github.com/alphagov/govuk-infrastructure
+    type: git
+
+groups:
+  - name: all
+    jobs:
+      - update-pipeline
+      - deploy-monitoring-infra
+      - deploy-grafana
+      - configure-grafana
+
+  - name: monitoring-infra
+    jobs:
+      - deploy-monitoring-infra
+      - deploy-grafana
+      - configure-grafana
+
+  - name: admin
+    jobs:
+      - update-pipeline
+
+jobs:
+  - name: update-pipeline
+    plan:
+    - get: govuk-infrastructure
+      trigger: true
+    - file: govuk-infrastructure/concourse/pipelines/monitoring.yml
+      set_pipeline: monitoring
+
+  - name: deploy-monitoring-infra
+    plan:
+    - get: govuk-infrastructure
+      trigger: true
+    - task: terraform-apply
+      config:
+        inputs:
+        - name: govuk-infrastructure
+        params:
+          AWS_REGION: eu-west-1
+          ASSUME_ROLE_ARN: 'arn:aws:iam::430354129336:role/govuk-concourse-deployer'
+        platform: linux
+        image_resource:
+          type: docker-image
+          source:
+            repository: governmentpaas/terraform
+            tag: terraform-0.13.3
+            username: ((docker_hub_username))
+            password: ((docker_hub_authtoken))
+        run:
+          dir: govuk-infrastructure/terraform/deployments/monitoring-test/infra
+          path: sh
+          args:
+          - '-c'
+          - |
+            set -eu
+
+            terraform init -backend-config "role_arn=$ASSUME_ROLE_ARN"
+            terraform apply \
+              -var "assume_role_arn=$ASSUME_ROLE_ARN" \
+              -var-file ../../variables/test/common.tfvars \
+              -var-file ../../variables/test/infrastructure.tfvars \
+              -auto-approve
+
+  - name: deploy-grafana
+    plan:
+    - get: govuk-infrastructure
+      passed:
+      - deploy-monitoring-infra
+      trigger: true
+    - task: update-grafana-task-definition
+      config:
+        inputs:
+        - name: govuk-infrastructure
+          path: src
+        outputs:
+        - name: terraform-outputs
+        params:
+          AWS_REGION: eu-west-1
+          ASSUME_ROLE_ARN: 'arn:aws:iam::430354129336:role/govuk-concourse-deployer'
+          GOVUK_ENVIRONMENT: test
+        platform: linux
+        image_resource:
+          type: docker-image
+          source:
+            repository: governmentpaas/terraform
+            tag: terraform-0.13.3
+            username: ((docker_hub_username))
+            password: ((docker_hub_authtoken))
+        run:
+          path: sh
+          args:
+          - '-c'
+          - |
+            set -eu
+
+            root_dir=$(pwd)
+
+
+            APP_DIR="src/terraform/deployments/monitoring-test/grafana"
+            cd ${APP_DIR}
+
+            terraform init -backend-config="role_arn=$ASSUME_ROLE_ARN"
+
+            terraform apply \
+            -var "assume_role_arn=$ASSUME_ROLE_ARN" \
+            -auto-approve
+
+            terraform output task_definition_arn > "$root_dir/terraform-outputs/grafana_task_definition_arn"
+
+    - task: update-grafana-service
+      config:
+        inputs:
+        - name: terraform-outputs
+        params:
+          APPLICATION: grafana
+          AWS_REGION: eu-west-1
+          ASSUME_ROLE_ARN: 'arn:aws:iam::430354129336:role/govuk-concourse-deployer'
+        platform: linux
+        image_resource:
+          type: docker-image
+          source:
+            repository: governmentpaas/awscli
+            tag: latest
+            username: ((docker_hub_username))
+            password: ((docker_hub_authtoken))
+        run:
+          path: sh
+          args:
+          - '-c'
+          - |
+            set -eu
+
+            mkdir -p ~/.aws
+
+            cat <<EOF > ~/.aws/config
+            [profile default]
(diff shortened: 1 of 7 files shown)


concourse/pipelines/monitoring.yml
```diff
@@ -0,0 +1,287 @@
...
+            root_dir=$(pwd)
+            echo "${root_dir}"
+
+            INFRA_DIR="${root_dir}/src/terraform/deployments/monitoring-test/infra"
+            cd ${INFRA_DIR}
+            terraform init -backend-config "role_arn=$ASSUME_ROLE_ARN"
+            grafana_fqdn=$(terraform output grafana_fqdn)
+            echo "Grafana base URL: https://${grafana_fqdn}"
+
+            wait_url() {
```
sengi, line 232: It shouldn't be necessary to reimplement readiness checking like this - the point of having healthchecks on the ECS service is to take care of this. It's ok for a quick demo but we'll need to do better for prod. It looks like you already have a healthcheck on Grafana - is it not doing what it needs to? i.e. is `wait services-stable` returning before Grafana is healthy?

terraform/deployments/monitoring-test/grafana/app-config/grafana_config_terraform_results
```diff
@@ -0,0 +1,9 @@
+[0m[1mdata.terraform_remote_state.monitoring: Refreshing state...[0m
```
sengi, line 1: Did you mean to commit this file?

concourse/pipelines/monitoring.yml
```diff
@@ -0,0 +1,285 @@
...
+              done;
+            }
+
+            wait_url https://${grafana_fqdn}
+
+            APP_DIR="${root_dir}/src/terraform/deployments/monitoring-test/grafana/app-config"
+            cd ${APP_DIR}
+
+            terraform_apply() {
+              # sed is used here to filter out the ansi colors in the text stream
```
sengi, line 263: I was about to mention that there's a [machine-readable output format](https://www.terraform.io/docs/internals/json-format.html) (`terraform ... -out $file`) then I saw that you're parsing the error messages rather than the actual output. That's both brittle and unnecessary - all you need to do is retry the `terraform apply` (and the clearing state workaround) if it exited nonzero the first time. No need for all this brittle stuff with sed and tee and temporary files.

sengi:
Design question: as far as I can tell, this doesn't address the case where the Grafana task gets rescheduled. In practical terms that's ok temporarily, but it'll need to be addressed before this goes into the wild. So given that it'll need to cope with task restarts sooner or later, surely you're going to need to run the Terraform loading of the Grafana config as part of the Grafana ECS task/service (a bit like the local docker-compose thing does), rather than doing it at deploy time? With this arrangement here, surely Concourse would have to detect when a Grafana task gets rescheduled and trigger the job to load the config (not sensible at all, because the availability of a running service shouldn't depend on the CD system).

sengi:
Unless the idea is to accept all the complexity of having a database for Grafana to persist its temporary config. That seems like a lot of unnecessary complexity for something that ought to be stateless though - plus a bunch of extra risk around users coming to rely on that temporary state, what happens when it goes out of sync with the Terraform code that's checked in, etc etc.

terraform/deployments/monitoring-test/grafana/app-config/grafana_config_terraform_results
fredericfran-gds: nope, good catch. It was more for debugging.

concourse/pipelines/monitoring.yml
fredericfran-gds: If it is suitable to run the workaround on all errors, then I can do that.

fredericfran-gds:
For the design questions about persistence, there was a chat with you and Richard. I created this card: https://trello.com/c/qe4qD3tv/165-persist-grafana-config .

It mentions 3 potential solutions:
1. AWS database
2. AWS EFS
3. Baked dashboard configs in Docker image

concourse/pipelines/monitoring.yml
bilbof: I think `ecs wait services-stable` should probably work here. It's not just a check for the correct task count.

It waits until the query returns true. The query is [roughly](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/ecs/wait/services-stable.html):

```
length(service.deployments) == 1 && service.runningCount == service.desiredCount
```

Services have a [deployment](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Deployment.html) list. It looks like each task definition associated with a service is represented by a deployment:

```
            "deployments": [
                {
                    "id": "ecs-svc/3010894344502080939",
                    "status": "PRIMARY",
                    "taskDefinition": "arn:aws:ecs:eu-west-1:430354129336:task-definition/frontend:201",
                    "desiredCount": 1,
                    "pendingCount": 0,
                    "runningCount": 1,
                    ...
```

So the service is stable when:

- the running count of the service matches the desired count, and
- the service has only one associated task definition (i.e. the deployment has completed, the INACTIVE state task definition no longer appears in the list).
 
I'm not sure if this is a guarantee that the new task is registered and healthy with the ALB target group. Perhaps we should open a ticket with AWS support to confirm this, but I think when the old instance is killed (i.e. runningCount == desiredCount) then the new instance should already be registered with the target group, otherwise this wouldn't be a zero downtime deploy for requests coming through the target group. It would be bizarre, but it wouldn't surprise me if it worked this way, it is very poorly documented.

If it is the case that the ALB health check will trigger ECS to kill tasks after the deploy completes, then maybe you could use a combination of `ecs wait services-stable` (deploy finished) and [elbv2 wait target-in-service](https://docs.aws.amazon.com/cli/latest/reference/elbv2/wait/target-in-service.html) (example below) but I think that probably points to us doing something wrong, if we are 1) bringing up a new instance, 2) killing the old one, and then 3) killing the new one.

```
new_task_ips=$(aws ecs describe-tasks ...)
target_group_arn=$(terraform output public_alb_target_group_arn)
aws elbv2 wait target-in-service \
  --target-group-arn $target_group_arn \
  --targets $new_task_ips
```

(Nb on this example, I can imagine this could be quite flaky, it is a good argument for baking the Grafana config into the docker image if we needed to do something like it).

concourse/pipelines/monitoring.yml
```diff
@@ -0,0 +1,285 @@
...
+                | sed 's/\x1b\[[0-9;]*m//g' \
+                | tee $root_dir/terraform-outputs/grafana_config_terraform_results
+            }
+
+            terraform init -backend-config "role_arn=$ASSUME_ROLE_ARN" -reconfigure
+            (terraform_apply) || true
+
+            # See bug https://github.com/grafana/terraform-provider-grafana/issues/129
+            # where terraform does not recreate some grafana resources if not found
+            if grep -q 'Error: status: 404, body: {"message":"Data source not found"}' "${root_dir}/terraform-outputs/grafana_config_terraform_results"; then
```
bilbof, line 274: 😢 I wonder if we can fix that bug upstream?

concourse/pipelines/monitoring.yml
bilbof: Is it possible to simply run `terraform state rm module.grafana-app-config.grafana_data_source.cloudwatch` before every `terraform apply` of the grafana configuration? Rather than parsing errors and doing retries, we could just remove it from the state every time.

bilbof (dismissed):
Great stuff, thank you for doing this 👍

I have a few questions and comments inline which would be good to address.

It's unfortunate we have to give Concourse access to Grafana, and have three Terraform applies to run Grafana. We can probably iron those out by looking at [persisting the Grafana config](https://trello.com/c/qe4qD3tv/165-persist-grafana-config) somehow (I'm leaning towards baking the config into the image) and seeing if we can use the [AWS Managed Grafana](https://trello.com/c/bFmlencF/194-can-we-use-aws-managed-grafana).

It would be worth looking into whether `aws ecs wait services-stable` is a good enough health check during deploys. I've created [a story](https://trello.com/c/nvwMSHFG/195-document-the-behaviour-of-ecs-deploys) for that.

concourse/pipelines/monitoring.yml
fredericfran-gds: `terraform state rm module.grafana-app-config.grafana_data_source.cloudwatch` can only be done if terraform was run before. If in clean-slate, we can't do that.

fredericfran-gds:
I already done `ecs wait services-stable` in 179. Hopefully, it will be enough.

bilbof (approved):
Thanks 👍

PR #130, opened 2021-01-15
Allow ingress from statsd in ECS to graphite in EC2

bilbof:
This adds the govuk_management_access security group to the statsd ECS service.

Graphite (hosted in EC2) has two security groups
• govuk_graphite_access for ELB ingress
• govuk_management_access for ingress from Statsd

We plan to remove govuk_management_access from govuk-infrastructure in future, as it is poorly named and is a bit of a bag of rules.


fredericfran-gds (approved):
looks good

PR #132, opened 2021-01-15
Promote modules/govuk to a deployment

richardTowers:
Previously we had the govuk-test deployment, which was supposed to
configure things which were specific to the test environment and then
pass them down to the govuk module.

But then we ended up pulling (almost) all of the environment specific
configuration out into .tfvars files in variables/test/*.tfvars, so the
govuk-test deployment ended up not knowing very much about the test
environment. Instead, it just became a really annoying layer of
indirection - any variable or outputs needed to be passed through both
the deployment and the module.

Removing this layer of indirection should make our lives much easier.

The terraform plan for this is going to be an absolute abomination,
but I think we should be able to fix that with:

```
terraform state list | while read -ra line; do
  # If the line starts with `module.govuk`
  if grep "^module\.govuk\." <<< "$line"; then
    # Then move it in the statefile to a location without the module.govuk. prefix
    terraform state mv "$line" "${line#module.govuk.}"
  fi
done
```


terraform/deployments/govuk-test/outputs.tf
```diff
@@ -2,26 +2,6 @@ output "private_subnets" {
...
-  value = module.govuk.signon_security_groups
-}
-
-output "content-store_security_groups" {
-  value = module.govuk.content_store_security_groups
-}
-
-output "draft-content-store_security_groups" {
-  value = module.govuk.draft_content_store_security_groups
-}
```
bilbof, line 23: OK to remove these for now, but we are likely to need them or an equivalent later on.

The reason for these is that unscheduled tasks in ECS must have the same security groups as the ECS Service they would be associated with (e.g. for db migrations, but also for ad-hoc rake tasks like `tools/run-task.sh`).

terraform/deployments/govuk-publishing-platform/outputs.tf
```diff
@@ -2,26 +2,6 @@ output "private_subnets" {
...
-  value = module.govuk.signon_security_groups
-}
-
-output "content-store_security_groups" {
-  value = module.govuk.content_store_security_groups
-}
-
-output "draft-content-store_security_groups" {
-  value = module.govuk.draft_content_store_security_groups
-}
```
bilbof, line 23: OK to remove these for now, but we are likely to need them or an equivalent later on.

The reason for these is that unscheduled tasks in ECS must have the same security groups as the ECS Service they would be associated with (e.g. for db migrations, but also for ad-hoc rake tasks like tools/run-task.sh).

bilbof (approved):
Excellent change 👏

terraform/deployments/govuk-publishing-platform/outputs.tf
richardTowers: Ah, I didn't realise that. I'll put them back.

bilbof: That's my bad - sorry, they should've had descriptions on them.

richardTowers: They're back now :)

bilbof (approved):
Great stuff 🥇

PR #133, opened 2021-01-15
Remove unnecessary _service suffix from app modules

richardTowers:
This resolves a long standing TODO, and makes things a little bit
less tedious.


bilbof (approved):
Excellent 👍

PR #134, opened 2021-01-16
Inline app modules

richardTowers:
These modules don't do anything, other than add a layer of indirection between the `govuk-publishing-platform` deployment and the `app` module.

Removing the indirection removes a confusing layer of variables and outputs, and should make it easier to maintain the codebase.

This PR covers all the apps except frontend, publisher and static - those ones have inline load balancers which make the refactoring slightly more annoying (though not impossible). I'll cover them in a separate PR.

This PR is just moving things about and doesn't change any resources. I've confirmed this by moving the modules in the terraform statefile around (using [`terraform state mv`](https://www.terraform.io/docs/commands/state/mv.html)) and running a terraform plan, which is a no-op.

I've paused the concourse pipelines for now because the statefile surgery will make plans from other branches very confusing.

It should be much easier to review this by looking at the individual commits.


richardTowers:
[Noop terraform plan from this branch](https://cd.gds-reliability.engineering/teams/govuk-ci/pipelines/ci-govuk-infrastructure-main/jobs/terraform-plan-govuk-deployment/builds/102)

terraform/deployments/govuk-publishing-platform/hacks.tf
```diff
@@ -0,0 +1,5 @@
+locals {
```
bilbof, line 1: Finally, one file where I can put all of my code.

bilbof (approved):
Thanks 👍

PR #135, opened 2021-01-18
Add descriptions to publishing platform outputs

bilbof:
This should make the purpose of outputting these security groups clearer.

You can supply a network configuration to tasks, including security
groups, which enables us to give the same permission to ad-hoc tasks
as we would if we were running the tasks in a service.

The security groups are necessary, since we'd like ad-hoc tasks to be
able to reach the DBs of the respective services (e.g. for db migrations).


sengi (approved):
Cool thanks for adding these.

PR #137, opened 2021-01-18
Fix git resource path globs

richardTowers:
`blah/**/*` means "files in directories inside blah", not "everything
inside blah" like I thought originally. https://github.com/concourse/git-resource/issues/36

Turns out we can just specify the directory without the globs, which
works properly and is less confusing to read.

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -37,11 +37,11 @@ resources:
     source:
       <<: *govuk-infrastructure-source
       paths:
-        - concourse/tasks/**/*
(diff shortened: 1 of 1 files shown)


richardTowers:
I tested this by doing a manual `fly set-pipeline`. Before, the git resources didn't find any matching versions. Afterwards, they did.

bilbof (approved):
Cool, nice change 👍

richardTowers:
Oops, everything in the pipeline went orange! 👀

bilbof (approved):

PR #138, opened 2021-01-19
Use the public-load-balancer module for frontend

richardTowers:
This reduces a big chunk of duplication, and will make my next
refactoring a little bit easier.

I did some terraform state moves to get a better feel for what this will
actually do. It looks like most of the resources will need replacing, but
for the most part the changes look quite small.

See [the terraform plan in concourse](https://cd.gds-reliability.engineering/teams/govuk-ci/pipelines/ci-govuk-infrastructure-main/jobs/terraform-plan-govuk-deployment/builds/114) for more details on what this will do.

### DIFF
--- a/terraform/deployments/govuk-publishing-platform/app_frontend.tf
+++ b/terraform/deployments/govuk-publishing-platform/app_frontend.tf
@@ -10,6 +10,7 @@ module "frontend" {
   desired_count                    = var.frontend_desired_count
   public_subnets                   = local.public_subnets
   public_lb_domain_name            = var.public_lb_domain_name
+  office_cidrs_list                = var.office_cidrs_list
 }
 
 module "draft_frontend" {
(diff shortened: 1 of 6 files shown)


sengi (approved):
It kills me that `office_cidrs_list` is spreading again but that's kinda unavoidable for now. Thanks for cleaning this up.

richardTowers:
> It kills me that office_cidrs_list is spreading again but that's kinda unavoidable for now. Thanks for cleaning this up.

I'll be able to remove one set of these variables when I inline modules/apps/frontend. I don't know where we got to with our Network ACL thinking? There was something tricky about it I think?

Anyway, good enough for now :)

sengi:
Yeah, I've only myself to blame for not finishing the ugly network ACL thing to be fair.

PR #139, opened 2021-01-19
Use public lb module for publisher and static

richardTowers:
These are the last two apps which duplicate the load balancer configuration instead of using the public-load-balancer module.

I haven't done any statefile updates, so the terraform plan will be "remove all the load balancers and add new ones". That might make it a bit tricky to work out what's changed from the plan, so let me know if you want me to do the statefile wrangling. On the other hand, it should be enough just to check that there's nothing unusual in the inline load balancer configuration that I'm removing.

### DIFF
--- a/terraform/modules/apps/publisher/main.tf
+++ b/terraform/modules/apps/publisher/main.tf
@@ -21,7 +21,7 @@ module "app" {
   extra_security_groups            = [var.govuk_management_access_sg_id, var.mesh_service_sg_id]
 
   load_balancers = [{
-    target_group_arn = aws_lb_target_group.public.arn
+    target_group_arn = module.public_alb.target_group_arn
     container_name   = "publisher-web"
     container_port   = 80
   }]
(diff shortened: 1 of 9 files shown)


terraform/modules/apps/publisher/main.tf
```diff
@@ -48,76 +48,16 @@ module "worker" {
...
+  app_name                  = var.service_name
+  vpc_id                    = var.vpc_id
+  dns_a_record_name         = var.service_name
+  public_subnets            = var.public_subnets
+  app_domain                = var.public_lb_domain_name # TODO: Change to app_domain
+  public_lb_domain_name     = var.public_lb_domain_name
+  workspace_suffix          = "govuk" # TODO: Changeme
+  service_security_group_id = module.app.security_group_id
+  external_cidrs_list       = var.office_cidrs_list
+}
```
sengi, line 63: 🧐

sengi (approved):
Thanks for the cleanup. Go for it 👍

PR #142, opened 2021-01-19
Fix healthchecks

richardTowers:
I broke the healthchecks for frontend and static in #139.

publisher has a /healthcheck endpoint, but the other apps don't.

This sets them back to the values they had before (which are not great, but at least they work).


fredericfran-gds (approved):
looks good to me

PR #143, opened 2021-01-20
Run smoke tests in ECS after a frontend deploy

bilbof:
This will trigger a smokey run in ECS after every deploy of the frontend app.

I also spiked running the smoke tests directly from Concourse. This required fewer code changes, but would require adding Concourse IPs to allowlists in Fastly / AWS. Also, we'll still need to run smokey as a ScheduledTask from ECS (since we don't want to rely on Concourse availability for our alerting), so it made sense to run the post-deploy smoke test run from ECS too.

We can implement the 'Smokey loop' as an ECS ScheduledTask later.

The smoke tests are a subset of the test suite in the test environment called 'replatforming', which are just the tests we expect to run successfully for the test environment.

In future, I expect we'll change this to 'app-frontend', which will run all the smoke tests that would be likely to fail should frontend introduce a bug or some other issue. Currently most frontend tests fail.

There are also a couple of minor improvements to the way we run tasks from Concourse, so I would review by commit.

**Example smoke test run**

Successful run of smoke test


concourse/tasks/run-task.sh
```diff
@@ -31,26 +31,21 @@ echo "Starting task..."
...
-aws ecs wait tasks-stopped --tasks=[$task_arn] --cluster $CLUSTER
-
+aws ecs wait tasks-stopped --tasks="$task_id" --cluster $CLUSTER
 echo "task finished."
-
-task_results=$(aws ecs describe-tasks --tasks=[$task_arn] --cluster $CLUSTER)
-echo $task_results
-
+task_results=$(aws ecs describe-tasks --tasks=$task_id --cluster $CLUSTER)
+ecs-cli logs --cluster $CLUSTER --task-id=$task_id --since "60"
```
fredericfran-gds, line 48: can you explain why you choose 60 minutes?

concourse/tasks/run-task.sh
```diff
@@ -31,26 +31,21 @@ echo "Starting task..."
...
 
-task_arn=$(echo $task | jq .tasks[0].taskArn)
-
+task_arn=$(echo $task | jq .tasks[0].taskArn -r)
+task_id=$(basename $task_arn)
 echo "waiting for task $task_arn to finish..."
-
-aws ecs wait tasks-stopped --tasks=[$task_arn] --cluster $CLUSTER
-
+aws ecs wait tasks-stopped --tasks="$task_id" --cluster $CLUSTER
```
fredericfran-gds, line 45: style: will it be still valid that we use space as separator, --tasks="$task_id" -> --tasks "$task_id" ? Just to keep consistency.

concourse/pipelines/deploy.yml
```diff
@@ -331,16 +331,22 @@ jobs:
 
   - name: smoke-test-frontend
     plan:
-    - get: govuk-infrastructure
-      resource: govuk-infrastructure-frontend
-      passed:
-      - deploy-frontend
-      trigger: true
-    - file: govuk-infrastructure/concourse/tasks/basic-smoke-test.yml
+    - in_parallel:
```
fredericfran-gds, line 334: does this `in_parallel` useful here, there seems to be only one resource being retrieved.

fredericfran-gds (commented):
Great work as usual.

One question is where is the Smokey task definition first created? I saw it being retrieved but not created. Maybe I'm not looking in the right place.

concourse/tasks/run-task.sh
bilbof: It's an arbitrary length of time that's longer than the default `since` value (0) and the length the smoke tests should realistically take.

fredericfran-gds: cool, thanks.

concourse/pipelines/deploy.yml
bilbof: Thanks, I've removed this.

concourse/tasks/run-task.sh
bilbof: Thanks, I've removed the `=` from them 👍

bilbof:
@fredericfran-gds thanks for your review!

> One question is where is the Smokey task definition first created?

I thought this was out of scope for this PR - Smokey doesn't have a CI/release pipeline, so we need to add this. I've added a [followup task](https://trello.com/c/hHyHGpCx/182-ci-pipeline-for-smokey) for this.

fredericfran-gds (approved):
happy to approve, thanks

PR #144, opened 2021-01-22
Inline the frontend module

richardTowers:
This gets rid of another layer of variables / outputs which we have to
pass around.

It should be possible to confirm this is a no-op by doing some `terraform
state mv` commands before running the plan.


fredericfran-gds (dismissed):
cool

richardTowers:
No-op plan: https://cd.gds-reliability.engineering/teams/govuk-ci/pipelines/ci-govuk-infrastructure-main/jobs/terraform-plan-govuk-deployment/builds/134

(I'd made a tiny mistake, which I corrected in https://github.com/alphagov/govuk-infrastructure/compare/22fe0e61f2db1ec431b0790bcbce706d1ce5fc84..e8c2b28d3ceceb0d11ddcc90f93377725aa35438)

fredericfran-gds (approved):
lol, I thought you checked the no-op already.

richardTowers:
😅  no, I only opined that "it should be possible to confirm this is a no-op" ... I didn't actually confirm it! (But I have now)

PR #145, opened 2021-01-22
Build & Push Smokey container image from Concourse

bilbof:
This adds a CI pipeline that:

1. detects new commits to the main branch
2. determines a new semantic version (version stored in its S3 bucket + 1 minor version)
3. builds a new container image
4. tags the image with `latest` and the new version
5. pushes the image to DockerHub
6. adds a tag to the commit on the GitHub repo, prefixed with "v" (e.g. v1.2.0)
7. updates S3 bucket containing the [semver-resource](https://github.com/concourse/semver-resource)

I expect we can copy this procedure for other pipelines - PaaS made their [integration test pipeline config](https://github.com/alphagov/paas-release-ci/blob/f6bef1a0cc3da5290e402506cf2e4e477817233f/pipelines/integration-test.yml) generic with variables - we can reuse the same config for multiple pipelines that do the same thing.

Tagging the commit is useful since we can watch for new version tags and trigger other builds.

I've added a pipeline: https://cd.gds-reliability.engineering/teams/govuk-ci/pipelines/build-images

Builds pipeline

### DIFF
--- a/concourse/pipelines/build-images.yml
+++ b/concourse/pipelines/build-images.yml
@@ -1,0 +1,90 @@
+---
+definitions:
+
+resources:
+  - name: smokey
+    icon: github
+    type: git
+    source:
+      uri: git@github.com:alphagov/smokey.git
+      branch: main
+      private_key: |
+        ((govukci_private_key))
+
+  - name: govuk-infrastructure
+    icon: github
+    type: git
+    source:
+      uri: git@github.com:alphagov/govuk-infrastructure.git
+      branch: main
+      private_key: |
+        ((govukci_private_key))
+
+  - name: smokey-image
+    type: registry-image
+    icon: docker
+    source:
+      repository: govuk/smokey
+      tag: latest
+      username: ((docker_hub_username))
+      password: ((docker_hub_authtoken))
+
+  - name: smokey-version
+    type: semver
+    source:
+      driver: s3
+      access_key_id: ((readonly_access_key_id))
+      secret_access_key: ((readonly_secret_access_key))
+      session_token: ((readonly_session_token))
+      bucket: ((readonly_private_bucket_name))
+      key: smokey-version
+      region_name: eu-west-2
+      initial_version: '1.0.0'
+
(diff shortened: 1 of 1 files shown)


concourse/pipelines/build-images.yml
```diff
@@ -0,0 +1,89 @@
...
+  - name: smokey
+    plan:
+    - in_parallel:
+      - get: smokey-commit
+        trigger: true
+      - get: smokey-version
+        params:
+          bump: minor
+    - task: build-image
+      privileged: true
```
richardTowers, line 61: Do we need `privileged: true` here?

bilbof: Yeah, though perhaps not forever: https://github.com/vito/oci-build-task#oci-build-task

> A stretch goal of this is to support running without privileged: true, though it currently still requires it.

concourse/pipelines/build-images.yml
```diff
@@ -0,0 +1,89 @@
+---
+definitions:
+
+resources:
+  - icon: github
+    name: smokey-commit
```
richardTowers, line 6: I think it's nice if the name comes first in these blocks (even thought it's equivalent).

I wonder if `smokey` would be just as good a name as `smokey-commit` (I wonder about the `-commit` suffix, as it's not really the _commit_ you're fetching, you're getting the whole repo. So `smokey-repo`? But I think just `smokey` is enough. We should probably have a styleguide / linter 😅 )

```suggestion
  - name: smokey-commit
    icon: github
```

concourse/pipelines/build-images.yml
richardTowers: "A stretch goal of this is to support running without privileged: true, though it currently still requires it." so... yes:

https://github.com/vito/oci-build-task

That's no problem then :)

richardTowers (dismissed):
Some nitpicks, but otherwise looks good to me.

richardTowers (approved):

PR #146, opened 2021-01-22
Create a smokey task definition for new versions

bilbof:
This will watch the smokey repo for new version tags added to commits on the main branch and create a new task definition for smokey.

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -151,6 +151,23 @@ resources:
         - terraform/modules/task-definitions/signon
 
   - <<: *git-repo
+    name: smokey
+    source:
+      branch: main
+      uri: https://github.com/alphagov/smokey
+      tag_filter: "*-release"
+
+  - <<: *git-repo
+    name: govuk-infrastructure-smokey
+    source:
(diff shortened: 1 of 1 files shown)


concourse/pipelines/deploy.yml
```diff
@@ -150,6 +150,23 @@ resources:
         - terraform/modules/task-definition
         - terraform/modules/task-definitions/signon
 
+  - <<: *git-repo
+    name: smokey
+    source:
+      branch: main
+      uri: https://github.com/alphagov/smokey
+      tag_filter: "*-release"
```
bilbof, line 158: I'm not completely happy with the tagging, we could make a change to the semver resource to support version prefixes (e.g. "v" or "release_"). At the moment adding a suffix to a tag (i.e. "1.2.3-release") is the best the resource can do.

richardTowers: It would probably be nicer to use a `registry-image` resource for smokey, instead of cloning the git repo just to work out what the tag of the image in the registry is 😄. I understand that it has to be like this for now to be compatible with the other apps though, so I think this is fine.

richardTowers (approved):

PR #147, opened 2021-01-25
Inline modules/apps/publisher

richardTowers:
This is a no-op refactoring, see the plan here:

https://cd.gds-reliability.engineering/teams/govuk-ci/pipelines/ci-govuk-infrastructure-main/jobs/terraform-plan-govuk-deployment/builds/141

In-lining this module reduces the amount of variable / output passing we have to do.

There's only one app module (static) to inline after this one.


richardTowers:
I'm going to close this and raise another one that includes static.

PR #148, opened 2021-01-26
Inline remaining "apps/" modules (publisher and static)

richardTowers:
These modules weren't providing enough value to justify their existence, so we're removing them. This removes an annoying layer of passing variables down and outputs up to and from calling modules.

These were the last two modules in modules/apps, so the entire directory is gone now. I've updated the docs and README so they don't refer to these modules anymore.

Hopefully having these modules gone should make a lot of our future work easier and less confusing.


richardTowers:
The Terraform plan shows this refactoring has no effect on the resources created:

https://cd.gds-reliability.engineering/teams/govuk-ci/pipelines/ci-govuk-infrastructure-main/jobs/terraform-plan-govuk-deployment/builds/143

```
------------------------------------------------------------------------


No changes. Infrastructure is up-to-date.


This means that Terraform did not detect any differences between your

configuration and real physical resources that exist. As a result, no

actions need to be performed.

```

fredericfran-gds (approved):
looks good, thanks for checking the no op.

PR #149, opened 2021-01-27
Add ECR repositories for GOV.UK apps

rtrinque:
Add ECR repositories for GOV.UK apps

### DIFF
--- a/terraform/deployments/ecr/main.tf
+++ b/terraform/deployments/ecr/main.tf
@@ -1,0 +1,112 @@
+# This is supposed to be applied in the production environment
+
+terraform {
+  backend "s3" {
+    bucket  = "govuk-terraform-steppingstone-production"
+    key     = "govuk/ecr.tfstate"
+    region  = "eu-west-1"
+    encrypt = true
+  }
+
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 3.13"
+    }
+  }
+}
+
+provider "aws" {
+  region = "eu-west-1"
+}
+
+
+resource "aws_ecr_repository" "content-store" {
+  name                 = "content-store"
+  image_tag_mutability = "MUTABLE"
+
+  image_scanning_configuration {
+    scan_on_push = true
+  }
+}
+
+resource "aws_ecr_repository" "frontend" {
+  name                 = "frontend"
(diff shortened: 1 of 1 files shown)


terraform/deployments/ecr/main.tf
```diff
@@ -0,0 +1,112 @@
...
+    }
+  }
+}
+
+provider "aws" {
+  region = "eu-west-1"
+}
+
+
+resource "aws_ecr_repository" "content-store" {
```
richardTowers, line 24: Do we want one repository per app? Can we not put all the apps in the same repository?

terraform/deployments/ecr/main.tf
```diff
@@ -0,0 +1,112 @@
+# This is supposed to be applied in the production environment
+
+terraform {
+  backend "s3" {
+    bucket  = "govuk-terraform-steppingstone-production"
```
richardTowers, line 5: I'm not sure about putting the terraform state file in the bucket that govuk-aws uses... I worry that could be confusing.

On the other hand, we haven't created a `govuk-terraform-production` bucket yet, because the terraform in this repo hasn't run against production...

🤷🏻  probably not a major problem... I'd be happy to stick with this so long as we've considered it.

terraform/deployments/ecr/main.tf
rtrinque: That's what I wanted to do originally but that would cause problems with tags as a tag can only exist once per repository.
https://stackoverflow.com/questions/46230838/ecr-what-is-the-registry-and-what-is-a-repository

terraform/deployments/ecr/main.tf
rtrinque: Yeah I only used it for the sake of expediency as it was already there and we don't really deal with environments yet.

terraform/deployments/ecr/main.tf
richardTowers: Today I learned! Thanks for explaining that, it makes sense to me now.

terraform/deployments/ecr/main.tf
richardTowers: 🤔 I guess it's fine for now then.

richardTowers (approved):

PR #150, opened 2021-01-27
Simplify app deployments using aws ecs register-task-definition

richardTowers:
Having separate terraform deployments for each application results in a lot of boilerplate, and maintenance headaches. Instead, we can simplify the pipeline by emitting the JSON required to register a task definition from the main terraform, and passing that to `aws ecs register-task-definition` using the `--cli-input-json` flag.

This PR might be a little tricky to review from a cold start, so let me explain it as a play in two acts. I'm going to be avante garde and work backwards, from the end to the beginning. I think it's easier to understand it that way (but if you don't like it, the individual commits have a dry and boring description of the work).

## Act II: In which concourse deploys the app

_Enter stage left: Concourse_
_Enter stage right: the AWS CLI_

image

The first file to look at is [concourse/tasks/update-task-definition.sh](150/files#diff-801a4a80f3bb3539732ad171b1ba966ef97c806a7cab52c2b712b6b44f461dfe).

This shell script runs:

```bash
aws ecs register-task-definition \
  --cli-input-json "file://task-definition.json"
```

Which is the absolute core of the idea in this PR. This replaces the old `terraform apply` approach, which the old version of update-task-definition used.

Where does it get `file://task-definition.json` from though?

That's a slightly modified version of `app-terraform-outputs/content-store.json`, which is provided as an input to the task.

The modification is done by [`jq`](https://stedolan.github.io/jq/), which fishes out the right task definition for this variant (`live` or `draft`) and overrides the docker image in the first container definition:

```bash
jq \
  ".${VARIANT}.task_definition_cli_input_json | .containerDefinitions[0].image = \"${IMAGE}\"" \
  "app-terraform-outputs/${APPLICATION}.json" \
  > task-definition.json
```

When the task finishes, it writes the ARN of the new task definition to an output, which is then used by the `update-ecs-service` to update the service. I've created a `v2` version of `update-ecs-service` in this PR too, but the differences to what was there before are mostly cosmetic.

Where does `app-terraform-outputs` come from though?

It comes from [a new S3 resource](150/files#diff-366d145c8de296d618d368829044ef16eeea1b060b814025c9aceea901fe2c96R185-R195), `content-store-terraform-outputs`. This is [provided as a `get` step to the deploy-content-store job](150/files#diff-366d145c8de296d618d368829044ef16eeea1b060b814025c9aceea901fe2c96R513-R514), having been [`put` by the run-terraform job](150/files#diff-366d145c8de296d618d368829044ef16eeea1b060b814025c9aceea901fe2c96R330-R332).

The run-terraform job creates the file in `content-store-terraform-outputs` by running:

```
terraform output -json content-store > "$root_dir/updated-content-store-terraform-outputs/content-store.json"
```

There is one tricky issue that I had to work around here though. By default, every `put` to the S3 resource creates a new version, even if the contents of the file is identical. That would mean redeploying content-store after every single run of terraform, even though most of the time that's not needed.

I've worked around this by passing `content-store-terraform-outputs` _in_ to run-terraform, and then [comparing the new terraform outputs with the old ones](150/files#diff-366d145c8de296d618d368829044ef16eeea1b060b814025c9aceea901fe2c96R316-R319). If the outputs are exactly the same, then I [remove the updated version](150/files#diff-366d145c8de296d618d368829044ef16eeea1b060b814025c9aceea901fe2c96R322) to ensure that the `put` step doesn't succeed in creating a new version. The [`put` step is in a `try` block](150/files#diff-366d145c8de296d618d368829044ef16eeea1b060b814025c9aceea901fe2c96R329-R332), so this can fail without failing the whole run-terraform job.

To understand how the JSON in this terraform output gets created in the first place, we need to think back to Act I.

_Exit: Concourse_
_Exit: the AWS CLI_

## Act I: In which terraform declares how to deploy the app

_Enter stage left: Terraform_

To set things up for the thrilling "Act II", we need our terraform deployment to generate some JSON which is compatible with the AWS CLI's `register-task-definition` command.

I've added this in [modules/app/task_definition.tf](150/files#diff-dba0bbcb376924c8d931ffc18fe1b125fe301deca9cbcfa42a73588af692463e). Note that this doesn't actually create a task definition in AWS, it just templates out the data structure, and provides it as an output.

Because I've added this functionality to the `app` module which is used everywhere, I've had to provide placeholder variables in from all of the apps. Even though these aren't all filled in correctly (e.g. the env vars are blank), this is harmless because the JSON isn't used anywhere for these apps.

The only app that's wired up to generate the task definition JSON properly is content-store. You can see how this works in [`deployments/govuk-publishing-platform/app_content_store.tf`](150/files#diff-ce84dd695036598d706e99f50a03c6d92da47ef5aecb5630e27b6e36fbec7d03). Many of the environment variables are the same for every GOV.UK app, so I've pulled these out into a [`defaults.tf`](150/files#diff-8772fb1082a6658efa92525bb1a413e6b7bb6fd4a4409314a5ab068d16368616) file. Other inputs come from remote state, or from secrets manager. These are also given their own files to keep things clear.

Ultimately, the `govuk-publishing-platform` terraform deployment provides all the details needed for concourse to deploy the content-store (and draft-content-store) applications in one output, which is structured to make it convenient to access - [`outputs.tf`](150/files#diff-4673ffe204d53a25944ce58048ac6326deae24b4354e942155d656ff3c75d98cR20-R32). This allows Concourse to use `jq` in Act II to access the task definition for the appropriate variant (`draft` or `live`).

_Exit, pursued by a bear_

## Epilogue

This has been an adventure, but it's not over yet! There are 8 more apps which need to be refactored to use this new-style deployment approach. Hopefully, this PR lays enough of the ground work that the rest of the refactoring should be much simpler.

For the sake of everyone's sanity, I'm going to split the remaining apps into separate trello cards.

## Trello


concourse/pipelines/deploy.yml
```diff
@@ -291,9 +309,27 @@ jobs:
               -auto-approve
 
             terraform output -json > "$root_dir/govuk-terraform-outputs/govuk-terraform-outputs.json"
+
+            for app in content-store; do
```
fredericfran-gds, line 313: I may have missed something but where is `content-store` defined?

concourse/pipelines/deploy.yml
```diff
@@ -291,9 +309,27 @@ jobs:
...
+              terraform output -json "${app}" > "$root_dir/updated-${app}-terraform-outputs/${app}.json"
+
+              if cmp \
+                "$root_dir/original-${app}-terraform-outputs/${app}.json" \
+                "$root_dir/updated-${app}-terraform-outputs/${app}.json"
```
fredericfran-gds, line 318: semantics: I wonder if `new` is not better than `updated`

concourse/pipelines/deploy.yml
```diff
@@ -291,9 +309,27 @@ jobs:
               -auto-approve
 
             terraform output -json > "$root_dir/govuk-terraform-outputs/govuk-terraform-outputs.json"
+
+            for app in content-store; do
+              terraform output -json "${app}" > "$root_dir/updated-${app}-terraform-outputs/${app}.json"
```
fredericfran-gds, line 314: I wonder if a new task to do the app-based terraform output comparison should be created? Because how will this code be scaled for the dozens of apps that we have? Maybe a parameterizable task would be more appropriate?

concourse/tasks/update-ecs-service-v2.yml
```diff
@@ -0,0 +1,18 @@
+platform: linux
+image_resource:
+  type: docker-image
+  source:
+    repository: governmentpaas/awscli
+    tag: latest
```
fredericfran-gds, line 6: optional: add todo to pin version

concourse/tasks/update-ecs-service.sh
```diff
@@ -0,0 +1,34 @@
...
+
+cat <<EOF > ~/.aws/config
+[profile default]
+role_arn = $ASSUME_ROLE_ARN
+credential_source = Ec2InstanceMetadata
+EOF
+
+new_task_definition_arn="$(cat "task-definition-arn/task-definition-arn")"
+if [ -z "${new_task_definition_arn}" ]; then
+  echo "failed to retrieve new task definition, exiting..."
```
fredericfran-gds, line 15: might be good to add the ECS service name as you did for the log messages below, i.e.
`failed to retrieve a new task definition for $ECS_SERVICE service, exiting...`

concourse/tasks/update-task-definition-v2.yml
```diff
@@ -0,0 +1,23 @@
+platform: linux
+image_resource:
+  type: docker-image
+  source:
+    repository: governmentpaas/awscli
+    tag: latest
```
fredericfran-gds, line 6: optional: add todo to pin version

terraform/deployments/govuk-publishing-platform/outputs.tf
```diff
@@ -17,6 +17,19 @@ output "signon_security_groups" {
   description = "Used by ECS RunTask to run short-lived tasks with the same SG permissions as the signon ECS Service."
 }
 
+output "content-store" {
+  value = {
+    draft = {
+      task_definition_cli_input_json = module.draft_content_store.cli_input_json,
+      security_groups                = module.draft_content_store.security_groups,
```
fredericfran-gds, line 24: based on the concourse pipeline, am I correct to say that there will be a task definition/service update if the security-group is only updated by not the cli_input_json? One would think only the `cli_input_json` warrants a new task definition/service update.

fredericfran-gds (commented):
good work on simplifying the terraform; I like that it is a flatter structure now. I made some comments for your consideration in line.

concourse/pipelines/deploy.yml
richardTowers: Heh, this is really gross code 😆  Shellcheck was telling me off for it to.

`content-store` isn't a variable here, it's the literal string `"content-store"`. The loop is there because so that when we do the next app we can just add it like:

```bash
for app in content-store fronted; do
```

... perhaps that's premature optimisation and I should have started with `app="content-store"` and dropped the loop.

Ideally I would have used an array here, but the shell is `dash`, not `bash`, so there are no arrays.

concourse/pipelines/deploy.yml
richardTowers: `new` / `old` instead of `original` / `updated` ? I could be sold on that - they're a bit shorter at least.

concourse/pipelines/deploy.yml
richardTowers: The intention is just to keep adding app names to the for loop. So it will end up like:

```bash
for app in content-store frontend publisher publshing-api router router-api etc...; do
   terraform output -json "${app}" ...
```

richardTowers: I agree that's a bit disgusting though.

concourse/pipelines/deploy.yml
fredericfran-gds: sorry, I got confused by it being without quotes. I always wondered why they make life so difficult and do not provide bash, does it make the container that much bigger?

terraform/deployments/govuk-publishing-platform/outputs.tf
fredericfran-gds, line 24: by tying both live and draft together, you assume that one draft can't be updated independently from live/vice versa?

fredericfran-gds (commented):
added some comments after you clarified the `for` loop, thanks.

concourse/pipelines/deploy.yml
richardTowers: We're piggy backing on [GOV.UK PaaS's images](https://github.com/alphagov/paas-docker-cloudfoundry-tools/blob/main/terraform/Dockerfile).

We get `dash` because they chose to do `FROM alpine`, which tries really hard to keep itself small. `dash` instead of `bash` is just one part of that, but it is definitely annoying. It's also annoying when you do `fly hijack`, because by default that tries to use `bash`, which then errors because it's not in the container. You have to explicitly tell it to use `sh`.

At some point we'll want to own our own concourse task images, at which point we might decide to use bash as the shell instead...

On the other hand, if I'd had bash here I probably would have done some even more confusing stuff with arrays, so 🤷🏻  maybe it's a good thing?

terraform/deployments/govuk-publishing-platform/outputs.tf
richardTowers: Yeah, this is by design, on the assumption that we won't ever be in the situation where we want to deploy draft independently from live, or web independently from worker.

I _think_ that's more or less the situation with Jenkins at the moment, although I haven't checked.

richardTowers: Yep, you're quite right.

There are some downstream deploys (not for content store, yet, but for publisher) which do depend on the security group though - the run-task thing we're using for DB migrations needs to know the security groups (so the task has permission to talk to the DB).

At the moment, those tasks get deployed in the same job as the task definition, so we want them to trigger when the SGs change as well as when the task definition changes.

concourse/pipelines/deploy.yml
richardTowers: I've changed the for loop to a function in 0c3c0dd - hopefully that's a little bit less confusing?

concourse/pipelines/deploy.yml
richardTowers: Resolved in 9a53cb4

concourse/pipelines/deploy.yml
richardTowers: I've changed the for loop to a function in 0c3c0dd

concourse/tasks/update-ecs-service-v2.yml
richardTowers: 54a0d72

concourse/tasks/update-task-definition-v2.yml
richardTowers: 54a0d72

concourse/tasks/update-ecs-service.sh
richardTowers: 26f34f3

terraform/deployments/govuk-publishing-platform/outputs.tf
richardTowers: Yeah, I had a look at what Jenkins does.

There's just one "Deploy_App" job, which specifies the app (say "content-store"). You don't get to specify draft or live.

Capistrano has a [`set_servers`](https://github.com/alphagov/govuk-app-deployment/blob/master/recipes/set_servers.rb) rake task, which uses `ovuk_node_list -c content-store` to get a list of servers to deploy to. That will include draft and live versions.

So people should already be used to the idea that it's not possible to deploy draft and live independently. Personally, I think this constraint is a good thing - I can't think of a legitimate reason to have them on different versions.

fredericfran-gds: I was more thinking if we want to tweak CPU/memory of live instances because of increased traffic, we may not need to do that for the draft ones. Anyway, we can cross that bridge when we get there if needed.

fredericfran-gds: Ummm... I'm not convinced about this because the security groups to the DBs are set once while the DB migrations happen when the developer introduces a new schema/migration.

fredericfran-gds (approved):
most of my questions have been answered, thanks. I'm happy to approve.

bilbof (approved):
Great work! Let's try it 👍 

This seems like a good approach, given we've struggled with managing task definition revisions/deploys through a mix of Terraform and the AWS CLI. Just using terraform outputs and the AWS CLI hopefully will simplify things a bit (like less directory spelunking in Concourse tasks).

I found the Concourse pipeline a little confusing. It would be great to have some documentation beyond this PR description (perhaps you could move your comment into the `docs` directory?) - not a blocker to merging.

terraform/deployments/govuk-publishing-platform/outputs.tf
richardTowers: Ah, they do get different task definitions, so they don't have to have the same memory / CPU. They just have to be deployed at the same time.

PR #151, opened 2021-02-02
Lint govuk-publishing-platform terraform deployment with tflint

richardTowers:
[tflint](https://github.com/terraform-linters/tflint) does some additional style checks on top of the validations done by `terraform validate` (which we already have a GitHub Action for).

It will catch things like:

```terraform
name = "${var.unnecessary_string_interpolations}"
```

It's configurable, and not all the rules will be the same for all of our deployments. I've written the action so it will only try to lint deployments that have a config file.

For now, I'm just covering the main govuk-publishing-deployment, which looks like it's mostly okay already! 🎉 

Downloading and not caching the 10.3 MB release zip is a bit lazy of me... but 🤷🏻 

Example of a failing action: https://github.com/alphagov/govuk-infrastructure/runs/1817276105?check_suite_focus=true

### DIFF
--- a/terraform/deployments/govuk-publishing-platform/.tflint.hcl
+++ b/terraform/deployments/govuk-publishing-platform/.tflint.hcl
@@ -1,0 +1,18 @@
+plugin "aws" {
+  enabled = true
+}
+
+config {
+  varfile = [
+    "../variables/test/infrastructure.tfvars",
+  ]
+}
+
+rule "terraform_comment_syntax" { enabled = true }
+rule "terraform_deprecated_index" { enabled = true }
+rule "terraform_required_providers" { enabled = true }
+rule "terraform_standard_module_structure" { enabled = true }
+rule "terraform_typed_variables" { enabled = true }
+rule "terraform_unused_declarations" { enabled = true }
+rule "terraform_unused_required_providers" { enabled = true }
+
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,3 +59,23 @@ jobs:
               echo -e '\n-------------------------\n'
             )
           done
+
+      - name: run tflint
+        env:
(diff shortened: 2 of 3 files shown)


karlbaker02 (approved):
Looks good 👍  You may want to consider getting this in before https://github.com/alphagov/govuk-infrastructure/pull/161

richardTowers:
Rebased onto `main` now that #161 has merged.

terraform/deployments/govuk-publishing-platform/.tflint.hcl
```diff
@@ -0,0 +1,18 @@
...
+  ]
+}
+
+rule "terraform_comment_syntax" { enabled = true }
+rule "terraform_deprecated_index" { enabled = true }
+rule "terraform_required_providers" { enabled = true }
+rule "terraform_standard_module_structure" { enabled = true }
+rule "terraform_typed_variables" { enabled = true }
+rule "terraform_unused_declarations" { enabled = true }
+rule "terraform_unused_required_providers" { enabled = true }
```
bilbof, line 17: Is there a way to bring this file up a level so it applies to all deployment modules?

bilbof (approved):
Small question about applying the linting rules across all deployments, but feel free to merge and we can think about it later

bilbof:
Oh and thank you for doing this! 🚀

terraform/deployments/govuk-publishing-platform/.tflint.hcl
richardTowers: At the moment I'm using the existence of this file to mark which deployments need linting. I think they'll all need slightly different config (e.g. monitoring-test/grafana doesn't use the AWS provider, monitoring-test/infra is one level deeper in the filesystem than the others, so relative links to variables would break).

Something we could look at later though.

PR #152, opened 2021-02-03
Refactor Publisher deployment

bilbof:
This rewrites the way we deploy the Publisher app, following on from the refactor of content store in PR #150.

The intent is to simplify the way we deploy publisher (web and worker) and run tasks using the publisher task definition.

This change:
- removes the deployment modules for publisher-web and publisher-worker
- adds logic to the govuk-publishing-platform/publisher_app file to generate the JSON required to register the web and worker task definitions for publisher
- removes the now unnecessary task definition module for publisher
- modifies the deploy pipeline for publisher to register the publisher task definitions using the AWS CLI, rather than Terraform
- modifies the deploy pipeline to store the JSON generated by the big Terraform apply (govuk-publishing-platform) in S3, so we can use it to register task definitions, overriding only the IMAGE key.
- modifies the deploy pipeline so that we can run db migrations for publisher, which will break smokey, and anything that uses run-task temporarily.

This introduces additional complexity into the Concourse pipeline (particularly run-task is not that simple), I'll address this in a future PR.


concourse/pipelines/deploy.yml
```diff
@@ -447,44 +458,54 @@ jobs:
...
     - task: run-db-migrations
       file: govuk-infrastructure/concourse/tasks/run-task.yml
+      input_mapping:
+        task-definition-arn: web-task-definition-arn
       params:
-        APPLICATION: publisher-web
+        APPLICATION: publisher
         COMMAND: "bundle exec rails db:migrate"
-        TASK_DEFINITION: web_task_definition
+        TASK_DEFINITION: web-task-definition-arn
```
richardTowers, line 495: I don't think you need this anymore, now that you've got the `input_mapping`. It doesn't look like run-task.yml refers to this anymore.

terraform/deployments/apps/variables.tf
```diff
@@ -23,12 +23,12 @@ variable "mesh_domain" {
 # Environment
 #--------------------------------------------------------------
 
-variable "app_domain" {
+variable "external_app_domain" {
   type        = string
   description = "Example: staging.publishing.service.gov.uk"
```
richardTowers, line 28: Is this example actually correct? I thought this one was like `test.govuk.digital`, and that we had another variable for the service domain?

terraform/deployments/govuk-publishing-platform/app_publisher.tf
```diff
@@ -1,3 +1,67 @@
...
+        FACT_CHECK_SUBJECT_PREFIX        = "dev"
+        FACT_CHECK_USERNAME              = "govuk-fact-check-test@digital.cabinet-office.gov.uk"
+        GOVUK_APP_DOMAIN_EXTERNAL        = var.external_app_domain
+        GOVUK_APP_NAME                   = local.publisher_app_name
+        GOVUK_APP_ROOT                   = "/app"
+        # TODO: how does GOVUK_ASSET_ROOT relate to ASSET_HOST? Is one a function of the other? Are they both really in use? Is GOVUK_ASSET_ROOT always just "https://${ASSET_HOST}"?
+        GOVUK_ASSET_ROOT    = "https://assets.test.publishing.service.gov.uk"
+        GOVUK_STATSD_PREFIX = "govuk-ecs.app.${local.publisher_app_name}"
+        # TODO: looks suspicious
+        PLEK_SERVICE_CONTENT_STORE_URI  = "https://www.gov.uk/api"
```
richardTowers, line 26: Maybe we could take this opportunity to fix this one?

```suggestion
        PLEK_SERVICE_CONTENT_STORE_URI  = local.defaults.content_store_uri
```

(assuming that's actually set in the defaults, which it probably isn't)

terraform/deployments/govuk-publishing-platform/app_publisher.tf
```diff
@@ -1,3 +1,67 @@
...
+        PLEK_SERVICE_PUBLISHING_API_URI = local.defaults.publishing_api_uri
+        PLEK_SERVICE_SIGNON_URI         = local.defaults.signon_uri
+        PLEK_SERVICE_STATIC_URI         = "https://assets.test.publishing.service.gov.uk"
```
richardTowers, line 29: I think we could sort this one out too.

terraform/deployments/govuk-publishing-platform/app_publisher.tf
```diff
@@ -46,21 +113,36 @@ module "publisher_public_alb" {
...
-  memory                           = 1024
+  cpu                              = local.publisher_defaults.cpu
+  memory                           = local.publisher_defaults.memory
   task_role_arn                    = aws_iam_role.task.arn
   execution_role_arn               = aws_iam_role.execution.arn
+  vpc_id                           = local.vpc_id
+}
+
+module "publisher_network_config" {
+  source  = "../../modules/task-network-config"
```
richardTowers, line 141: At some point, we could probably consider moving the `task-network-config` module into the app module, since every app has some network config.

Out of scope for this PR though.

terraform/deployments/govuk-publishing-platform/outputs.tf
```diff
@@ -40,6 +41,21 @@ output "draft-content-store_security_groups" {
   description = "Used by ECS RunTask to run short-lived tasks with the same SG permissions as the draft-content-store ECS Service."
 }
 
+output "publisher" {
+  value = {
+    web = {
+      task_definition_cli_input_json = module.publisher_web.cli_input_json,
+      # TODO: Use publisher.web.security_groups for ECS RunTask
```
richardTowers, line 48: Could we use `publisher.network_config` in run-task instead?

terraform/modules/monitoring/grafana.tf
```diff
@@ -15,8 +15,8 @@ module "grafana_app" {
 
   load_balancers = [{
     target_group_arn = module.grafana_public_alb.target_group_arn
-    container_name   = local.service_name
-    container_port   = 3000
+    # container_name   = local.service_name
```
richardTowers, line 18: Does this need to be commented out, or can it be removed?

richardTowers (commented):
🎉  looks absolutely amazing, thanks!

I spotted a few minor things, but generally this looks great.

terraform/deployments/govuk-publishing-platform/app_publisher.tf
bilbof: Good idea. I've made a start by moving this block into the app module.

terraform/deployments/govuk-publishing-platform/app_publisher.tf
```diff
@@ -1,3 +1,66 @@
...
+        EMAIL_GROUP_CITIZEN              = "test-address@digital.cabinet-office.gov.uk"
+        EMAIL_GROUP_DEV                  = "test-address@digital.cabinet-office.gov.uk"
+        EMAIL_GROUP_FORCE_PUBLISH_ALERTS = "test-address@digital.cabinet-office.gov.uk"
+        FACT_CHECK_SUBJECT_PREFIX        = "dev"
+        FACT_CHECK_USERNAME              = "govuk-fact-check-test@digital.cabinet-office.gov.uk"
+        GOVUK_APP_DOMAIN_EXTERNAL        = var.external_app_domain
+        GOVUK_APP_NAME                   = local.publisher_app_name
+        GOVUK_APP_ROOT                   = "/app"
+        # TODO: how does GOVUK_ASSET_ROOT relate to ASSET_HOST? Is one a function of the other? Are they both really in use? Is GOVUK_ASSET_ROOT always just "https://${ASSET_HOST}"?
+        GOVUK_ASSET_ROOT                = "https://assets.test.publishing.service.gov.uk"
```
richardTowers, line 23: Oh, just noticed this one. Could at least be:

```suggestion
        GOVUK_ASSET_ROOT                = "https://assets.${var.publishing_service_domain}"
```

PR #153, opened 2021-02-03
Refactor signon deployments

bilbof:
This follows PR #150 in refactoring the way we deploy applications.

This should look similar to https://github.com/alphagov/govuk-infrastructure/pull/152.

We no longer register new task definitions in Terraform. Instead, our Big Terraform (💪) `apply` outputs a task definition JSON which is passed to `aws ecs register-task-definition`. This should make the Terraform simpler. See PRs #150 and #152, and the [design doc](https://docs.google.com/document/d/1MzMscVFqzwW0SAZmxBkZv4fHn2JrbHMlEmd1h3lw07I/edit#) (internal) for the motivation behind this refactoring.


richardTowers (approved):
Looks good as far as I can tell. Concourse seems happy, [signon](https://signon-ecs.test.govuk.digital/users/sign_in) still loads in the test env.

PR #154, opened 2021-02-04
Static deployment refactor

fredericfran-gds:
In PR #150, we refactored the way we deploy applications by outputting the
application task definition json when applying the govuk platforming terraform.
This json file is then diffed with previous version and if changes exist, a new
task definition is created for the app using AWS cli.

This PR refactors  (draft-)static app to use this new deployment style.

Similar PRs are #152 and #153 for other apps.

Ref:
1. [trello card](https://trello.com/c/5JT7LE7k/382-update-the-static-application-to-use-the-new-deployment-approach)


terraform/deployments/govuk-publishing-platform/app_static.tf
```diff
@@ -1,51 +1,98 @@
...
   cluster_id                       = aws_ecs_cluster.cluster.id
-  source                           = "../../modules/app"
+  vpc_id                           = local.vpc_id
+  subnets                          = local.private_subnets
   desired_count                    = var.static_desired_count
   extra_security_groups            = [local.govuk_management_access_security_group, aws_security_group.mesh_ecs_service.id]
+  environment_variables = merge(
+    local.static_defaults.environment_variables,
+    {
+      ASSET_HOST          = "https://static-ecs.${local.defaults.environment_variables.GOVUK_APP_DOMAIN_EXTERNAL}",
```
richardTowers, line 37: 🤔  I wonder what this should actually be... The other apps use `local.defaults.asset_host`, which is currently set to `https://frontend.${var.external_app_domain}`.

If this needs to be `https://static-ecs...` then I think:

```suggestion
      ASSET_HOST          = "https://static-ecs.${var.external_app_domain}",
```

would be nicer.

terraform/deployments/govuk-publishing-platform/app_static.tf
```diff
@@ -1,51 +1,98 @@
...
   cluster_id                       = aws_ecs_cluster.cluster.id
-  source                           = "../../modules/app"
+  vpc_id                           = local.vpc_id
+  subnets                          = local.private_subnets
   desired_count                    = var.draft_static_desired_count
   extra_security_groups            = [local.govuk_management_access_security_group, aws_security_group.mesh_ecs_service.id]
+  environment_variables = merge(
+    local.static_defaults.environment_variables,
+    {
+      ASSET_HOST          = "https://draft-static-ecs.${local.defaults.environment_variables.GOVUK_APP_DOMAIN_EXTERNAL}",
```
richardTowers, line 75: As above - `${var.external_app_domain}` is a better reference for the external app domain.

richardTowers (commented):
Looks good - a couple of little comments about ASSET_HOST, but otherwise 🚀

terraform/deployments/govuk-publishing-platform/app_static.tf
fredericfran-gds: I can do that. In `test` environment, ASSET_HOST for static is `https://www.test.publishing.service.gov.uk` and router is supposed to route it to the right app. Since we don't have router working properly, this is a temp fix. I can add a TODO

terraform/deployments/govuk-publishing-platform/app_static.tf
fredericfran-gds: same as above.

richardTowers (commented):
I think you should also delete `terraform/modules/task-definitions/static` (which isn't used now you've removed the deployment).

That's my last comment, then I'll approve!

fredericfran-gds:
Thanks, deleted the directory.

richardTowers (approved):
`+161 -349`

🔥 boilerplate 🔥 

Thanks for this!

PR #155, opened 2021-02-04
Remove redundant router module

richardTowers:
This was removed in 11c7e13, but it was accidentally added back in in 90f328f


fredericfran-gds (approved):
thanks, sorry about that.

PR #158, opened 2021-02-05
Add build pipelines

karlbaker02:
This PR adds image build jobs for a number of apps, based on the existing `smokey` build job. It refactors the `build-image` step of the jobs into a separate `build-image-definition.yml` task config to avoid repeating this across all jobs. The `registry-image` resource is also updated so that it now points to repositories in AWS ECR rather than Docker Hub, requiring user credentials and the assumption of a role using those credentials to access ECR.

Trello: https://trello.com/c/NkxXLeJE

### DIFF
--- a/concourse/pipelines/build-images.yml
+++ b/concourse/pipelines/build-images.yml
@@ -2,36 +2,151 @@
 definitions:
 
 resources:
-  - name: smokey
+  - &git-repo
+    name: smokey-repo
     icon: github
     type: git
-    source:
+    source: &git-repo-source
       uri: git@github.com:alphagov/smokey.git
       branch: main
-      private_key: |
+      private_key: | # pragma: allowlist secret
         ((govukci_private_key))
 
-  - name: govuk-infrastructure
-    icon: github
-    type: git
+  - <<: *git-repo
+    name: content-store-repo
     source:
+      <<: *git-repo-source
+      uri: git@github.com:alphagov/content-store.git
+      branch: master
+
+  - <<: *git-repo
+    name: frontend-repo
+    source:
+      <<: *git-repo-source
+      uri: git@github.com:alphagov/frontend.git
+      branch: master
+
+  - <<: *git-repo
+    name: publisher-repo
+    source:
+      <<: *git-repo-source
+      uri: git@github.com:alphagov/publisher.git
+      branch: master
+
+  - <<: *git-repo
+    name: publishing-api-repo
+    source:
+      <<: *git-repo-source
+      uri: git@github.com:alphagov/publishing-api.git
+      branch: master
+
+  - <<: *git-repo
+    name: router-repo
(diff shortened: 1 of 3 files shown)


bilbof (commented):
Great stuff! Thanks for doing this 🎉 👍 

I’ve put some questions below, but it’d be good to put the container builds design into a doc. I think it'd probably be better to answer the questions below in a doc, since it'll be easier to collaborate :D

Let me know if you would like to pair on writing a design doc for the container builds stuff.

- Is there a reason we've chosen to create a separate deployment module for ECR, rather than creating this within govuk-publishing-platform?
- Since most of the build-images.yml file is duplicated code, and will be quite large once other apps are added; should we remove the duplicate code by creating multiple pipelines from a smaller yaml file, and provide the varying parts as pipeline variables, e.g. `((git_repo_url))`?
- The pipeline pushes to one ECR (currently integration, but will switch to production). It'd be great if we could also push to DockerHub, but we should ensure we don't interfere with the current platform (e.g. perhaps pushing the `latest` tag can impact e2e tests). As an aside, as you mentioned before the ECR repo is private - is it possible to make it public with Terraform?
- It looks like tagging commits with `1.0.*-release` shouldn't affect the current platform. Are we happy using this versioning system? I presume we'll change the deploy pipeline to watch for commits tagged with `*-release` to trigger deploys.
- Currently we build an image when a new commit is merged into the main branch. Have you thought about where we'll run tests against the main branch before we build an image for it?
- Have you considered how we'll build images for each branch (this is what the current platform does), to enable branch deploys?

bilbof (dismissed):
I'm happy for this to go in. I reckon a good next step would be document the design so we can think about how we want to iterate on this.

🎏 Flagging that this has broken the smoke tests (ECS can't pull from the ECR repository, and the task definition doesn't specify the repository); we can fix this later.

karlbaker02:
Thanks @bilbof for the review. I think you've raised some great questions on why we're doing what we're doing, the answers to which I unfortunately don't have on my person.

Likewise, I didn't do any research into how we might run tests against the main branch before building an image for the app in question, nor how we'll build images for each branch; I based my changes off of the existing `smokey` job and didn't consider the above as they weren't within the remit of the card. Happy to do this thinking with you and others though!

bilbof (approved):

bilbof:
Another couple of things:

- it would be good to have a look at the assume role thing again, perhaps we can modify the concourse resource to enable Concourse to assume a role, to enable us not to create a user
- this enables a pipeline in the govuk-ci team to write to our AWS account (production), which I think would be surprising. We have the govuk-ci team separate from the cd-govuk-tools team to enable us to run tests etc in an environment with fewer permissions. Thought building images may be a CI job, it might be worth considering whether to have the build pipeline in the cd-govuk-tools team.

PR #159, opened 2021-02-05
Refactor Smokey deployment

bilbof:
Follows PR #150 in simplifying how we register new task definitions for smokey.

I've moved the task definition and container definition JSON generators out of the app module into their own modules. This enables us to create and register task definitions in a consistent way without needing to create a corresponding ECS Service. This is useful for Smokey and other tasks that aren't going to run as services.

I've modified the `fetch-task-definition` task so we can run tasks either with a task definition ARN or just the task definition family.

I've also changed the smoke tests to get triggered after content-store deploys. I'll switch this back to frontend once frontend uses the new deploy method ([story](https://trello.com/c/FLnLyrnQ/376-update-the-frontend-application-to-use-the-new-deployment-approach)).


karlbaker02 (approved):
All looks good as discussed. 

Making Terraform simpler is 💯  Great job!

PR #160, opened 2021-02-05
Publishing API deployment refactor

bilbof:
This follows PR #150 in changing how we register new task
definitions. Rather than having a separate state file for
each task definition, we output the task definition from
the main govuk-publishing-platform module, and use the
task definition JSON to register the task definition using
the AWS CLI.

This should result in less boilerplate in the Terraform config
and fewer state refreshes when registering a task definition.

[Trello card](https://trello.com/c/oA5wjym7/378-update-the-publishing-api-application-to-use-the-new-deployment-approach)


bilbof:
The Concourse CI build is currently failing due to an issue affecting Concourse.

fredericfran-gds (approved):
looks good.

PR #161, opened 2021-02-05
Upgrade Terraform version: 0.13.3 -> 0.14.6

bilbof:
This is the latest version of Terraform: https://www.terraform.io/downloads.html

Applying Terraform will automatically download the new version.

The biggest change is the addition of the Terraform lock files in each deployment. See the commits for more details.


concourse/pipelines/deploy.yml
```diff
@@ -333,7 +333,7 @@ jobs:
           type: docker-image
           source:
             repository: governmentpaas/terraform
-            tag: terraform-0.13.3
+            tag: terraform-0.14.6
```
richardTowers, line 336: You have discovered my secret shame here:

```
docker run -it --rm governmentpaas/terraform:terraform-0.14.6
Unable to find image 'governmentpaas/terraform:terraform-0.14.6' locally
docker: Error response from daemon: manifest for governmentpaas/terraform:terraform-0.14.6 not found: manifest unknown: manifest unknown.
```

... the terraform-0.13.3 is actually not a proper thing, it's just the name of the branch of [the PR I raised to trick paas's CI into building an image for us](https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/200). 😬

richardTowers: We could pull the same sneaky move again, or we could bite the bullet and build the terraform image ourselves.

karlbaker02: Not sure if it's of any use, but there _is_ an image available which uses Terraform `0.14.6`: https://github.com/alphagov/paas-docker-cloudfoundry-tools/pull/209

bilbof: Oh! 🙈 Thanks for the heads up, I missed that.

@karlbaker02 @richardTowers if it's good with you I'll use the official image - [hashicorp/terraform:0.14.6](https://hub.docker.com/layers/hashicorp/terraform/0.14.6/images/sha256-2c56a1e1d8b10f665e9fa49bc477e66b9afdde12659c90463d4b3f6672663a00?context=explore). I checked our tasks that use `governmentpaas/terraform` and it looks like they will be compatible with the new image.

richardTowers: Yeah, that sounds like a good plan. We already have to trust hashicorp for the terraform binaries, so using their official image seems acceptable to me even in the long term.

bilbof: @richardTowers thanks, I've pushed up a commit with that change.

richardTowers (approved):
Looks good to me!

PR #162, opened 2021-02-08
Refactor frontend

MahmudH:
This PR is for the Frontend application to use the new deployment, which is a follow on from PR [#150](https://github.com/alphagov/govuk-infrastructure/pull/150).

We have refactored the way we deploy applications but outputting the application task definition JSON when applying Terraform. This JSON will be used to check with the previous version and if there is a change then a new task definition is created using the AWS CLI `aws ecs register-task-definition`. 

Trello: https://trello.com/c/FLnLyrnQ


concourse/pipelines/deploy.yml
```diff
@@ -33,15 +33,14 @@ resources:
...
-      paths:
-        - concourse/tasks
-        - terraform/deployments/apps/*frontend
-        # TODO replace these once frontend is refactored to use the new modules
-        - terraform/modules/task-definition
-        - terraform/modules/task-definitions/frontend
+      bucket: ((readonly_private_bucket_name))
+      region_name: eu-west-2
+      versioned_file: frontend.json
+      initial_version: "0"
```
richardTowers, line 43: This should be moved down with the other `*-terraform-outputs` resources (like `static-terraform-outputs`).

It doesn't need to use the `*git-repo` YAML anchor, because it's not using any properties from that bit of YAML (the anchor is defined on line 20 - look for `&git-repo`)

concourse/pipelines/deploy.yml
```diff
@@ -420,25 +430,43 @@ jobs:
...
+         VARIANT: draft
+    - task: update-live-task-definition
+      file: govuk-infrastructure/concourse/tasks/update-task-definition-v2.yml
+      output_mapping:
+       task-definition-arn: live-task-definition-arn
       params:
-        ECS_SERVICE: frontend
-        GOVUK_ENVIRONMENT: test
+       APPLICATION: frontend
+       VARIANT: live
```
richardTowers, line 456: This looks like it needs to be outdented by two more spaces, so it's within the `in_parralel` block that starts on line 431.

terraform/deployments/govuk-publishing-platform/app_frontend.tf
```diff
@@ -1,3 +1,47 @@
...
+
+    environment_variables = merge(
+      local.defaults.environment_variables,
+      {
+        GOVUK_APP_NAME                  = "frontend",
+        GOVUK_CONTENT_SCHEMAS_PATH      = "/govuk-content-schemas",
+        PLEK_SERVICE_PUBLISHING_API_URI = local.defaults.publishing_api_uri
+        PLEK_SERVICE_SIGNON_URI         = local.defaults.signon_uri
+        UNICORN_WORKER_PROCESSES        = 12,
+        ASSET_HOST                      = local.defaults.environment_variables.GOVUK_WEBSITE_ROOT,
```
richardTowers, line 14:
```suggestion
        ASSET_HOST                      = local.defaults.asset_host,
```

terraform/deployments/govuk-publishing-platform/app_frontend.tf
```diff
@@ -1,3 +1,47 @@
...
+    cpu    = 512  # TODO parameterize this
+    memory = 1024 # TODO parameterize this
+
+    environment_variables = merge(
+      local.defaults.environment_variables,
+      {
+        GOVUK_APP_NAME                  = "frontend",
+        GOVUK_CONTENT_SCHEMAS_PATH      = "/govuk-content-schemas",
+        PLEK_SERVICE_PUBLISHING_API_URI = local.defaults.publishing_api_uri
+        PLEK_SERVICE_SIGNON_URI         = local.defaults.signon_uri
```
richardTowers, line 12: These two are defined twice - see lines 17 and 18.

terraform/deployments/govuk-publishing-platform/app_frontend.tf
```diff
@@ -53,12 +97,12 @@ module "draft_frontend" {
     target_group_arn = module.draft_frontend_public_alb.target_group_arn
     container_port   = 80
   }]
-  environment_variables = {} # TODO
-  secrets_from_arns     = {} # TODO
+  environment_variables = local.frontend_defaults.environment_variables
```
richardTowers, line 100: Interestingly, I can't see how the code before handled the difference in environment variables between frontend and draft-frontend. Perhaps they were just getting the same set of variables?

I think we should probably be overriding some of the env vars here though. At least something like:

```terraform
  environment_variables = merge(
    local.frontend_defaults.environment_variables,
    {
        PLEK_SERVICE_CONTENT_STORE_URI  = local.defaults.draft_content_store_uri
        PLEK_SERVICE_STATIC_URI         = local.defaults.draft_static_uri
    }
  )
```

... maybe it doesn't have to be perfect if it wasn't working before this PR though.

richardTowers (commented):
I think you can also delete `terraform/modules/task-definitions/frontend`, as it's not used anywhere anymore 🎉

terraform/deployments/govuk-publishing-platform/app_frontend.tf
```diff
@@ -53,14 +95,21 @@ module "draft_frontend" {
...
-  cpu                   = 512
-  memory                = 1024
-  task_role_arn         = aws_iam_role.task.arn
-  execution_role_arn    = aws_iam_role.execution.arn
+  environment_variables = merge(
+    local.frontend_defaults.environment_variables,
+    {
+      DRAFT_STATIC_URI               = local.defaults.draft_static_uri
+      PLEK_SERVICE_CONTENT_STORE_URI = local.defaults.draft_content_store_uri,
+      PLEK_SERVICE_STATIC_URI        = local.defaults.static_uri
```
richardTowers, line 103:
```suggestion
      PLEK_SERVICE_STATIC_URI        = local.defaults.draft_static_uri
```

terraform/deployments/govuk-publishing-platform/app_frontend.tf
```diff
@@ -53,14 +95,21 @@ module "draft_frontend" {
...
-  log_group             = local.log_group
-  aws_region            = data.aws_region.current.name
-  cpu                   = 512
-  memory                = 1024
-  task_role_arn         = aws_iam_role.task.arn
-  execution_role_arn    = aws_iam_role.execution.arn
+  environment_variables = merge(
+    local.frontend_defaults.environment_variables,
+    {
+      DRAFT_STATIC_URI               = local.defaults.draft_static_uri
```
richardTowers, line 101: I think this line can be removed - we want to set the environment variable that tells Plek to use the Draft version of static when it wants to talk to static. No need to set this variable, which the app will just ignore.

richardTowers: I pushed a fix for this in 4af5c66.

terraform/deployments/govuk-publishing-platform/app_frontend.tf
richardTowers: I pushed a fix for this in 4af5c66.

richardTowers (approved):
Looks good to me, but could use another review because I've just pushed one little commit up 😅

karlbaker02 (approved):
Looks good 💯

PR #163, opened 2021-02-09
Bump Envoy image

bilbof:
This is the latest available image: https://docs.aws.amazon.com/app-mesh/latest/userguide/envoy.html.


richardTowers (approved):
Looks good! Will be nice when this is just defined in one place :)

PR #164, opened 2021-02-17
Update tools/run-task.sh to work with new deployments

richardTowers:
We've changed the way we deploy our apps now, so the individual
terraform deploys don't exist.

Instead, we can get the network config from govuk-publishing-platform's
terraform outputs and the task defintion ARN from aws ecs.

Seems to work okay.

I've also made it guess the URL to look at for logs.

I've only tested this for publisher-web. It probably won't work for things that are draft / live instead of web / worker. I was thinking fix that later.

### DIFF
--- a/tools/run-task.sh
+++ b/tools/run-task.sh
@@ -11,15 +11,18 @@ cluster="task_runner"
 govuk_env="test"
 
 function show_help {
-  echo "Usage: cd govuk-infrastructure && gds aws govuk-tools-internal-admin ./tools/run-rake-task -a frontend rake db:migrate"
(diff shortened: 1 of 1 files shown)


tools/run-task.sh
```diff
@@ -11,16 +11,19 @@ cluster="task_runner"
 govuk_env="test"
 
 function show_help {
-  echo "Usage: cd govuk-infrastructure && gds aws govuk-tools-internal-admin ./tools/run-rake-task -a frontend rake db:migrate"
+  echo "Usage: cd govuk-infrastructure && gds aws govuk-test-poweruser ./tools/run-rake-task -a frontend -v live rake db:migrate"
```
richardTowers, line 14: Actually this needs to be `-admin` because there's an IAM policy that needs attaching.

bilbof (approved):
Thank you!

PR #165, opened 2021-02-17
Tail the logs instead of trying to build a link

richardTowers:
AWS make this _super annoying_. `aws logs tail` only supports getting
all of the logs from an entire group (in our case the whole govuk
cluster), so we have to use `grep` on the client side to filter out the
logs we want. Why do you have to make this so bad AWS? Why?

### DIFF
--- a/tools/run-task.sh
+++ b/tools/run-task.sh
@@ -5,6 +5,9 @@
 
 set -eu
 
+trap "exit" INT TERM
+trap "kill -- -$$" EXIT
+
 OPTIND=1 # Reset in case getopts has been used previously in the shell.
 
 cluster="task_runner"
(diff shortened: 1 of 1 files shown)


tools/run-task.sh
```diff
@@ -5,6 +5,9 @@
 
 set -eu
 
+trap "exit" INT TERM
+trap "kill -- -$$" EXIT
```
bilbof, line 9: I needed [StackOverflow](https://superuser.com/a/562804) to understand this.

bilbof (approved):
Thanks 👍

tools/run-task.sh
richardTowers: I needed stackoverflow to work out how to do it 🙈 

Shame there's not a more obvious way of doing it, but y'know... shell-scripts.

At some point this thing either gets wrapped up as a gds-cli command or replaced with nicer AWS tools, so it's only temporary.

PR #166, opened 2021-02-18
Add Virtual Service backends to App Mesh virtual nodes

bilbof:
This fixes a misconfiguration of our AWS App Mesh [Virtual Nodes][].

**What is App Mesh and Envoy?**

[App Mesh][] is the control plane for our Envoy proxies. Envoy, a distributed (or client-side) load balancer, pulls configuration from App Mesh that tells Envoy how to handle outgoing requests from an app.

Every app we run in ECS has an [Envoy proxy][] in front of it. Significantly, Envoy doesn't just proxy incoming requests to an app, it also _intercepts outgoing requests_ via an iptables rule. Envoy then can redirect the request to a virtual service (an abstraction of a real service) which will route requests to a virtual node (such as an ECS Service).

By intercepting requests Envoy can load balance requests to hosts in upstream clusters (and more!), and log and produce metrics in various formats. This is handy for both observability and reliability. We're also thinking about using Envoy for mTLS.

See [Life of a Request](https://www.envoyproxy.io/docs/envoy/latest/intro/life_of_a_request) through Envoy for more details of what happens when Envoy intercepts a request.

## What does this fix?

We hadn't set any virtual service backends on our virtual nodes. This meant that all outgoing requests from apps in the service mesh were considered to be leaving the service mesh.

How did that happen?

When Envoy intercepts an outgoing request from an app, it checks whether the app's virtual node has a 'backend' virtual service that corresponds to the request it has intercepted. These 'backends' are also called "upstream clusters" by Envoy, but we'll use the App Mesh terminology. If it does have a corresponding backend service, Envoy will load balance the request to the virtual service (ignoring whatever IP address the app had already acquired for the service).

If Envoy does not have a corresponding backend, then what happens next depends on the App Mesh [egress filter][], which can be `ALLOW_ALL` or `DROP_ALL`. If `DROP_ALL` (default) is used, then the request to an unknown virtual service will be dropped. DROP_ALL allows egress only from virtual nodes to other defined resources in the service mesh (and any traffic to *.amazonaws.com for AWS API calls). If `ALLOW_ALL` is used, the request will be proxied to the endpoint regardless, as this allows egress to any endpoint inside or outside of the service mesh.

When we enabled the egress filter `DROP_ALL` on our App Mesh mesh, this caused all requests made by apps in the mesh to be dropped.

The only reason service to service requests in the mesh succeeded is because of the `ALLOW_ALL` egress filter. This troubleshooting section "[Connectivity succeeds to service not listed as a virtual service backend for a virtual node](https://docs.aws.amazon.com/app-mesh/latest/userguide/troubleshoot-connectivity.html#ts-connectivity-not-virtual-service)" helped.

## What is the fix?

This happened because our configuration currently doesn't define any backing virtual services for our virtual nodes (apart from permitting the virtual services to send requests to themselves).

This changes the service mesh node module such that one can provide a list of backing virtual services for a virtual node, so that these virtual services can be treated as an Upstream Cluster by Envoy.

## More details

I'd recommend reading the AWS App Mesh documentation, but particularly the Envoy documentation.

In the course of discovering this bug with @richardTowers (props to Richard for actually spotting this!), we found that the [admin panel][] and endpoints for Envoy are quite useful. Use `/config_dump` and Envoy's port 9901. Prometheus metrics from Envoy will also help in future, as will `aws ecs exec`.

## Comment on the implementation

Having the service-mesh node implemented in the app module is a convenience but we may find we want to pull this module up and have the services/nodes defined in the deployment module. We should consider this when we start adding external services such as DocumentDB, EC2-hosted databases, Sentry, etc. as virtual services.

For now, we've left the egress filter as `ALLOW_ALL` rather than `DROP_ALL` since it'd be a lot of work and a confusing debugging experience to block all outgoing requests. We think we can add virtual services gradually - this would be worth doing since we'll get some load balancing properties and observability into outgoing requests to external services.

Richard also had an interesting idea about using a black hole IP address for the A record for services, rather than having CloudMap create DNS records. All we need is for the applications to resolve the IP address, and then for Envoy to re-resolve the domain to an IP from its cache of IPs. The IP address the app gets is ignored by Envoy if the service is recognised, so this could work - needs experimentation.

[App Mesh]: https://docs.aws.amazon.com/app-mesh/index.html
[Virtual Nodes]: https://docs.aws.amazon.com/app-mesh/latest/userguide/virtual_services.html
[Envoy proxy]: https://www.envoyproxy.io/docs/envoy/latest/
[egress filter]: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/appmesh_mesh#egress-filter
[admin panel]: https://www.envoyproxy.io/docs/envoy/latest/operations/admin

### DIFF
--- a/terraform/deployments/govuk-publishing-platform/app_router.tf
+++ b/terraform/deployments/govuk-publishing-platform/app_router.tf
@@ -28,6 +28,7 @@ module "router" {
   source = "../../modules/app"
 
   service_name                     = "router"
+  backend_virtual_service_names    = module.frontend.virtual_service_names
   mesh_name                        = aws_appmesh_mesh.govuk.id
   service_discovery_namespace_id   = aws_service_discovery_private_dns_namespace.govuk_publishing_platform.id
   service_discovery_namespace_name = aws_service_discovery_private_dns_namespace.govuk_publishing_platform.name
@@ -60,6 +61,7 @@ module "draft_router" {
   source = "../../modules/app"
 
   service_name                     = "draft-router"
+  backend_virtual_service_names    = module.draft_frontend.virtual_service_names
   mesh_name                        = aws_appmesh_mesh.govuk.id
   service_discovery_namespace_id   = aws_service_discovery_private_dns_namespace.govuk_publishing_platform.id
   service_discovery_namespace_name = aws_service_discovery_private_dns_namespace.govuk_publishing_platform.name
--- a/terraform/deployments/govuk-publishing-platform/app_router_api.tf
+++ b/terraform/deployments/govuk-publishing-platform/app_router_api.tf
@@ -3,6 +3,11 @@ locals {
     cpu    = 512  # TODO parameterize this
     memory = 1024 # TODO parameterize this
 
+    backend_services = flatten([
+      local.defaults.virtual_service_backends,
+      module.signon.virtual_service_names,
+    ])
+
     environment_variables = merge(
       local.defaults.environment_variables,
       {
@@ -27,6 +32,7 @@ module "router_api" {
   source = "../../modules/app"
 
   service_name                     = "router-api"
+  backend_virtual_service_names    = local.router_api_defaults.backend_services
   mesh_name                        = aws_appmesh_mesh.govuk.id
   service_discovery_namespace_id   = aws_service_discovery_private_dns_namespace.govuk_publishing_platform.id
   service_discovery_namespace_name = aws_service_discovery_private_dns_namespace.govuk_publishing_platform.name
@@ -62,6 +68,7 @@ module "draft_router_api" {
   source = "../../modules/app"
 
   service_name                     = "draft-router-api"
+  backend_virtual_service_names    = local.router_api_defaults.backend_services
   mesh_name                        = aws_appmesh_mesh.govuk.id
   service_discovery_namespace_id   = aws_service_discovery_private_dns_namespace.govuk_publishing_platform.id
   service_discovery_namespace_name = aws_service_discovery_private_dns_namespace.govuk_publishing_platform.name
--- a/terraform/deployments/govuk-publishing-platform/app_static.tf
+++ b/terraform/deployments/govuk-publishing-platform/app_static.tf
@@ -23,6 +23,7 @@ module "static" {
   source = "../../modules/app"
 
   service_name                     = "static"
+  backend_virtual_service_names    = [] # Static doesn't use any other services
   mesh_name                        = aws_appmesh_mesh.govuk.id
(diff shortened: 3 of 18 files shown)


terraform/deployments/govuk-publishing-platform/app_router.tf
```diff
@@ -28,6 +28,7 @@ module "router" {
   source = "../../modules/app"
 
   service_name                     = "router"
+  backend_virtual_service_names    = [] # TODO what services does Router use?
```
richardTowers, line 31: I'm pretty sure router doesn't use any services (other than its DB, which we're not modelling as a mesh service yet).

```suggestion
  backend_virtual_service_names    = [] # Router doesn't use any other services
```

terraform/deployments/govuk-publishing-platform/app_router_api.tf
```diff
@@ -27,6 +32,7 @@ module "router_api" {
   source = "../../modules/app"
 
   service_name                     = "router-api"
+  backend_virtual_service_names    = local.router_api_defaults.backend_services # TODO what services does Router API use?
```
richardTowers, line 35: I guess it should probably have Router in its services (although I'm not sure we've really hooked that up properly). I don't think it talks to any other services, other than maybe signon which you've already got.

terraform/deployments/govuk-publishing-platform/app_static.tf
```diff
@@ -23,6 +23,7 @@ module "static" {
   source = "../../modules/app"
 
   service_name                     = "static"
+  backend_virtual_service_names    = [] # TODO what services does Static use?
```
richardTowers, line 26: I don't think it uses any, so you could probably remove this TODO too.

```suggestion
  backend_virtual_service_names    = [] # Static doesn't use any other services
```

richardTowers (dismissed):
Looks good to me, thanks for this investigation.

A couple of suggestions to resolve the TODO comments (I think router and static don't make any requests to other services).

terraform/deployments/govuk-publishing-platform/app_router.tf
bilbof: I've added frontend for now, since other than the DB, I think Router sends requests to many other apps registered as backends.

bilbof:
@richardTowers thanks! I've amended the last commit. Let me know if I've addressed your comments :)

terraform/deployments/govuk-publishing-platform/app_router.tf
richardTowers: Oh yeah 🤦🏻

richardTowers (approved):

PR #169, opened 2021-02-25
fix test env

fredericfran-gds:
fix test env

Includes:
1. fixes to frontend secrets and size of fargate container
2. fixes to deploy concourse pipeline: frontend and publishing-api
3. fixes to healthcheck endpoint of Static which has changed


richardTowers (approved):
Thanks for this Fred!

Sorry, I think some of these slipped through a review I did. I should have done some testing before merging.

PR #170, opened 2021-02-25
introduce www/draft-origin for assets of apps 

fredericfran-gds:
This PR is the continuation of the work to add Fastly CDN in front of
the Test environment for basic authentication.

In order to route all public user traffic (except traffic usually going to
assets.*.publishing.service.gov.uk) via the CDN so that they can be
authenticated, we need to introduce the origin component of GOV.UK.

The origin component of GOV.UK acts as a `single` entry point to the
GOV.UK environment and fans out the requests to different other apps of
GOV.UK. The origin component of GOV.UK is implemented in the EC2 platform
using nginx configs and GOV.UK router app

This PR implements the nginx routing using a AWS Application Load
Balancer in the new ECS environment with the use of path prefix
matching to redirect requests to the correct frontend GOV.UK apps.

In effect, if HTTP request arrives with:
1. /assets/frontend/* this is directed to the Frontend app
2. /assets/static/* this is directed to the Static app
3. otherwise, all traffic goes to the Frontend app
   (when Router is properly implemented, this should go to the Router
app)

Ref:
1. [trello card](https://trello.com/c/TjJOKSKp/404-basic-auth-for-integration-environment)

### DIFF
--- a/terraform/deployments/apps/common.tf
+++ b/terraform/deployments/apps/common.tf
@@ -11,10 +11,10 @@ locals {
   sentry_environment               = "${var.govuk_environment}-ecs"
   static_url                       = "http://static.${var.mesh_domain}"
   draft_static_url                 = "http://draft-static.${var.mesh_domain}"
-  statsd_host                      = "statsd.${var.mesh_domain}"                   # TODO: Put Statsd in App Mesh
-  website_root                     = "https://frontend.${var.external_app_domain}" # TODO: Change back to www once router is up
-  router_urls                      = "router.${var.mesh_domain}:3055"              # TODO(https://trello.com/c/gmzObCBG/95): router-api expects a list of individual instances, so this won't work as-is.
-  draft_router_urls                = "draft-router.${var.mesh_domain}:3055"        # TODO(https://trello.com/c/gmzObCBG/95): router-api expects a list of individual instances, so this won't work as-is.
+  statsd_host                      = "statsd.${var.mesh_domain}" # TODO: Put Statsd in App Mesh
+  website_root                     = "https://${module.www_origin.fqdn}"
+  router_urls                      = "router.${var.mesh_domain}:3055"       # TODO(https://trello.com/c/gmzObCBG/95): router-api expects a list of individual instances, so this won't work as-is.
+  draft_router_urls                = "draft-router.${var.mesh_domain}:3055" # TODO(https://trello.com/c/gmzObCBG/95): router-api expects a list of individual instances, so this won't work as-is.
 }
 
 data "aws_iam_role" "execution" {
--- a/terraform/modules/origin/main.tf
+++ b/terraform/modules/origin/main.tf
@@ -1,0 +1,115 @@
+locals {
+  mode = var.live ? "www" : "draft"
+}
+
+
+# TODO: use a single, ACM-managed cert with both domains on. There is already
+# such a cert in integration/staging/prod (but it needs defining in Terraform).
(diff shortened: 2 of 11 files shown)


terraform/deployments/apps/common.tf
```diff
@@ -11,10 +11,10 @@ locals {
...
-  website_root                     = "https://frontend.${var.external_app_domain}" # TODO: Change back to www once router is up
-  router_urls                      = "router.${var.mesh_domain}:3055"              # TODO(https://trello.com/c/gmzObCBG/95): router-api expects a list of individual instances, so this won't work as-is.
-  draft_router_urls                = "draft-router.${var.mesh_domain}:3055"        # TODO(https://trello.com/c/gmzObCBG/95): router-api expects a list of individual instances, so this won't work as-is.
+  statsd_host                      = "statsd.${var.mesh_domain}" # TODO: Put Statsd in App Mesh
+  website_root                     = "https://${module.www_origin.fqdn}"
+  router_urls                      = "router.${var.mesh_domain}:3055"       # TODO(https://trello.com/c/gmzObCBG/95): router-api expects a list of individual instances, so this won't work as-is.
+  draft_router_urls                = "draft-router.${var.mesh_domain}:3055" # TODO(https://trello.com/c/gmzObCBG/95): router-api expects a list of individual instances, so this won't work as-is.
```
bilbof, line 17: I think we could omit the ports here since Router will run on port 80.

terraform/modules/origin/main.tf
```diff
@@ -0,0 +1,115 @@
...
+  }
+}
+
+data "aws_route53_zone" "public" {
+  name = var.external_app_domain
+}
+
+resource "aws_route53_record" "origin_alb" {
+  zone_id = data.aws_route53_zone.public.zone_id
+  name    = "${local.mode}-origin-ecs"
```
bilbof, line 107:
```suggestion
  name    = "${local.mode}-origin-${var.workspace_suffix}"
```

terraform/modules/origin/security_groups.tf
```diff
@@ -0,0 +1,36 @@
...
+  for_each                 = var.apps_security_config_list
+  description              = "${each.key} receives requests from the ${local.mode}-origin ALB over HTTP"
+  type                     = "ingress"
+  from_port                = each.value.target_port
+  to_port                  = each.value.target_port
+  protocol                 = "tcp"
+  security_group_id        = each.value.security_group_id
+  source_security_group_id = aws_security_group.origin_alb.id
+}
+
```
bilbof, line 17: 🤔 can we move these rules to `govuk-publishing-platform/security_group_rules.tf`? We output the security group ID already, so I guess it would be easy to shift the rules to the central place.

bilbof (approved):
Excellent work as usual. Couple of minor changes requested.

terraform/modules/origin/main.tf
bilbof: Oh, actually, please change this to `www-origin.ecs.test.govuk.digital`, as in the [DNS design doc](https://docs.google.com/document/d/1QpFPXRSqzWYefl7a9MmAH_9QuHAcwmMxgvftRwZbPCc/edit#)

terraform/modules/origin/security_groups.tf
fredericfran-gds: we need to ask @richardTowers , it is a copy of the public-load-balancer module?

fredericfran-gds:
we will create cards to fix other issues.

PR #171, opened 2021-03-01
Replace 'dummy' bootstrap task definition

bilbof:
During bootstrapping we need to create a task definition for a service to be brought up.

Previously we just used a dummy task definition that would be replaced by a real task once the apps are deployed.

This bootstrap task definition was confusing - here it is replaced with another bootstrap task def. The difference is that this one should match the definition created later when we deploy the app. This will enable us to bootstrap a new environment in one step.


terraform/deployments/govuk-publishing-platform/.terraform.lock.hcl
```diff
@@ -1,6 +1,27 @@
 # This file is maintained automatically by "terraform init".
```
sengi, line 1: Accidental commit?

terraform/modules/app/task_definition.tf
```diff
@@ -81,6 +82,25 @@ module "task_definition" {
   task_role_arn = var.task_role_arn
 }
 
+resource "aws_ecs_task_definition" "bootstrap" {
```
richardTowers, line 85: Should we have a `lifecycle { ignore_changes = all }` on this resource?

I don't think it's a huge issue it not being there, but something like this will probably happen:

* First terraform run - task definition created, and service told to use that task definition
* aws cli used to change the service's task definition
* Future terraform runs may update the task definition, but it won't be re-attached to the service because there's a lifecycle rule `ignore_changes = [task_definition]`

If we put an `ignore_changes = all` on this then we'd avoid needlessly updating the bootstrap task definition.

richardTowers: https://www.terraform.io/docs/language/meta-arguments/lifecycle.html

> Instead of a list, the special keyword all may be used to instruct Terraform to ignore all attributes, which means that Terraform can create and destroy the remote object but will never propose updates to it.

terraform/deployments/govuk-publishing-platform/.terraform.lock.hcl
```diff
@@ -1,6 +1,27 @@
...
+    "zh:2a2662e4c51aa2d215e474aea79d678660703ae8a003c40a4acb213b1d7e2206",
+    "zh:2ff07718842afe1f933c46cb4444695108259fcbf56c4385f81b3e5a81a773c6",
+    "zh:43a8e1808e5832d64bdf195124524e6d0ef09f3e4299f1353ecafb87ae975f5a",
+    "zh:55d3ca56b3f6da17c0b459d4d00e87a6be0b70adb292320f2e2c2e5620a07bff",
+    "zh:6238fa89b4ad71f7cc5608eb9c97e918c679e320dd3ebfec66243cb8dca5fb35",
+    "zh:85690b1c4ae5ce8530bd4bcea7cde66f93512a5c5bb306b9aac0811f62447a5e",
+    "zh:874b138dad72287e66f1738b74eac8a63ce68488a1c87a27adb50db3295cc1fd",
+    "zh:f687f359caf38720cc833a1e96392cd8e2b68094135255c4572d21781b59b9b6",
+  ]
+}
```
richardTowers, line 23: 🤔  should this be in a separate commit?

terraform/modules/app/task_definition.tf
```diff
@@ -32,6 +32,7 @@ locals {
 
 module "app_container_definition" {
   source                = "../../modules/container-definition"
+  image                 = "govuk/${var.image_name}:deployed-to-production"
```
sengi, line 35: I guess we can't have "deployed-to-${env}" because the environment name isn't in scope. Fair enough :)

terraform/modules/app/task_definition.tf
bilbof: @richardTowers sure, thanks for the suggestion!

terraform/deployments/govuk-publishing-platform/.terraform.lock.hcl
bilbof: Terraform advices that we include [dependency lock files](https://www.terraform.io/docs/language/dependency-lock.html) (a recent addition to Terraform) in version control partly for code review:

> You should include this file in your version control repository so that you can discuss potential changes to your external dependencies via code review, just as you would discuss potential changes to your configuration itself.

terraform/modules/app/task_definition.tf
bilbof: Plus, the `deployed-to-env` tag is only generated when an image is deployed to a specific environment - so this wouldn't work for new environments. Another reason is that the `deployed-to-production` tag is more likely to point to a working release (since folks frequently test things out on integration).

bilbof:
@sengi @richardTowers thanks for your comments. I've made the requested changes :)

terraform/deployments/govuk-publishing-platform/.terraform.lock.hcl
sengi: Yuck. Fair enough, let's go with their recommendation then :)

sengi (approved):

PR #172, opened 2021-03-03
Add cron resource and rebuild infrastructure tasks

MahmudH:
This PR is to ensure that our infrastructure can be brought up and torn down easily.

The steps taken are:
- Added a cron resource
- Added a job for terraform destroy to run Mon-Fri 11pm
- Added a task to run terraform apply Mon-Fri 6am

`terraform destroy` and `terraform apply` will execute when the cron job specifies. 

Trello card: https://trello.com/c/wP84gqKR/290-tear-down-and-rebuild-infrastructure-in-test-env-daily

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -316,7 +316,11 @@ jobs:
 
               cd govuk-infrastructure/terraform/deployments/govuk-publishing-platform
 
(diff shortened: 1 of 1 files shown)


bilbof (approved):
Nice work! This looks great. I'm sure this will help us find problems in our config.

We just need to remember to not enable it in production 🙉

PR #173, opened 2021-03-10
Remove remaining app deployment config

bilbof:
We've changed the way we deploy apps to ECS so we can remove these last bits of config.

The `common.tfvars` file is no longer needed - all deployments use `infrastructure.tfvars` anyway.

The v2 files were only for our refactoring process and can replace the v1 files now.

The `terraform/deployments/apps` directory can be deleted since it is no longer used, as with the other changes.


richardTowers (approved):
Nice one - looks much better.

PR #175, opened 2021-03-11
Remove app secrets from Envoy containers

bilbof:
App secrets were erroneously provided to Envoy instances
as environment variables. Envoy requires no secrets provided
via env vars.


fredericfran-gds (approved):
looks good

PR #176, opened 2021-03-11
Workspace

rtrinque:
This modify the codebase to make it "workspace aware".
It make the name attribute of resources generated through the terraform.workspace variable to avoid interference between workspaces.
It adds management of DNS zones and certificates as those are needed for new workspaces and were previously not created through terraform.
It adds new input variables for modules that need DNS zone/certificate (like public-load-balancer)

### DIFF
--- a/terraform/modules/origin/main.tf
+++ b/terraform/modules/origin/main.tf
@@ -103,7 +103,7 @@ data "aws_route53_zone" "public" {
 }
 
 resource "aws_route53_record" "origin_alb" {
-  zone_id = data.aws_route53_zone.public.zone_id
+  zone_id = var.public_zone_id
   name    = "${local.mode}-origin-ecs"
   type    = "A"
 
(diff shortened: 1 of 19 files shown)


terraform/modules/origin/main.tf
```diff
@@ -103,7 +103,7 @@ data "aws_route53_zone" "public" {
 }
 
 resource "aws_route53_record" "origin_alb" {
-  zone_id = data.aws_route53_zone.public.zone_id
+  zone_id = var.public_zone_id
```
sengi, line 106: We should probably be more consistent about when we pass zone IDs vs domain names. But that's something to think about later.

sengi (approved):
Thanks for seeing this grungy refactor through to the end!

All good - just a couple of trivial suggestions and a couple of thoughts/questions for later. Feel free to merge and fix forward.

sengi (approved):

PR #178, opened 2021-03-15
Add certificate workspace awareness to origin module

rtrinque:
Add certificate workspace awareness to origin module

### DIFF
--- a/terraform/modules/origin/variables.tf
+++ b/terraform/modules/origin/variables.tf
@@ -3,6 +3,10 @@ variable "external_app_domain" {
   description = "e.g. test.govuk.digital. Domain in which to create DNS records for the app's Internet-facing load balancer."
 }
 
+variable "certificate" {
+  type = string
+}
+
 variable "live" {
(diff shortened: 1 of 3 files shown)


terraform/modules/origin/variables.tf
```diff
@@ -3,6 +3,10 @@ variable "external_app_domain" {
   description = "e.g. test.govuk.digital. Domain in which to create DNS records for the app's Internet-facing load balancer."
 }
 
+variable "certificate" {
```
richardTowers, line 6: Maybe this should be called `certificate_arn` to make it clear it's the ARN of the cert, not the cert itself. Or you could add a description? Or you could do both?

rtrinque: Yep fair, I'll raise another PR to fix this as it impacts other modules

smford: Roch and I have discussed and I'm with Roch in having this changed in a subsequent PR.

smford (approved):

richardTowers (approved):

PR #180, opened 2021-03-15
Rails assets uploads

fredericfran-gds:
Proposal to add rails assets to s3 by using the precompiled assets in the docker
image. For now, we use only static and frontend as test cases.

The deployment process for static and frontend app is as follows:
1. a new resource monitors the app image container with tag:
   `release`, when the sha256/digest of the image changes
   because a new container image has uploaded, a new app deploy happens.

   Later, we can filter via semantic versioning later when we integrate fully
   the AWS ECR.

2. The app image already contains the precompiles assets and are
   uploaded to S3, we uploaded all the public assets of an app to
   <bucket_name>/assets/<app_name>. While there will be only assets
   directory in the bucket, this is easier since no S3 rerouting rule
   needs to be added.

3. The task definition uses the sha256/digest rather than the tag.
   This can be changed later when we moved to semantic versioning.

Ref:
1. [trello card](https://trello.com/c/NRLwGdBJ/172-static-assets-in-ecs)

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -26,13 +26,6 @@ resources:
     type: git
 
   - <<: *git-repo
-    name: frontend
-    source:
-      branch: master
-      uri: https://github.com/alphagov/frontend
-      tag_filter: release_*
-
-  - <<: *git-repo
     name: publisher
     source:
       branch: master
@@ -105,13 +98,6 @@ resources:
       region_name: eu-west-2
       versioned_file: smokey.json
       initial_version: "0"
-
-  - <<: *git-repo
-    name: static
-    source:
-      branch: master
-      uri: https://github.com/alphagov/static
-      tag_filter: release_*
 
   - name: deploy-slack-channel
     type: slack-notification
@@ -195,6 +181,25 @@ resources:
       region_name: eu-west-2
       versioned_file: router.json
       initial_version: "0"
+
+
+  - name: frontend-image
+    type: registry-image
+    icon: docker
+    source:
+      repository: govuk/frontend
+      tag: release
+      username: ((docker_hub_username))
+      password: ((docker_hub_authtoken))
+
+  - name: static-image
+    type: registry-image
+    icon: docker
+    source:
+      repository: govuk/static
+      tag: release
+      username: ((docker_hub_username))
+      password: ((docker_hub_authtoken))
 
 groups:
   - name: all
@@ -433,12 +438,20 @@ jobs:
         passed:
         - run-terraform
         trigger: true
-      - get: release
-        resource: frontend
+      - get: frontend-image
         trigger: true
     - in_parallel:
+      - task: upload-rails-assets
+        file: govuk-infrastructure/concourse/tasks/upload-rails-assets.yml
+        input_mapping:
+          app-image: frontend-image
+        params:
+          IMAGE_ASSETS_PATH: app-image/rootfs/app/public/assets/frontend
+          S3_BUCKET_PATH: s3://govuk-test-ecs-rails-assets/assets/frontend/
       - task: update-draft-task-definition
         file: govuk-infrastructure/concourse/tasks/update-task-definition.yml
+        input_mapping:
+          app-image: frontend-image
         output_mapping:
          task-definition-arn: draft-task-definition-arn
         params:
@@ -446,6 +459,8 @@ jobs:
          VARIANT: draft
       - task: update-live-task-definition
         file: govuk-infrastructure/concourse/tasks/update-task-definition.yml
+        input_mapping:
+          app-image: frontend-image
         output_mapping:
          task-definition-arn: live-task-definition-arn
         params:
@@ -624,6 +639,7 @@ jobs:
       - get: release
         resource: content-store
         trigger: true
+
     - in_parallel:
       - task: update-draft-task-definition
         file: govuk-infrastructure/concourse/tasks/update-task-definition.yml
@@ -818,12 +834,20 @@ jobs:
         passed:
         - run-terraform
         trigger: true
-      - get: release
-        resource: static
+      - get: static-image
         trigger: true
     - in_parallel:
+      - task: upload-rails-assets
+        file: govuk-infrastructure/concourse/tasks/upload-rails-assets.yml
+        input_mapping:
+          app-image: static-image
+        params:
+          IMAGE_ASSETS_PATH: app-image/rootfs/app/public/assets/static
+          S3_BUCKET_PATH: s3://govuk-test-ecs-rails-assets/assets/static/
       - task: update-draft-task-definition
         file: govuk-infrastructure/concourse/tasks/update-task-definition.yml
+        input_mapping:
+          app-image: static-image
         output_mapping:
           task-definition-arn: draft-task-definition-arn
         params:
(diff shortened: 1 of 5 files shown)


richardTowers:
Nice. 

So you've needed to model docker images as concourse resources, since we need the contents of the image now (previously we just worked out the tag based on the git repo, and then set that in ECS). Rather than running the image, concourse just lets you pull its root filesystem, from where you can just sync the assets to S3 :)

You haven't gone all the way to using images from ECR, because that's an unrelated (and probably hairy) yak shave. I think using the `deployed-to-production` tag is a bit of a change though - at the moment we use the git hash matching the `release_*` tag filter. Those tags get created before the deploy to integration (I think). It would be a shame for us to be downstream of production, even temporarily. It looks like there are `release_3719` tags in docker hub too, and a `release` tag which is presumably the latest release - could we use those instead perhaps?

fredericfran-gds:
> Nice.
> 
> So you've needed to model docker images as concourse resources, since we need the contents of the image now (previously we just worked out the tag based on the git repo, and then set that in ECS). Rather than running the image, concourse just lets you pull its root filesystem, from where you can just sync the assets to S3 :)
> 
> You haven't gone all the way to using images from ECR, because that's an unrelated (and probably hairy) yak shave. I think using the `deployed-to-production` tag is a bit of a change though - at the moment we use the git hash matching the `release_*` tag filter. Those tags get created before the deploy to integration (I think). It would be a shame for us to be downstream of production, even temporarily. It looks like there are `release_3719` tags in docker hub too, and a `release` tag which is presumably the latest release - could we use those instead perhaps?

Thanks. The reasons why I choose the `deployed-to-production` tag are:
1. Incomplete tests before a release tag mean we are not sure about the quality of the images. Fewer things to troubleshoot if we go with the Production images for now.
2. The concourse resource does not understand non semver versioning of the images that we use in docker hub, i.e. release_<number>. This is solved with the ECR images though.

concourse/pipelines/deploy.yml
```diff
@@ -196,6 +182,25 @@ resources:
       versioned_file: router.json
       initial_version: "0"
 
+
+  - name: frontend-image
+    type: registry-image
+    icon: docker
+    source:
+      repository: govuk/frontend
+      tag: deployed-to-production
```
richardTowers, line 191: I think I'd prefer `release` to `deployed-to-production` here. I hear what you're saying about incomplete testing, but I'm pretty sure the unit tests will have run before an image with the `release` tag gets pushed. I'd like us to keep thinking about this as the left-hand-side of the pipeline and having things depend on a manual deploy to the old production environment just breaks that for me.

concourse/pipelines/deploy.yml
```diff
@@ -196,6 +182,25 @@ resources:
...
+      username: ((docker_hub_username))
+      password: ((docker_hub_authtoken))
+
+  - name: static-image
+    type: registry-image
+    icon: docker
+    source:
+      repository: docker.io/govuk/static
```
richardTowers, line 199: Why has this one got `docker.io` as a prefix, when frontend doesn't?

concourse/pipelines/deploy.yml
```diff
@@ -818,19 +834,29 @@ jobs:
...
-        resource: static
+      - get: static-image
         trigger: true
+    - task: upload-rails-assets
+      file: govuk-infrastructure/concourse/tasks/upload-rails-assets.yml
+      input_mapping:
+        app-image: static-image
+      params:
+        IMAGE_ASSETS_PATH: image/rootfs/app/public/assets/static
+        S3_BUCKET_PATH: s3://govuk-test-ecs-rails-assets/assets/static/
```
richardTowers, line 845: I think this could happen in parallel with the task definition update (might be a small performance win).

concourse/tasks/update-task-definition.yml
```diff
@@ -10,7 +10,11 @@ inputs:
   - name: govuk-infrastructure
     path: src
   - name: release
+    optional: true
   - name: app-terraform-outputs
+  - name: app-image
+    path: image
```
richardTowers, line 16: Why customise the path? Wouldn't it be simpler to let it default to placing the file in `app-image`, and then using `[ -f "app-image/digest" ]` in the script? Or using `input_mapping: { image: static-image }` if you prefer that?

concourse/tasks/upload-rails-assets.sh
```diff
@@ -0,0 +1,19 @@
...
+[profile default]
+role_arn = $ASSUME_ROLE_ARN
+credential_source = Ec2InstanceMetadata
+EOF
+
+echo "Uploading rails assets from $IMAGE_ASSETS_PATH to $S3_BUCKET_PATH ..."
+
+aws s3 sync \
+  $IMAGE_ASSETS_PATH \
+  $S3_BUCKET_PATH
```
richardTowers, line 17: These should be quoted to prevent word splitting (we should set up shellcheck as part of the tests for this repo 🤔)

terraform/deployments/govuk-publishing-platform/rails_assets.tf
```diff
@@ -0,0 +1,13 @@
+locals {
+  workspace_transformation = terraform.workspace == "default" ? "ecs" : terraform.workspace
```
richardTowers, line 2: I wonder if this would be nicer, since this `workspace_transformation` local is only used twice and in both cases it's the same interpolation?

```terraform
bucket_name = join("-", [
  "govuk",
  var.govuk_environment,
  terraform.workspace == "default" ? "ecs" : terraform.workspace,
  "rails-assets"
])
```

concourse/pipelines/deploy.yml
fredericfran-gds: cool, I can do that.

concourse/pipelines/deploy.yml
fredericfran-gds: I can fix that.

concourse/pipelines/deploy.yml
fredericfran-gds: I can do that

concourse/tasks/update-task-definition.yml
fredericfran-gds: I can do that

terraform/deployments/govuk-publishing-platform/rails_assets.tf
fredericfran-gds: I think this code will be pushed up somewhere else as this is used in many places in the code base.

fredericfran-gds (commented):
I will fix the comments. thanks.

terraform/deployments/govuk-publishing-platform/rails_assets.tf
richardTowers: True, but we already have `workspace_suffix`, `workspace_external_domain` etc. for that. Adding another form of the same thing doesn't improve the situation, unless you want to actually pull this up as part of this PR (which I think would be scope creep).

fredericfran-gds: I will pull it up in another PR, I need to fix some issues with the origin later anyway.

concourse/pipelines/deploy.yml
```diff
@@ -21,17 +21,10 @@ resources:
     icon: github
     name: govuk-infrastructure
     source: &govuk-infrastructure-source
-      branch: main
+      branch: rails_assets_uploads
```
richardTowers, line 24: This will need setting back before you merge.

richardTowers (approved):

concourse/pipelines/deploy.yml
fredericfran-gds: yes, doing a sanity check first in the pipeline :-)

richardTowers (approved):

PR #181, opened 2021-03-15
Fixup bootstrapping tasks

bilbof:
This includes a couple of fixes for bootstrapping and workspaces.

* The mesh URIs provided to apps as env vars are made workspace aware
* The bootstrap tasks come up with a task role

Once we get this merged we'll be able to bring up a working instance of the platform in a new workspace with one `terraform apply`.


fredericfran-gds (approved):
great work.

might need to update docs etc

PR #188, opened 2021-03-16
Stream task logs during deployment

richardTowers:
We switched to using the ecs-cli image for run-task so we could see
application logs during tasks. It would be nice to be able to see app
logs during deploys too.

ecs-cli needs a task ID to stream logs, so we need to fish that out of
`aws ecs list-tasks`. I've assumed that there will only be one taskArn
in the response for this call, but I'm not 100% sure that's true. The
code will just try to use the first one it finds.

No need to cleanup the background process, as concourse will do it for
us when it kills the container this script runs in.

You can see the output of this in this snowflaked pipeline run:
https://cd.gds-reliability.engineering/teams/govuk-tools/pipelines/deploy-apps-test/jobs/deploy-router/builds/224
(VPN required)

### DIFF
--- a/concourse/tasks/update-ecs-service.sh
+++ b/concourse/tasks/update-ecs-service.sh
@@ -18,16 +18,33 @@ fi
 
 echo "Updating $ECS_SERVICE service..."
 
+CLUSTER=govuk
+
 aws ecs update-service \
-  --cluster govuk \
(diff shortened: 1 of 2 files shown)


bilbof (approved):
Fantastic 👍

PR #192, opened 2021-03-18
Add origin cloudfront

fredericfran-gds:
### Please review #196 and #197 first.

Creates the public entry point (as opposed to the publishing one).
1. public requests should enter via:
    (a) Fastly CDN if using the default/ecs workspace or
    (b) AWS CloudFront for other workspaces
2. CloudFront has an AWS WAF ACL which only allows requests from Fastly or office IPs
3. CloudFront routes requests with path matching /assets/* to the S3 rails asset bucket
4. CloudFront routes all other requests to the origin ALB

Future work:
1. Tunning Cloudfront parameters (E.g. Should Fastly health checks be terminated at the ALB rather than sent to the container?)
2. Configure logging of AWS Cloudfront and WAF

Ref:
1. [trello card](https://trello.com/c/NRLwGdBJ/172-static-assets-in-ecs)
2. [PR for assets uploads](https://github.com/alphagov/govuk-infrastructure/pull/180)

### DIFF
--- a/terraform/deployments/govuk-publishing-platform/main.tf
+++ b/terraform/deployments/govuk-publishing-platform/main.tf
@@ -9,12 +9,17 @@ terraform {
   required_providers {
     aws = {
       source  = "hashicorp/aws"
-      version = "~> 3.13"
+      version = "~> 3.33"
     }
 
     fastly = {
       source  = "fastly/fastly"
       version = "0.24.0"
+    }
+
+    random = {
+      source  = "hashicorp/random"
+      version = "3.0.0"
     }
   }
 }
@@ -27,10 +32,21 @@ provider "aws" {
   }
 }
 
+provider "aws" {
+  region = "us-east-1"
+  alias  = "us_east_1"
+
+  assume_role {
+    role_arn = var.assume_role_arn
+  }
+}
+
 provider "fastly" {
   # We only want to use fastly's data API
   api_key = "test"
 }
+
+provider "random" {}
 
 locals {
   vpc_id                        = data.terraform_remote_state.infra_networking.outputs.vpc_id
--- a/terraform/deployments/govuk-publishing-platform/origins.tf
+++ b/terraform/deployments/govuk-publishing-platform/origins.tf
@@ -1,36 +1,82 @@
+resource "aws_acm_certificate" "public_north_virginia" {
+  provider    = aws.us_east_1
+  domain_name = "*.${local.workspace_external_domain}"
+
+  subject_alternative_names = local.is_default_workspace ? ["*.${local.workspace}.${var.publishing_service_domain}"] : null
+
+  validation_method = "DNS"
+
+  lifecycle {
+    create_before_destroy = true
+  }
+}
+
+resource "aws_route53_record" "public_north_virginia" {
+  provider = aws.us_east_1
+  for_each = {
+    for dvo in aws_acm_certificate.public_north_virginia.domain_validation_options : dvo.domain_name => {
+      name   = dvo.resource_record_name
+      record = dvo.resource_record_value
+      type   = dvo.resource_record_type
+    }
+  }
+
+  allow_overwrite = true
+  name            = each.value.name
+  records         = [each.value.record]
+  ttl             = 60
+  type            = each.value.type
+  zone_id         = aws_route53_zone.workspace_public.zone_id
+}
+
+resource "aws_acm_certificate_validation" "public_north_virginia" {
+  provider                = aws.us_east_1
+  certificate_arn         = aws_acm_certificate.public_north_virginia.arn
+  validation_record_fqdns = [for record in aws_route53_record.public_north_virginia : record.name]
+}
+
+
 module "www_origin" {
   source = "../../modules/origin"
 
-  vpc_id                    = local.vpc_id
-  public_subnets            = local.public_subnets
-  public_zone_id            = aws_route53_zone.workspace_public.zone_id
-  external_app_domain       = var.external_app_domain
-  certificate               = aws_acm_certificate.workspace_public.arn
-  publishing_service_domain = var.publishing_service_domain
-  workspace                 = local.workspace
-  external_cidrs_list       = concat(var.office_cidrs_list, data.fastly_ip_ranges.fastly.cidr_blocks)
+  vpc_id                               = local.vpc_id
+  aws_region                           = data.aws_region.current.name
+  assume_role_arn                      = var.assume_role_arn
+  public_subnets                       = local.public_subnets
+  public_zone_id                       = aws_route53_zone.workspace_public.zone_id
+  external_app_domain                  = aws_route53_zone.workspace_public.name
+  load_balancer_certificate_arn        = aws_acm_certificate.workspace_public.arn
+  cloudfront_certificate_arn           = aws_acm_certificate.public_north_virginia.arn
+  publishing_service_domain            = var.publishing_service_domain
+  workspace                            = local.workspace
+  is_default_workspace                 = local.is_default_workspace
+  external_cidrs_list                  = concat(var.office_cidrs_list, data.fastly_ip_ranges.fastly.cidr_blocks)
+  rails_assets_s3_regional_domain_name = aws_s3_bucket.rails_assets.bucket_regional_domain_name
 
   apps_security_config_list = {
     "frontend" = { security_group_id = module.frontend.security_group_id, target_port = 80 },
-    "static"   = { security_group_id = module.static.security_group_id, target_port = 80 },
   }
 }
 
 module "draft_origin" {
   source = "../../modules/origin"
 
-  vpc_id                    = local.vpc_id
-  public_subnets            = local.public_subnets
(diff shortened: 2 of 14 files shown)


richardTowers (commented):
Sorry - half finished review - comments might still be useful.

terraform/deployments/govuk-publishing-platform/app_publisher.tf
```diff
@@ -14,7 +14,6 @@ locals {
     environment_variables = merge(
       local.defaults.environment_variables,
       {
-        ASSET_HOST                       = local.defaults.asset_host
```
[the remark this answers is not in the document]
fredericfran-gds: We only did for static and frontend when exported assets to the s3 buckets in https://github.com/alphagov/govuk-infrastructure/pull/180. I need to check for publisher and signon if these are valid static assets export or just added.

terraform/modules/origin/alb.tf
```diff
@@ -0,0 +1,78 @@
...
+  priority     = 100
+
+  action {
+    type             = "forward"
+    target_group_arn = aws_lb_target_group.origin-frontend.arn
+  }
+
+  condition {
+    http_header {
+      http_header_name = "X-Custom-Header"
```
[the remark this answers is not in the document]
fredericfran-gds: I can do that

terraform/modules/origin/main.tf
```diff
@@ -1,97 +1,161 @@
...
+}
+
+resource "aws_secretsmanager_secret" "origin_alb_x_custom_header_secret" {
+  name = "${local.mode}_origin_${var.workspace}_alb_x_custom_header_secret"
+}
+
+resource "aws_secretsmanager_secret_version" "origin_alb_x_custom_header_secret" {
+  secret_id     = aws_secretsmanager_secret.origin_alb_x_custom_header_secret.id
+  secret_string = random_password.origin_alb_x_custom_header_secret.result
+}
```
richardTowers, line 54: Does using secrets manager actually buy us anything here, compared to just passing `random_password.origin_alb_x_custom_header_secret.result` directly into the `aws_cloudfront_distribution`?

My understanding is that terraform will still need to keep track of the secret value in the state file, so putting it in SecretsManager is just duplicating it.

I appreciate there's some thinking we need to do on how to rotate this secret, which maybe secrets manager will help with? I'd lean towards keeping it simple for now though, and raising a trello card for rotation.

terraform/modules/origin/main.tf
```diff
@@ -1,97 +1,161 @@
...
+      http_port              = 80
+      https_port             = 443
+      origin_protocol_policy = "https-only"
+      origin_ssl_protocols   = ["TLSv1.2"]
+    }
 
-  condition {
-    path_pattern {
-      values = ["/assets/frontend/*"]
+    #TODO: frequent rototation for additional security,
```
richardTowers, line 68: I suspect that this is a TODO which will never get prioritized. Could you raise a trello card for it, and maybe link to that here?

Something like "Automate the rotation of the X-Cloudfront-Token secret"?

terraform/modules/origin/main.tf
```diff
@@ -1,97 +1,161 @@
...
 
-resource "aws_lb_listener_rule" "origin-static" {
-  listener_arn = aws_lb_listener.origin.arn
-  priority     = 99
+  enabled         = true
+  is_ipv6_enabled = false
+  comment         = "${local.mode}-origin ${var.workspace} CDN in front of origin ALB and s3 rails assets bucket"
+  web_acl_id      = aws_wafv2_web_acl.origin_cloudfront_web_acl.arn
+
+  aliases = var.workspace == "ecs" && local.mode == "www" ? ["www.ecs.${var.publishing_service_domain}", "www-origin.${var.external_app_domain}"] : ["${local.mode}-origin.${var.external_app_domain}"]
```
richardTowers, line 91: You could use `var.live` instead of `local.mode == "www"`. Would be nice to have a similar thing for `workspace`, so we can check `var.is_default_workspace && var.live`. (Related thought: `is_live` might be a bit more self explanatory than `live`)

I think you could also avoid the repetition of the external app domain one with something like:

```terraform
  aliases = compact([
    var.is_default_workspace && local.is_live ? "www.ecs.${var.publishing_service_domain}" : null,
    "${local.mode}-origin.${var.external_app_domain}"
  ])
```

terraform/modules/origin/security_groups.tf
```diff
@@ -15,13 +15,14 @@ resource "aws_security_group_rule" "service_from_origin_alb_http" {
   source_security_group_id = aws_security_group.origin_alb.id
 }
 
-resource "aws_security_group_rule" "origin_alb_from_any_https" {
-  description       = "${local.mode}-origin ALB allows requests from CIDRs list over HTTPS"
+# We use a secret custom header to authenticate request via the origin ALB
```
richardTowers, line 18: Maybe move this comment down to the `cidr_blocks` line?

richardTowers (commented):
Excellent work. This general approach looks good to me, and the configuration looks solid. I've tested this out in the `towers` workspace, and it works like a charm. It's pretty exciting to see it all come together.

A few more comments (mostly stylistic things) then I'm happy for this to go 🚀

terraform/deployments/govuk-publishing-platform/main.tf
```diff
@@ -27,11 +32,22 @@ provider "aws" {
   }
 }
 
+provider "aws" {
+  region = "us-east-1"
+  alias  = "use1"
```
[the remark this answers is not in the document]
fredericfran-gds: thanks, I can do that.

terraform/deployments/govuk-publishing-platform/origins.tf
```diff
@@ -1,36 +1,80 @@
+resource "aws_acm_certificate" "workspace_public_nv" {
```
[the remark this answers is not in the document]
fredericfran-gds: thanks, I can do that.

terraform/modules/origin/README.md
```diff
@@ -1,5 +1,11 @@
...
- 
+Creates the public traffic entry point to the environment(as opposed to the publishing one).
+
+The traffic flow is as follows:
+1. it is intended that public requests enters via:
+    (a) Fastly CDN if using default/ecs workspace or
+    (b) AWS CloudFront CDN directly for other workspaces (defined in this module)
+2. CloudFront is protected via AWS WAF ACL which allows request coming only Fastly and office IPs only
+3. CloudFront routes requests with path matching /assets/* to S3 rails asset bucket or
+   origin Application Load-balancer which is connected to the frontend for now until router is ready.
```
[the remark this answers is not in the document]
fredericfran-gds: thanks, I can do that.

terraform/modules/origin/main.tf
```diff
@@ -1,97 +1,161 @@
 locals {
-  mode = var.live ? "www" : "draft"
+  mode          = var.live ? "www" : "draft"
```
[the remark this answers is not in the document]
fredericfran-gds: thanks, I can do that.

terraform/modules/origin/main.tf
fredericfran-gds: The main reason why I added the "password" to secrets manager was so that I can retrieve it easily from AWS console for testing. Probably could do the same with terraform state.

terraform/modules/origin/main.tf
fredericfran-gds: thanks, I can do that.

terraform/modules/origin/security_groups.tf
fredericfran-gds: thank. I can do that.

terraform/modules/origin/main.tf
fredericfran-gds: thanks, I can do that

terraform/modules/origin/variables.tf
```diff
@@ -43,3 +47,22 @@ variable "public_zone_id" {
 variable "workspace" {
   type = string
 }
+
+variable "is_default_workspace" {
+  type    = bool
+  default = true
```
richardTowers, line 53: I think it would be safer not to have the default here - callers of this module should always be passing this in explicitly (otherwise they won't be workspace aware).

terraform/modules/origin/main.tf
```diff
@@ -1,97 +1,154 @@
...
+
+resource "aws_cloudfront_origin_access_identity" "cloudfront_s3_access" {
+  comment = "${local.live_or_draft_prefix}-origin ${var.workspace} cloudfront accessing the Rails assets s3 bucket"
+}
 
-  condition {
-    path_pattern {
-      values = ["/assets/frontend/*"]
+resource "random_password" "origin_alb_x_custom_header_secret" {
+  length  = 128
```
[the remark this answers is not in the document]
fredericfran-gds: ok, can do.

terraform/modules/origin/variables.tf
fredericfran-gds: ok, can do

richardTowers (approved):

PR #193, opened 2021-03-19
Pull update-pipeline to the left of the pipeline

richardTowers:
While the run-terraform task definition is inlined, it makes sense to run it after the pipeline is up-to-date. Otherwise we get a weird race condition where the pipeline and run-terraform update at the same time.

| Before | After |
|-----|-----|
| image | image |

(Note: ignore the background image, that's a red herring)


bilbof (approved):
🦖

PR #194, opened 2021-03-19
Add tflint config for the workspaces plugin

richardTowers:
See https://github.com/richardTowers/tflint-ruleset-workspaces/ (I wrote this, so it's probably worth a quick review of that codebase as part of this PR).

Note that we're linting a non-default workspace because in the `default` workspace resources are namespaced as "-govuk" or "-ecs", which the plugin doesn't understand.

Going to mark this as draft for now so I can push some broken commits to demonstrate that it works.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,10 +63,15 @@ jobs:
       - name: run tflint
         env:
           TFLINT_VERSION: v0.24.1
+          TFLINT_RULESET_WORKSPACES_VERSION: 0.2.0
         run: |
           TEMP_PATH="$(mktemp -d)"
           PATH="${TEMP_PATH}:$PATH"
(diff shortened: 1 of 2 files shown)


richardTowers:
Example failing lint because of ac37851 in this action:

https://github.com/alphagov/govuk-infrastructure/pull/194/checks?check_run_id=2149108336

```
1 issue(s) found:

Error: aws_ecs_cluster resource name "govuk" does not include the workspace (example_workspace) (aws_ecs_cluster_name_includes_workspace)

  on ecs_cluster.tf line 10:
  10:   name               = var.mesh_name

Reference: https://github.com/richardTowers/tflint-ruleset-workspaces/blob/master/README.md
```

richardTowers:
Closing this as it's way too far behind `main` to be worth rebasing.

PR #195, opened 2021-03-22
Allow force_destroy assets S3 bucket in non-prod

richardTowers:
Unless we're in staging or production, it's okay to delete this bucket
even if it still contains objects.

In the lower environments, we want to be able to run `terraform destroy`
without having to manually delete all the assets from the bucket.


MahmudH (approved):
Looked over the docs: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket#force_destroy.

LGTM 👍

PR #196, opened 2021-03-23
Refactor workspace

fredericfran-gds:
At entry point of variables definition, defaults.tf, we map
newly defined variable `workspace` to `ecs` for default terraform.workspace
or value of terraform workspace. This variable is then used throughout the code
base for simplicity.

Other main fixes:
1. main cluster is named `govuk-<workspace>`
2. mesh name is `mesh.<workspace>.<internal_domain_name>`

PR #197, opened 2021-03-23
Fix certificates

fredericfran-gds:
### Please review #196 first.

Now,
1. for default/ecs workspace, we use: a single certificate with
   *.ecs.test.govuk.digital and *.ecs.test.publishing.service.gov.uk
   on both relevant ALBs and Cloudfronts
2. for other workspaces, we use only a certificate with
   *.ecs.<workspace>.govuk.digital

Ref:
1. [pr for dns validation *.ecs.test.publishing.service.gov.uk](alphagov/govuk-dns-config@9ea5f1f)

### DIFF
--- a/terraform/deployments/govuk-publishing-platform/dns.tf
+++ b/terraform/deployments/govuk-publishing-platform/dns.tf
@@ -31,6 +31,8 @@ resource "aws_route53_zone" "internal_private" {
 resource "aws_acm_certificate" "workspace_public" {
   domain_name = "*.${local.workspace_external_domain}"
 
+  subject_alternative_names = local.is_default_workspace ? ["*.${local.workspace}.${var.publishing_service_domain}"] : null
+
   validation_method = "DNS"
 
   lifecycle {
@@ -57,5 +59,5 @@ resource "aws_route53_record" "workspace_public" {
 
 resource "aws_acm_certificate_validation" "workspace_public" {
   certificate_arn         = aws_acm_certificate.workspace_public.arn
-  validation_record_fqdns = [for record in aws_route53_record.workspace_public : record.fqdn]
+  validation_record_fqdns = [for record in aws_route53_record.workspace_public : record.name]
 }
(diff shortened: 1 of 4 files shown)


terraform/deployments/govuk-publishing-platform/dns.tf
```diff
@@ -31,6 +31,8 @@ resource "aws_route53_zone" "internal_private" {
 resource "aws_acm_certificate" "workspace_public" {
   domain_name = "*.${local.workspace_external_domain}"
 
+  subject_alternative_names = local.workspace == "ecs" ? ["*.${local.workspace}.${var.publishing_service_domain}"] : null
```
richardTowers, line 34: I wonder if `terraform.workspace == "default"` would be clearer here. Possibly less likely to break if we decide to change the default local.workspace name too.

terraform/deployments/govuk-publishing-platform/dns.tf
```diff
@@ -57,5 +59,5 @@ resource "aws_route53_record" "workspace_public" {
 
 resource "aws_acm_certificate_validation" "workspace_public" {
   certificate_arn         = aws_acm_certificate.workspace_public.arn
-  validation_record_fqdns = [for record in aws_route53_record.workspace_public : record.fqdn]
+  validation_record_fqdns = [for record in aws_route53_record.workspace_public : record.name]
```
richardTowers, line 62: Is this right? `validation_record_fqdns` makes it sound like it wants FQDNs, not subdomains? Am I misunderstanding?

terraform/deployments/govuk-publishing-platform/dns.tf
fredericfran-gds: I want the terraform workspace to exist in `default.tf` only, it is confusing to see terraform.workspace/workspace side by side in other places.

terraform/deployments/govuk-publishing-platform/dns.tf
fredericfran-gds: In theory you are right it should be done this way, the issue is we don't control `*.publishing.service.gov.uk` domain in this repo, which makes it tedious to get the zone_id to include in `aws_route53_record.workspace_public`. 

This is the workaround I found.

terraform/deployments/govuk-publishing-platform/dns.tf
richardTowers: What do you think about adding a `local.is_default_workspace` variable in that file then?

fredericfran-gds: happy to add that to defaults.tf

richardTowers (approved):

PR #201, opened 2021-04-01
Add container healthcheck for Envoy

bilbof:
We saw intermittent connection errors which were caused by Envoy not being ready to receive requests. This change ensures the Envoy process is up and healthy before starting the app container.


smford (approved):
LGTM

PR #202, opened 2021-04-06
Add SecretsManager rotation lambda for Signon bearer tokens

bilbof:
Note: This branch should be merged after the [signon bootstrapping branch](https://github.com/alphagov/signon/compare/bilbof/boostrapping) is released.

This implements a custom [Lambda rotation function](https://docs.aws.amazon.com/secretsmanager/latest/userguide/rotating-secrets-lambda-function-overview.html) to create and rotate bearer tokens issued by [Signon](https://github.com/alphagov/signon) for app to app authentication.

Story: https://trello.com/c/Q2j0LyDO/419.

**Motivation**

The motivation for this change is that it will enable us to bootstrap Signon in a new environment. When creating a new environment, or bringing a version of the platform up in an existing environment, we won't need to manually create many bearer tokens by hand and add them to SecretsManager.

Another benefit of this change is that we can entirely automate the rotation of bearer tokens. This will also save us time, and will enable us to rotate bearer tokens much more frequently than is currently the case.

**Design**

I recommend reviewing the [documentation](https://docs.aws.amazon.com/secretsmanager/latest/userguide/rotating-secrets-lambda-function-overview.html) for a custom rotation Lambda.

There will be a lambda per bearer token.

During bootstrapping of Signon the lambda is responsible for creating the first bearer tokens for an app. It can take some time (minutes) for this to complete; the lambda will retry creation of a bearer token until signon is spun up. Other apps will not start until their bearer tokens have been created in Signon.

The lambda creates a Signon bearer token for an app using Signon's new API. The API is protected by IP restrictions (added in this PR) and a bearer token generated per-workspace. Currently we use Terraform's [random_password](https://registry.terraform.io/providers/hashicorp/random/latest/docs/resources/password) to generate the API token. Since we do not want secrets stored in our Terraform statefile (as using `random_password` would cause), in future we'll replace `random_password` with _another_ lambda which will create the token.

SecretsManager is responsible for trigger the lambda to rotate a bearer token every 30 days. To rotate a token Signon will issue a new token, without modifying the existing token. In future Signon will delete old, unused bearer tokens that are configured for rotation.

Apps do not immediately pick up new versions of secrets stored in SecretsManager. In a future piece of work we'll need to set up apps to cycle when necessary (ideally on a daily schedule, as is more or less the case).

Further work is required to apply a similar approach to OAuth application secrets issued by Signon.


terraform/modules/public-load-balancer/path_restriction_rule.tf
```diff
@@ -0,0 +1,53 @@
+locals {
+  # Needed due to an AWS quota restriction (limit of 5 conditions per rule)
+  # In future we may need to request a service quota increase of
+  # "Rules per Application Load Balancer" from AWS.
+  rules = setproduct(var.restricted_paths, var.unrestricted_cidrs)
+}
+
+resource "aws_lb_listener_rule" "allow_unrestricted_cidrs" {
```
sengi, line 8: This is quite complex and uses `count`, so it'll likely do some quite unintuitive things when we make changes to the allowlists. It might also have complex failure modes if someone adds one too many entries. I'm wondering if it'd be the lesser of evils just to define the aws_lb_listener_rules where we call the module - i.e. let the client customise the rules in the few cases where we need. I guess we still need some way to not have a ton of duplicated stuff for the basic office-only allowlist in staging/test. Perhaps a separate module for generating the forwarding rules might be the way to go? (thinking SRP) Could then evolve it a bit more easily over time?

sengi (commented):
Haven't reviewed the Ruby stuff, just the TF. Overall solution looks great but I'm a bit concerned about how the path_restriction_rule stuff might pan out in reality - this kind of complex stuff in TF can easily become a tarpit where simple changes (like adding/removing an entry from an allowlist) lead to unpredictable / hard-to-understand effects or even changes that can't be rolled out. Maybe I'm worrying unduely though.

Usually I'd recommend splitting something like this into smaller PRs but I don't want to create even more work for you on this one - I'd just get someone to review the Ruby stuff.

lambdas/signon_bearer_token_rotater.rb
```diff
@@ -0,0 +1,216 @@
...
+    logger.info("finishSecret: Version #{token} already marked as AWSCURRENT for #{arn}")
+    return
+  end
+
+  # Finalize by staging the secret version current
+  service_client.update_secret_version_stage(
+    secret_id: arn,
+    version_stage: 'AWSCURRENT',
+    move_to_version_id: token,
+    remove_from_version_id: current_version
```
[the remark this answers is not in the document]
bilbof: Good point. It shouldn't cause any issues - it's OK for there to be no current version. I've verified that the Lambda can create and finish a secret rotation when there is no current version.

The [update_secret_version_stage docs](https://docs.aws.amazon.com/sdk-for-ruby/v2/api/Aws/SecretsManager/Client.html#update_secret_version_stage-instance_method) say that `current_version` must not be nil only if there is a current version.

lambdas/signon_bearer_token_rotater.rb
```diff
@@ -0,0 +1,216 @@
...
+    version_stages: ['AWSPENDING']
+  )
+
+  logger.info("createSecret: Successfully put secret for ARN #{arn} and version #{token}.")
+end
+
+# Set the secret
+# This method should set the AWSPENDING secret in Signon
+# Should take the value of the AWSPENDING secret and set the secret to
+# this value in the database.
```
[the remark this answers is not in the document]
bilbof: Thanks, I've changed the method name and corrected the comment. It's a bit confusing that we're not following the Lambda guide exactly, though it avoids making larger changes to signon.

bilbof:
@sengi @karlbaker02 thanks for your reviews. I've made the changes you recommended to the commits.

One thing that's unsolved is the path restriction implementation - I've left a note that we should come back to this.

LMK if the PR is too big to review, I'd be happy to split it up, otherwise it might be easier to review by commit.

terraform/modules/origin/variables.tf
```diff
@@ -22,7 +22,7 @@ variable "apps_security_config_list" {
   description = "map in the format {<app_name> = { security_group_id=<security_group_id>, target_port=<target_port>}}"
 }
 
-variable "external_cidrs_list" {
+variable "cidrs_allowlist" {
```
sengi, line 25: `allowlist_cidrs` like the other ones? (and similarly elsewhere?)

terraform/modules/alb-listener-ip-restriction-rules/main.tf
```diff
@@ -0,0 +1,56 @@
+locals {
+  # Note: Needed due to an AWS quota restriction (limit of 5 conditions per rule)
+  # In future we may need to request a service quota increase of
+  # "Rules per Application Load Balancer" from AWS.
+  rules = setproduct(var.restricted_path_patterns, var.fully_trusted_source_ips)
+}
+
+# TODO: Uses count: This may lead to unintuitive behaviour when adding/removing
+# restricted_path_patterns and fully_trusted_source_ips. Impact could be
+# reduced by letting client assign priority to rules.
```
sengi, line 10: My hunch is that we might need to do that in order to keep this maintainable in practice - I suspect this is gonna be a supercharged version of the problematic [lb-listener-rules](https://github.com/alphagov/govuk-aws/blob/master/terraform/modules/aws/lb_listener_rules/main.tf) and friends from govuk-aws (which became so brittle we can no longer add/remove things from it without major terraform-state surgery) but I might be wrong, so let's see how it goes.

terraform/modules/public-load-balancer/variables.tf
```diff
@@ -1,3 +1,9 @@
+variable "allowlist_cidrs" {
+  type        = list(string)
+  default     = ["0.0.0.0/0"]
+  description = "Allowlist for HTTPS access (not necessarily all paths)."
```
sengi, line 4: Can we make it more specific? e.g. would this be accurate?
```suggestion
  description = "Allowlist for HTTPS access (but doesn't allow access to paths in restricted_path_patterns)."
```

sengi (approved):
LGTM for Terraform stuff. Haven't looked at the Ruby.

karlbaker02 (approved):
Ruby looks good to me 💯

bilbof:
@karlbaker02 @sengi I've done a final fix up of a commit, could I request a re-review?

sengi (approved):

PR #203, opened 2021-04-08
Fastly cdn terraform

fredericfran-gds:
Manage Fastly CDN Service with Terraform

As part of the [firebreak
project](https://trello.com/c/crnmDAG6/450-%F0%9F%94%A5%F0%9F%8C%B4%F0%9F%8C%B4-manage-fastly-configuration-with-terraform-in-test),
this PR adds the necessary terraform code to create a new Fastly service
for the test GOV.UK environment.

This Fastly service is a stripped down version of the one already in
integration except that there are no tables, cookies etc but basic
authentication is present.

Fastly service URL is: https://www1.ecs.test.publishing.service.gov.uk/

**Deployment**

In order to deploy this code, you need to export the environment
variable `FASTLY_API_KEY` with the key being stored under
`fastly/engineer_test_environment` in the 2ndline Pass (see
[PR](alphagov/govuk-secrets#1132))

**Caveats**
Known caveats:
1. Code depends on the Basic authentication encoded secret to exist in
AWS Secrets Manager under the name: `CDN_basic_authentication_encoded_secret`
2. the TLS certificate for `*.ecs.test.publishing.service.gov.uk` was
already created in Fastly portal for another service and is reused here.
3. the size of the frontend needs to be increased in terms of CPU since twice
the number of Fastly cache nodes where pointing to it since there will be
2 Fastly services.

### DIFF
--- a/terraform/deployments/govuk-publishing-platform/cdn.tf
+++ b/terraform/deployments/govuk-publishing-platform/cdn.tf
@@ -1,0 +1,23 @@
+resource "fastly_service_v1" "www_service" {
+  count = local.is_default_workspace ? 1 : 0
+  name  = "www_service_${var.govuk_environment}"
+
+  domain {
+    name    = "www1.ecs.${var.publishing_service_domain}" #TODO: replace www1 with www as this is for testing only
+    comment = "www CDN entry point for the ${var.govuk_environment} GOV.UK environment"
+  }
+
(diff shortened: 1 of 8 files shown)


fredericfran-gds:
it's outdated and time to remove. People can still search for it if they want some prior art in this area.

PR #205, opened 2021-04-14
Make the pipeline workspace aware

richardTowers:
Terraform and concourse work together to deploy the infrastructure and the apps. Running `terraform apply` gets you most of the way there, but to get a fully working workspace you also need to upload the CSS / JS / fonts generated by the rails apps to S3. Concourse does this (amongst other things).

Workspaces are therefore much more useful if each workspace has its own concourse pipeline.

Engineers working on a branch can use the workspace to test changes to the concourse pipeline and terraform before they raise a PR, without affecting the rest of the team.

Test created was:
terraform_workspace: bob
entry_point: https://www-origin.bob.test.govuk.digital/
concourse_pipeline: https://cd.gds-reliability.engineering/teams/govuk-tools/pipelines/deploy-bob

### DIFF
--- a/concourse/pipelines/deploy-parameters.yml
+++ b/concourse/pipelines/deploy-parameters.yml
@@ -1,0 +1,2 @@
+workspace: default
+govuk_infrastructure_branch: main
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -37,7 +37,7 @@ resources:
     icon: github
     name: govuk-infrastructure
     source: &govuk-infrastructure-source
-      branch: main
+      branch: ((govuk_infrastructure_branch))
       uri: https://github.com/alphagov/govuk-infrastructure
     type: git
 
@@ -329,9 +329,8 @@ jobs:
       trigger: true
     - file: govuk-infrastructure/concourse/pipelines/deploy.yml
       set_pipeline: self
-      vars:
-        workspace: ((workspace))
-        pipeline_name: ((pipeline_name))
+      var_files:
+        - govuk-infrastructure/concourse/pipelines/deploy-parameters.yml
 
   - name: run-terraform
     serial: true
@@ -522,7 +521,8 @@ jobs:
           app-image: frontend-image
         params:
           IMAGE_ASSETS_PATH: app-image/rootfs/app/public/assets/frontend
-          S3_BUCKET_PATH: s3://govuk-test-ecs-rails-assets/assets/frontend/
+          S3_BUCKET_PATH_PATTERN: s3://govuk-test-%s-rails-assets/assets/frontend/ # %s will be replaced by interpolated workspace in upload-rails-assets.yml
+          WORKSPACE: ((workspace))
       - task: update-draft-task-definition
         file: govuk-infrastructure/concourse/tasks/update-task-definition.yml
         input_mapping:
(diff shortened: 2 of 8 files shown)


concourse/pipelines/deploy.yml
```diff
@@ -327,7 +327,10 @@ jobs:
     - get: govuk-infrastructure
       trigger: true
     - file: govuk-infrastructure/concourse/pipelines/deploy.yml
-      set_pipeline: deploy-apps-test
+      set_pipeline: ((pipeline_name))
```
bilbof, line 330: I think one can use `self` here. It might be deprecated in future, but should be good for now.

concourse/pipelines/deploy.yml
```diff
@@ -422,6 +426,9 @@ jobs:
             cd govuk-infrastructure/terraform/deployments/govuk-publishing-platform
 
             terraform init -backend-config "role_arn=$ASSUME_ROLE_ARN"
+
+            terraform workspace select "$WORKSPACE"
```
bilbof: Sorry I think we will need to do the same for the `destroy-terraform` job.

concourse/pipelines/deploy.yml
```diff
@@ -327,7 +327,9 @@ jobs:
     - get: govuk-infrastructure
       trigger: true
     - file: govuk-infrastructure/concourse/pipelines/deploy.yml
-      set_pipeline: deploy-apps-test
+      set_pipeline: ((pipeline_name))
+      var_files:
+        - govuk-infrastructure/concourse/pipelines/deploy-parameters.yml
```
bilbof, line 332: I wonder if we should replace `var_files` with `instance_vars`? This would enable us to bring up a pipeline for a new workspace, without needing to push a branch to GitHub.

```yaml
instance_vars
      workspace: ((workspace))
      govuk_infrastructure_branch: ((govuk_infrastructure_branch))
```

Apparently `instance_vars` is currently experimental and requires the feature flag `--enable-pipeline-instances` but I think it's worth switching to it, even if we later need to use a config file.

bilbof:
I see `slack_webhook` is unset.. Perhaps we should set this globally as the `deploy_apps_slack_webhook`. Not a blocker for merging this though.

concourse/pipelines/deploy.yml
fredericfran-gds: Yeah, I can do that

concourse/pipelines/deploy.yml
fredericfran-gds: yep

concourse/pipelines/deploy.yml
fredericfran-gds: I prefer to keep it out, easier for a newcomer to see which file they need to modify without modifying their stuff.

I couldn't see where to pass the flag also easily

richardTowers: I think `instance_vars` is new in concourse 7, and we're still on 6. So I think we need to stick with plain ol' vars for now. Might be worth asking autom8 if they'll enable `--enable-pipeline-instances` for us when they upgrade, if we want to get on the bleeding edge though.

concourse/pipelines/deploy-parameters.yml
```diff
@@ -0,0 +1,2 @@
+workspace: bob
+govuk_infrastructure_branch: test_workspace
```
bilbof, line 2: Should we change these back to `default` and `main`?

bilbof (dismissed):

bilbof:
I've added a minor comment, but no need for a re-review, thanks for doing this!

concourse/pipelines/deploy-parameters.yml
fredericfran-gds: yeah, that's the issue with playing around too much with it. thanks for catching this.

bilbof (approved):

PR #206, opened 2021-04-14
Add support for ECS exec-command

bilbof:
The AWS ECS team recently released ECS Exec Command which permits us to run an interactive shell in a
running container: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-exec.html

This change introduces the policies to enable the feature. We won't be enabling this feature on our ECS Services, only on tasks in the `task_runner` cluster - as we don't want to give developers access to containers serving user requests.

Further work is required to improve the tooling - currently one needs to run a task and then use `aws ecs exec-command` to start an SSM session.

The workflow currently will look like this:

```
cd govuk-infrastructure
./tools/run-task.sh -a publisher -v web sleep 1800
# get task id from output of task
aws ecs execute-command --cluster task_runner --container app --command /bin/bash --interactive --task <task-id>
```

### DIFF
--- a/terraform/modules/container-definition/main.tf
+++ b/terraform/modules/container-definition/main.tf
@@ -9,6 +9,9 @@ output "json_format" {
       command = ["/bin/bash", "-c", var.health_check]
     }
     image = var.image
+    linuxParameters = {
+      initProcessEnabled = true
+    }
     logConfiguration = {
       logDriver = "awslogs",
       options = {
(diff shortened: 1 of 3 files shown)


terraform/modules/container-definition/main.tf
```diff
@@ -9,6 +9,9 @@ output "json_format" {
       command = ["/bin/bash", "-c", var.health_check]
     }
     image = var.image
+    linuxParameters = {
+      initProcessEnabled = true
```
sengi, line 13: Interesting. This might also help in case any of our containers happen to be less than well-behaved in the zombie department 🧟 Apparently it uses https://github.com/krallin/tini

bilbof: Awesome, I didn't know about Tini, thanks!

PR #207, opened 2021-04-14
Temporary hack to authenticate to DockerHub.

sengi:
Authenticate to DockerHub when pulling images, so that we don't get severely rate-limited.

I had to set the secret manually. Since it's just a temporary workaround for the test account until we're up and running with ECR, I think that's ok.

Briefly considered passing the ARN in as a var, but this workaround needs to be reverted before we deploy anything outside the test account anyway, so I decided it would just lend this a false air of legitimacy 😅


bilbof (dismissed):
LGTM, thanks for fixing this

sengi:
Woohoo, obsoleted by #208 🎉 Thanks, @fredericfran-gds !

PR #208, opened 2021-04-15
Change from Docker Hub to AWS ECR

fredericfran-gds:
GOV.UK containers are already being built and pushed into our own AWS
ECR (see #149 and #158).

This PR involves changing the code base so that the task definitions
refer to the images in the AWS ECR rather than Docker Hub.

Other work include:
1. giving permission to the test environment to pull from the AWS ECR in
production
2. creating a concourse user which is able to pull from AWS ECR so that
concourse can monitor new container versions and automatically deploy
them

Some caveats:
1. for content-store and publishing-api, we were using special versions
for our ECS environment. Therefore, we have pushed these container
versions from Docker Hub to AWS ECR. Hence, if these versions are
modified, these container images will have to be updated manually.
(see
[ecr manual](https://docs.aws.amazon.com/AmazonECR/latest/userguide/getting-started-cli.html))

Ref:
1. [trello card](https://trello.com/c/PcU9eVue/461-migrate-from-docker-hub-to-aws-ecr)

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -16,6 +16,14 @@ resource_types:
       repository: governmentpaas/s3-resource
       tag: latest # TODO - don't use latest (once we've worked out a policy for third party images)
 
+  - name: registry-image
+    type: docker-image
+    source:
+      repository: concourse/registry-image-resource
+      tag: 1.2.1
+      username: ((docker_hub_username))
+      password: ((docker_hub_authtoken))
+
 resources:
   - &git-repo
     icon: github
@@ -24,20 +32,6 @@ resources:
       branch: main
       uri: https://github.com/alphagov/govuk-infrastructure
     type: git
-
-  - <<: *git-repo
-    name: publisher
-    source:
-      branch: master
-      uri: https://github.com/alphagov/publisher
-      tag_filter: release_*
-
-  - <<: *git-repo
-    name: publishing-api
-    source:
-      branch: master
-      uri: https://github.com/alphagov/publishing-api
-      tag_filter: release_*
 
   - name: publishing-api-terraform-outputs
     type: s3
@@ -49,46 +43,11 @@ resources:
       initial_version: "0"
 
   - <<: *git-repo
-    name: content-store
-    source:
-      branch: master
-      uri: https://github.com/alphagov/content-store
-      tag_filter: release_*
-
-  - <<: *git-repo
     name: govuk-infrastructure-concourse-tasks
     icon: concourse-ci
     source:
       <<: *govuk-infrastructure-source
       paths: [ concourse/tasks ]
-
-  - <<: *git-repo
-    name: router
-    source:
-      branch: master
-      uri: https://github.com/alphagov/router
-      tag_filter: release_*
-
-  - <<: *git-repo
-    name: router-api
-    source:
-      branch: master
-      uri: https://github.com/alphagov/router-api
-      tag_filter: release_*
-
-  - <<: *git-repo
-    name: signon
-    source:
-      branch: master
-      uri: https://github.com/alphagov/signon
-      tag_filter: release_*
-
-  - <<: *git-repo
-    name: smokey
-    source:
-      branch: main
-      uri: https://github.com/alphagov/smokey
-      tag_filter: "*-release"
 
   - name: smokey-terraform-outputs
     type: s3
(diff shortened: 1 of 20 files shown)


concourse/pipelines/deploy.yml
```diff
@@ -16,6 +16,14 @@ resource_types:
       repository: governmentpaas/s3-resource
       tag: latest # TODO - don't use latest (once we've worked out a policy for third party images)
 
+  - name: registy-image-v2
```
sengi, line 19: typo
```suggestion
  - name: registry-image-v2
```

concourse/pipelines/deploy.yml
```diff
@@ -183,23 +142,68 @@ resources:
       initial_version: "0"
 
 
-  - name: frontend-image
-    type: registry-image
+  - &registry-image
+    name: frontend-image
+    type: registy-image-v2
```
sengi, line 147:
```suggestion
    type: registry-image-v2
```

concourse/pipelines/deploy.yml
```diff
@@ -183,23 +142,68 @@ resources:
...
     icon: docker
+    source: &registry-source
+      repository: frontend
+      aws_role_arn: arn:aws:iam::172025368201:role/pull_images_from_ecr_role
+      aws_region: eu-west-1
+      aws_access_key_id: ((concourse-ecr-readonly-user_aws-access-key))
+      aws_secret_access_key: ((concourse-ecr-readyonly-user_aws-secret-access-key))
```
sengi, line 154: typo - check if the actual secret name needs to be fixed too
```suggestion
      aws_secret_access_key: ((concourse-ecr-readonly-user_aws-secret-access-key))
```

terraform/deployments/ecr/main.tf
```diff
@@ -20,6 +20,20 @@ provider "aws" {
   region = "eu-west-1"
 }
 
+locals {
+  repositories = ["content-store",
```
sengi, line 24:
```suggestion
  repositories = [
    "content-store",
```

terraform/deployments/ecr/main.tf
```diff
@@ -176,3 +190,80 @@ resource "aws_iam_role_policy_attachment" "push_to_ecr_role_attachment" {
   role       = aws_iam_role.push_image_to_ecr_role.name
   policy_arn = aws_iam_policy.push_image_to_ecr_policy.arn
 }
+
+resource "aws_ecr_repository_policy" "pull_images_from_ecr_policy_policy" {
+  for_each   = toset(local.repositories)
+  repository = each.key
+  policy     = <<POLICY
```
sengi, line 197: Could you change these to use `jsonencode()` please, so that Terraform can syntax-check them? (It's an easy change - literally all you need to do is write `jsonencode(...)` instead of `<<EOF ... EOF`, then run `terraform fmt` to keep the linter happy. There's an example [here](https://github.com/alphagov/govuk-infrastructure/compare/sengi/prometheus#diff-06308fdc5bf7909e591ecc888974eba1d8e36394b6a1068b4bbb845e1cb7028dR237) if you need.)

terraform/modules/monitoring/grafana.tf
```diff
@@ -4,7 +4,8 @@ locals {
 
 module "grafana_app" {
   source                        = "../app"
-  image_name                    = "grafana/grafana"
+  image_name                    = "grafana"
+  registry_fqdn                 = "grafana"
```
sengi, line 8: If specifying a short name works (which I presume it does) then perhaps `registry_fqdn` isn't the best name. Maybe just `registry`?

sengi (changes requested):

sengi:
Thanks for sorting this out btw! Will be awesome to have this working.

concourse/pipelines/deploy.yml
fredericfran-gds: will fix

concourse/pipelines/deploy.yml
fredericfran-gds: will fix

concourse/pipelines/deploy.yml
fredericfran-gds: will fix, need secret fix too but easy to do

terraform/deployments/ecr/main.tf
fredericfran-gds: will fix

terraform/deployments/ecr/main.tf
fredericfran-gds: will fix

terraform/modules/monitoring/grafana.tf
fredericfran-gds: yeah, a short name works. I can fix that name to `registry`. I forgot about the Grafana part until the last minute.

terraform/deployments/ecr/main.tf
fredericfran-gds: done in ecr and concourse-iam main files

concourse/pipelines/deploy.yml
```diff
@@ -16,6 +16,14 @@ resource_types:
       repository: governmentpaas/s3-resource
       tag: latest # TODO - don't use latest (once we've worked out a policy for third party images)
 
+  - name: registry-image-v2
```
sengi, line 19: Why `registry-image-v2`? It appears to be v1 (1.2.1). If this is to distinguish it from the anchor called `registry-image` below then it'd be much better to rename the anchor (say to `image`) and then use the expected name for this resource so that it's consistent with the others (i.e. named after its repo, minus the `-resource` part).

If you wanna do that in a follow up then that's fine.

sengi (approved):

concourse/pipelines/deploy.yml
fredericfran-gds: I have overridden the default resource type now as discussed in slack. Thanks

sengi (approved):

PR #210, opened 2021-04-16
Adding ability to set desired count for the following services:

nsabri1:
Publisher worker
	Statsd

### DIFF
--- a/terraform/modules/statsd/variables.tf
+++ b/terraform/modules/statsd/variables.tf
@@ -40,3 +40,9 @@ variable "task_role_arn" {
 variable "vpc_id" {
   type = string
 }
+
+variable "desired_count" {
+  description = "Desired count of Application instances"
+  type        = number
+  default     = 1
+}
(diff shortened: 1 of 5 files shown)


richardTowers (approved):
Looks good to me - consistent with the other applications.

terraform/modules/statsd/variables.tf
```diff
@@ -40,3 +40,9 @@ variable "task_role_arn" {
 variable "vpc_id" {
   type = string
 }
+
+variable "desired_count" {
+  description = "Desired count of Application instances"
```
sengi, line 45: I'd recommend omitting the description in this case as it doesn't really add anything.

sengi (approved):
Cool - thanks for fixing

terraform/modules/statsd/variables.tf
nsabri1: sorry missed this comment - will keep that in mind for next

PR #211, opened 2021-04-16
Remove .terraform.lock.hcl files from version control

bilbof:
Terraform 0.14 introduced [HCL lock files](https://www.terraform.io/docs/language/dependency-lock.html).

By default generating a lock file with `terraform init` will create a lock file that is only valid in the platform that you ran terraform init in. There's no way to override this default behaviour as far as I can see.

For example, on macOS running `terraform init` will autogenerate a lock file that will cause `terraform init` to fail on Linux (the platform we use for CI/CD).

Our current workaround is to run `terraform providers lock` and [specify all target platforms](https://www.terraform.io/docs/cli/commands/providers/lock.html#specifying-target-platforms) - we could possibly do this in a pre-commit

Terraform recommend committing the lock files to version control.

However, we're not seeing a benefit to committing the lock files, so we are going to see if we can do without them.


sengi (approved):
Go for it. I don't think these files are adding any value (neither in terms of security nor change management) and they've caused us significant pain already.

PR #212, opened 2021-04-16
fix cron

fredericfran-gds:
fixes #172 where `run-terraform` job was stuck waiting for everyday-at-6am resource to available.

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -223,6 +223,7 @@ resources:
     source:
       expression: "0 06 * * 1-5"
       location: "Europe/London"
+      fire_immediately: true
 
 groups:
   - name: all


concourse/pipelines/deploy.yml
```diff
@@ -223,6 +223,7 @@ resources:
     source:
       expression: "0 06 * * 1-5"
       location: "Europe/London"
+      fire_immediately: true
```
sengi, line 226: Why does `everyday-at-6am` need fire_immediately but not `everyday-at-11pm`?

fredericfran-gds: Screenshot 2021-04-16 at 11 39 47
 This is the issue we have.

fredericfran-gds: it's not a problem for the 11pm one as it is stand-alone task so it is ok to wait 11pm for a version to be created

bilbof (approved):

concourse/pipelines/deploy.yml
richardTowers: Should we mark the input as `optional: true` as well (or instead) do you think? I don't think it makes any difference 🤔.

fredericfran-gds: it doesn't work on its own

PR #213, opened 2021-04-16
fix resources creation attempt before cert creation

fredericfran-gds:
fix situation where resources, which are dependent on certificate(s) being created before, are created beforehand and terraform erroring


sengi (approved):
Cool. Thanks for fixing!

PR #214, opened 2021-04-16
Use team-level slack webhook Concourse variable

bilbof:
I've set `deploy_apps_slack_webhook` as a team-wide variable, so that when we create duplicate pipelines such as deploy-apps-test-bill and deploy-apps-test-roch we won't need to run `gds cd secrets add deploy-apps-test-bill/slack_webhook`.


nsabri1 (approved):
LGTM

PR #215, opened 2021-04-16
Re-enable logging for concourse/run-task

bilbof:
We saw intermittent failures with esc-cli logs, but they seem to have gone away.

Run-task shouldn't fail if fetching logs fails, though, so if this keeps happening perhaps we should trap the error.


sengi (approved):

richardTowers:
Yeah, maybe we should stick `|| true` on the end of the line or something...

PR #216, opened 2021-04-16
Refactor: remove statsd and task-definition modules

bilbof:
We can remove the Statsd and task-definition modules by deploying Statsd in the same way we deploy other apps to ECS.

I haven't added Statsd to the Concourse deployment pipeline yet since we'd need to first fix an issue with the Statsd image, and we unsure whether that'll be necessary given we are moving from Statsd+Graphite to Prometheus.


terraform/deployments/govuk-publishing-platform/app_statsd.tf
```diff
@@ -1,15 +1,33 @@
...
   execution_role_arn               = aws_iam_role.execution.arn
-  internal_app_domain              = var.internal_app_domain
+  extra_security_groups            = [local.govuk_management_access_security_group, aws_security_group.mesh_ecs_service.id]
+  image_name                       = "statsd:test-0.1.3" # TODO: hardcoded image tag
+  log_group                        = local.log_group
+  memory                           = local.statsd_defaults.memory
   mesh_name                        = aws_appmesh_mesh.govuk.id
-  private_subnets                  = local.private_subnets
-  security_groups                  = [aws_security_group.mesh_ecs_service.id, local.govuk_management_access_sg_id]
+  registry                         = "govuk"
```
sengi, line 24: Should this be `var.registry`?

terraform/deployments/govuk-publishing-platform/defaults.tf
```diff
@@ -38,7 +38,7 @@ locals {
     website_root            = local.public_entry_url,
 
     virtual_service_backends = [
-      module.statsd.virtual_service_name
+      module.statsd.virtual_service_names[0]
```
sengi, line 41: Just fyi this is an example of unsafe use of `count`. No action needed here, because there's not much you can do about it until we fix the module, but wanted to raise it anyway since it's a good example of a smell that we should be looking out for in future. This is the card for fixing the issue: https://trello.com/c/iIsJYdHs/227-fix-unsafe-use-of-terraform-count-for-non-identical-objects

sengi (approved):
Cool - great to see the end of the confusing situation with task-definition/task-definition-v2 🎉 Thanks for cleaning this up!

terraform/deployments/govuk-publishing-platform/defaults.tf
bilbof: Thanks. Yes I'd like to remove the container services input - it's great this is already in a card.

terraform/deployments/govuk-publishing-platform/app_statsd.tf
bilbof: In this case it'll need to stay as govuk for now, since the Statsd image is not yet in ECR.

PR #218, opened 2021-04-19
Limit log lines for Concourse job to 5000

bilbof:
It's not possible to view the output for the update-ecs-service
job if the task outputs too many log lines.

For example, during a deploy Router generates 45k log lines
in about 200 seconds, causing the Concourse UI to crash.

https://trello.com/c/L7pANVwe/463-router-over-logging


richardTowers (approved):
👍🏻 

... although we should probably also think about why router ships 45k log lines when it starts up, right?

sengi (approved):
Seems reasonable 👍  I can think of a bunch of potential usability issues (e.g. is the tail more useful than the head? should we combine both? can we indicate in the output that the logs might be truncated?) but in practice we'd be better off just fixing the logspew if this starts truncating stuff often, rather than finessing this. It's nice and predictable at least.

PR #220, opened 2021-04-20
fix destroy_terraform task

fredericfran-gds:
terraform init should happen before the other terraform cmds

Refs: #205 and #214

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -318,9 +318,10 @@ jobs:
 
               cd govuk-infrastructure/terraform/deployments/govuk-publishing-platform
 
+              terraform init -backend-config "role_arn=$ASSUME_ROLE_ARN"
+
               terraform workspace select "$WORKSPACE"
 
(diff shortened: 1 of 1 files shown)


sengi (approved):
I couldn't find anything canonical about this in the TF docs but from what I can tell it indeed sounds like the init is supposed to happen first. You might need to watch out for [this issue](https://support.hashicorp.com/hc/en-us/articles/360043550953-Selecting-a-workspace-when-running-Terraform-in-automation) - might be worth considering setting `TF_WORKSPACE` before running tf init, rather than running `workspace select` as a separate step. Haven't tried any of this so YMMV - go with what works of course, but probably a good idea to check that it works with both new (first-use) and existing workspaces.

fredericfran-gds:
thanks, I never encountered it but good to know if things go pear shape. It seems to work in the `run_terraform` bit: https://github.com/alphagov/govuk-infrastructure/blob/0cf2279307e061945885437a3355c7aad4cef1e4/concourse/pipelines/deploy.yml#L431

PR #223, opened 2021-04-22
allow default bootstrap image tag to be overridden

fredericfran-gds:
This PR allows the default bootstrap image tag `latest` to be overridden. This is useful in the case of statsd where we want to use the tag `test-0.1.3` for now


sengi (approved):
We might wanna remove this once statsd is sorted out (one way other another) - but seems a sensible workaround for now 👍

PR #224, opened 2021-04-22
Push terraform output in parallel

bilbof:
This put can be done in parallel with the others. It should speed up the `run-terraform` job.


fredericfran-gds (approved):
looks good

PR #226, opened 2021-04-22
Refactor app module

bilbof:
This removes the `custom_container_services` and `service_mesh` variables from the `app` module.

These seem to have been required by router and grafana, but are no longer needed.

The app module no longer explicitly supports apps which can serve traffic on multiple ports. If this need arises (for example, apps which have an admin port and cannot easily be modified) I'd recommend extracting the service mesh node module call from the app module, and have the clients of the app module also create the service mesh virtual services.

### DIFF
--- a/terraform/modules/app/main.tf
+++ b/terraform/modules/app/main.tf
@@ -12,10 +12,7 @@ terraform {
 }
 
 locals {
-  container_name = "app"
-  subdomain      = var.service_name
-  # TODO: Do we need container services?
-  container_services      = var.custom_container_services == null ? [{ container_service = local.subdomain, port = 80, protocol = "http" }] : var.custom_container_services
+  container_name          = "app"
   service_security_groups = concat([aws_security_group.service.id], var.extra_security_groups)
 }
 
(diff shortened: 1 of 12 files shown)


terraform/modules/app/main.tf
```diff
@@ -62,16 +60,17 @@ resource "aws_ecs_service" "service" {
...
   backend_virtual_service_names    = var.backend_virtual_service_names
   mesh_name                        = var.mesh_name
-  port                             = local.container_services[count.index].port
-  protocol                         = local.container_services[count.index].protocol
+  port                             = var.port
+  protocol                         = "http"
   service_discovery_namespace_id   = var.service_discovery_namespace_id
   service_discovery_namespace_name = var.service_discovery_namespace_name
-  service_name                     = local.container_services[count.index].container_service
+  service_name                     = local.subdomain
```
fredericfran-gds, line 67: I know it is effectively the same, but I would have chosen, `var.service_name` rather than `local.subdomain`

fredericfran-gds (commented):
looks, good, see my one inline comment. In addition, have you tested that Grafana can still be deployed?

terraform/modules/app/main.tf
bilbof: Good spot, I've changed this to `var.service_name`.

fredericfran-gds (approved):
looks good

PR #227, opened 2021-04-23
Log from ECS to Splunk

karlbaker02:
This PR changes logging for ECS so that logs from both `app` and `envoy` containers move from AWS CloudWatch Logs to Splunk. [Splunk is the recommended tool for storing infrastructure, application and audit logs](https://gds-way.cloudapps.digital/standards/logging.html#how-to-store-and-query-logs).

Splunk has been set-up so that it logs to a new `govuk_replatforming` index, which will need to change as and when we move the new platform into any real environments. As we're using the Splunk Docker driver and aren't changing the logging of individual applications in this commit, we're using the `tag` option in the logging configuration to specify some additional metadata which will be useful for debugging, such as the image name, container name and container ID. The `splunk-sourcetype` option is currently being hard-coded to the value `log`, and the `splunk-format` option is being set to `raw` which enables us to query the values that were specified earlier with the `tag` option from within the Splunk interface.

Finally, the `splunk-url` and `splunk-token` options are set from values stored using AWS Secrets Manager.

Trello: https://trello.com/c/Y4aLvX2n

### DIFF
--- a/terraform/deployments/govuk-publishing-platform/app_content_store.tf
+++ b/terraform/deployments/govuk-publishing-platform/app_content_store.tf
@@ -74,12 +74,15 @@ module "content_store" {
       ROUTER_API_BEARER_TOKEN     = data.aws_secretsmanager_secret.content_store_router_api_bearer_token.arn
     }
   )
-  log_group          = local.log_group
-  aws_region         = data.aws_region.current.name
-  cpu                = local.content_store_defaults.cpu
-  memory             = local.content_store_defaults.memory
-  task_role_arn      = aws_iam_role.task.arn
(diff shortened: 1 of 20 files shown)


fredericfran-gds (dismissed):
looks good to me, might have been good to include some docs about how to use this new functionality.

karlbaker02:
> looks good to me, might have been good to include some docs about how to use this new functionality.

Thanks Fred. I agree we need some documentation, though I think this should be a separate card and one we do later in case we decide to make improvements in that time to the default logging we're doing in this PR.

sengi (approved):
Re-approving since the rebase invalidated the previous approval. (Only had a quick skim-through because it's already been reviewed.)

PR #233, opened 2021-04-30
adding/updating security groups for publisher and publishing-api

fredericfran-gds:
- workers of publisher and publishing-api needs access to redis as they use sidekiq.

- adding egress rule for publisher worker and publishing-api

### DIFF
--- a/terraform/deployments/govuk-publishing-platform/security_group_rules.tf
+++ b/terraform/deployments/govuk-publishing-platform/security_group_rules.tf
@@ -251,8 +251,8 @@ resource "aws_security_group_rule" "postgres_from_publishing_api_5432" {
 # Publisher
 #
 
-resource "aws_security_group_rule" "publisher_to_any_any" {
-  description       = "Publisher sends requests to anywhere over any protocol"
+resource "aws_security_group_rule" "publisher_web_to_any_any" {
+  description       = "Publisher web sends requests to anywhere over any protocol"
   type              = "egress"
   from_port         = 0
   to_port           = 0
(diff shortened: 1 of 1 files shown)


terraform/deployments/govuk-publishing-platform/security_group_rules.tf
```diff
@@ -261,10 +261,40 @@ resource "aws_security_group_rule" "publisher_to_any_any" {
...
+  protocol          = -1
+  cidr_blocks       = ["0.0.0.0/0"]
+  security_group_id = module.publisher_worker.security_group_id
+}
+
 #
 # Publishing API
 #
 
+resource "aws_security_group_rule" "publishing_web_to_any_any" {
```
karlbaker02, line 278: This is for Publishing API, can we rename it something like `publishing_api_web_to_any_any`?

terraform/deployments/govuk-publishing-platform/security_group_rules.tf
```diff
@@ -261,10 +261,40 @@ resource "aws_security_group_rule" "publisher_to_any_any" {
...
+  description       = "Publishing web sends requests to anywhere over any protocol"
+  type              = "egress"
+  from_port         = 0
+  to_port           = 0
+  protocol          = -1
+  cidr_blocks       = ["0.0.0.0/0"]
+  security_group_id = module.publishing_api_web.security_group_id
+}
+
+resource "aws_security_group_rule" "publishing_worker_to_any_any" {
```
karlbaker02, line 288: Similarly this is for Publishing API, can we rename it something like `publishing_api_worker_to_any_any`?

fredericfran-gds: yes.

terraform/deployments/govuk-publishing-platform/security_group_rules.tf
fredericfran-gds: yes!

karlbaker02 (approved):
Looks good Fred! 💯

PR #235, opened 2021-05-04
Use Envoy's /ready endpoint for healthcheck

bilbof:
This path returns a 200 when Envoy is in a LIVE state,
returns a 503 status otherwise.

This will fix some transient issues we have experienced
where Envoy was not ready to proxy a request.

https://www.envoyproxy.io/docs/envoy/latest/operations/admin#get--ready


sengi (approved):
Oof, well spotted!

PR #236, opened 2021-05-04
Create Signon bearer tokens during first deployment

bilbof:
Note: I would recommend reviewing by commit.

This introduces a `bootstrap-secrets` task to app deployment jobs in Concourse, which creates the initial versions of the Signon bearer tokens an app needs. Rotation of the tokens is subsequently handled by the SecretsManager rotation lambda.

This change is needed since AWS SecretManager triggers the Lambda rotation functions (which would create the initial bearer token version), but the Lambda retries can end before Signon is started (a race condition).

The solution is to continue rotating tokens asynchronously, but create the initial secret versions synchronously during the first deployment of the apps. This ensures that when we deploy an app that requires signon-generated secrets, the secrets have already been created and pushed to SecretsManager.

Additionally I've included some improvements to the lambda rotation functions, and two quality of life improvements (ruby linting and a Rakefile (Ruby's make utility). I hope this will encourage us to use Ruby rather than Bash for future Concourse tasks, since (for me) they are substantially easier to test, and will be more approachable to GOV.UK developers.


terraform/deployments/govuk-publishing-platform/signon_secrets.tf
```diff
@@ -66,109 +32,127 @@ locals {
     router_api          = "Router API"
   }
   signon_api_user = {
-    content_store       = join("-", compact([local.api_user_prefix, "content-store@alphagov.co.uk"]))
-    draft_content_store = join("-", compact([local.api_user_prefix, "draft-content-store@alphagov.co.uk"]))
-    frontend            = join("-", compact([local.api_user_prefix, "frontend@alphagov.co.uk"]))
-    publisher           = join("-", compact([local.api_user_prefix, "publisher@alphagov.co.uk"]))
-    publishing_api      = join("-", compact([local.api_user_prefix, "publishing-api@alphagov.co.uk"]))
+    content_store       = join("-", compact([local.api_user_prefix, "content-store@digital.cabinet-office.gov.uk"]))
```
sengi, line 35: Are we sure we want these to share a namespace with staff usernames and anything else that might need a Google Workspace account (including in the future)? Bear in mind that GDS IT owns this namespace, not GOV.UK, so we're not necessarily in control of future things which might want an @digital.cabinet-office.gov.uk address. (Sorry for commenting late - hadn't seen this in the earlier PR where it was first introduced.)

bilbof: Sure, these can be any unique email address. Do you think `<workspace>.<env>.govuk.digital` would be better?

sengi: Yeah, that sounds like a better bet to me. We (GOV.UK) own the domain and we don't run email on it so much less potential for clashes or confusion 👍

bilbof: Great! @sengi I've pushed this corresponding fix up, lmk if you don't feel comfortable reviewing as it's a Ruby commit https://github.com/alphagov/signon/pull/1631

spec/lib/signon/bootstrap_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+      receive(:describe_secret)
+      .with(secret_id: arn)
+      .and_return(metadata)
+
+    stub_get_secret
+      .with(secret_id: password, version_stage: "AWSCURRENT")
+      .and_return(admin_password)
+  end
+
+  context "when when secret rotation is disabled" do
```
bilbof, line 58: TODO

spec/lib/signon/bootstrap_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    let(:metadata) { double(rotation_enabled: false) }
+
+    it "raises an error" do
+      expect { call }.to raise_error Signon::Bootstrap::NotRotatable
+    end
+  end
+
+  context "when secret has already been created" do
+    it "does not create a new token" do
+      expect { call }.not_to raise_error
```
bilbof, line 68: TODO: Assert that we don't create the secret, or make other calls to signon.

sengi (approved):
LGTM - ran through in person - thanks for explaining the Ruby stuff to me!

I'll leave you to fix the couple of trivial bits we found along the way.

PR #237, opened 2021-05-04
Link to Splunk for logs when updating ECS service

karlbaker02:
This PR links to Splunk for logs when updating an ECS service. This change is necessary as we're no longer logging to AWS CloudWatch Logs, so the call to `ecs-cli logs` is failing. As we have the `runtimeId` of ECS containers logged to Splunk (via the `container_id` property) we can instead track ECS updates there.

### DIFF
--- a/concourse/tasks/update-ecs-service.sh
+++ b/concourse/tasks/update-ecs-service.sh
@@ -30,17 +30,34 @@ aws ecs update-service \
   --task-definition "$new_task_definition_arn" \
   --region "$AWS_REGION" > /dev/null
 
-# Get the ID of the latest task so we can stream its logs.
-# ecs-cli just wants the numeric ID of the task, not the whole ARN.
-# So we have to use `awk` to fish that out.
-task_id=$(aws ecs list-tasks \
+# Get the ID of the latest task.
+task_arn=$(aws ecs list-tasks \
   --cluster "$CLUSTER" \
   --service-name "$ECS_SERVICE" \
   --region "$AWS_REGION" \
   --query 'taskArns | [0]' \
-  --output text \
-  | awk -F / '{print $NF}'
+  --output text
 )
-echo "Task ID: $task_id"
+
+echo "Task ARN: $task_arn"
+
+# Pull the container ID from the task's associated `app` container
+container_id=$(aws ecs describe-tasks \
+  --cluster "$CLUSTER" \
+  --tasks $task_arn \
+  --query 'tasks[0] | containers[?name==`app`].runtimeId' \
+  --output text
+)
+
+echo "App container ID: $container_id"
+
+echo "Waiting for $ECS_SERVICE ECS service to reach steady state..."
+
+echo "Check Splunk for logs: https://gds.splunkcloud.com/en-GB/app/gds-006-govuk/search?q=search%20index%3D%22govuk_replatforming%22%20container_id%3D$container_id"
+
+aws ecs wait services-stable \
+  --cluster "$CLUSTER" \
+  --services "$ECS_SERVICE" \
+  --region "$AWS_REGION"
 
 echo "Updated $ECS_SERVICE to task definition $new_task_definition_arn."
--- a/concourse/tasks/run-task.sh
+++ b/concourse/tasks/run-task.sh
@@ -62,7 +62,8 @@ aws ecs wait tasks-stopped --tasks $task_id --cluster $CLUSTER
 echo "task finished."
 task_results=$(aws ecs describe-tasks --tasks $task_id --cluster $CLUSTER)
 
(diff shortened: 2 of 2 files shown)


concourse/tasks/update-ecs-service.sh
```diff
@@ -30,22 +30,30 @@ aws ecs update-service \
...
   --service-name "$ECS_SERVICE" \
   --region "$AWS_REGION" \
   --query 'taskArns | [0]' \
-  --output text \
-  | awk -F / '{print $NF}'
+  --output text
 )
-echo "Task ID: $task_id"
+
+echo "Task ARN: $task_arn"
```
sengi, line 42: This log message could be a bit confusing/strange as there would normally be more than one task. This prints the Task ARN of one of the tasks.

bilbof (approved):
LGTM thanks for this!

Could we do the same for `concourse/tasks/run-task.sh` (and other places where we are using `ecs-cli logs`)?

karlbaker02:
> LGTM thanks for this!
> 
> Could we do the same for `concourse/tasks/run-task.sh` (and other places where we are using `ecs-cli logs`)?

Absolutely, apologies for missing those. Will add them to this PR.

concourse/tasks/update-ecs-service.sh
```diff
@@ -30,22 +30,30 @@ aws ecs update-service \
...
+
+# Pull the container ID from the task's associated `app` container
+container_id=$(aws ecs describe-tasks \
+  --cluster "$CLUSTER" \
+  --tasks $task_arn \
+  --query 'tasks[0] | containers[?name==`app`].runtimeId' \
+  --output text
+)
+
+echo "App container ID: $container_id"
```
sengi, line 52: Does the container ID of (say) the app container stay consistent between all task replicas within a service (for a given task definition version)? If it does then great - but it does seem a little surprising to me.

karlbaker02: My ECS knowledge doesn't yet go that far, so I will have to get back to you on that.

bilbof: I'm fairly sure the container ID isn't consistent between tasks. Only the name of the container is consistent. This should be fine provided the `task_arn` belongs to the new task version, but as @sengi wrote in his other comment we should see if we can ensure we get the new task version each time.

Another issue is that some ECS Services (only frontend at the moment!) have multiple tasks, therefore we will have multiple container IDs, but we don't need to solve that now.

concourse/tasks/update-ecs-service.sh
karlbaker02: So the previous definition of _update-csv-service.sh_ was only concerned with a single task, and when I tested this against my cluster it did not occur to me that we would have multiple tasks per service. In the old way of doing things, we would only be monitoring a single task via CloudWatch Logs during the deploy; in the case of multiple tasks, am I assuming that we want to monitor each?

concourse/tasks/update-ecs-service.sh
karlbaker02: In terms of getting the new task version each time, will it not be the case that whenever we push changes and a new task is created, the service would not be spun up against that new and latest task definition? If this does happen then there is no need to check the task definition version as this will be a given?

karlbaker02 (commented):
Apologies for the delay in replying to this, didn't get a chance during 2nd line.

concourse/tasks/update-ecs-service.sh
bilbof: Sorry, I shouldn't have used 'version', it's confusing.

Perhaps in the `update-ecs-service` task we can simply provide a link to the service-level logs, rather than container-level. Something like this:

```shell
echo "Check Splunk for logs: https://gds.splunkcloud.com/en-GB/app/gds-006-govuk/search?q=search%20index%3D%22govuk_replatforming%22%20GOVUK_APP_NAME%3D$ECS_SERVICE"
```

For the `run-task` job we would probably want a link to view the logs for a particular app container. So to get the app container ID. Luckily, the AWS CLI will return the Task ID when you call `aws ecs run-task`, so you can just use `describe-tasks` to get the Container ID.

My reasoning is:

ECS does a [rolling update deployment](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/deployment-type-ecs.html) when we update an ECS Service's Task Definition to a new _Task Definition revision_.

During the deploy, the ECS Service will be running Tasks that use the old and new Task Definition revisions side by side (the link above has some detail).

As a consequence of this, if you were to call `aws ecs list-tasks --service-name frontend` during a deploy you would get a list of four task definitions (2 old, 2 new), but you'd only want to see the logs of the new tasks. Annoyingly there's no way to filter by task definition 😅

PR #238, opened 2021-05-05
use for_each to create different ECR repositories

fredericfran-gds:
Since the different ECR repos have the same config, we use `for_each` terraform functionality to make the code more succint.


sengi (approved):
Thanks - this is a nice improvement!

PR #239, opened 2021-05-05
Wait 15minutes for ECS deploy to complete

bilbof:
Motivation:

This should make deploy jobs fail less frequently due to the following error:

```
Waiter ServicesStable failed: Max attempts exceeded
```

I would expect this job to succeed after this change: https://cd.gds-reliability.engineering/teams/govuk-test/pipelines/deploy-apps/jobs/deploy-publisher/builds/6.

Implementation:

This change will make the deploy job wait for 20mins for an ECS deployment to complete. This feels excessive, but as seen in our deploy job logs, it's unfortunately necessary.

If we find this is too long to wait for success, we can now configure the timeout/attempts per deployment with this change.

Waiting a longer time for a deploy to complete could mask a bigger problem. For example, the ECS error `CouldNotPullContainerError` can occur when ECS is unable to pull an image from a registry. After this change we would not discover this problem for an extra 10 minutes. I think this is acceptable, since the change won't cause additional problems or make them worse, and it will fix an existing problem.

Explanation:

An ECS service is stable when `service.deployments == 0 and runningCount == desiredCount` ([source](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/ecs/wait/services-stable.html)). In other words, when
there are no active deployments associated with the service and the desired number of tasks are running.

The [`ecs wait service-stable` command](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/ecs/wait/services-stable.html) is used to wait for a deployment to complete.

However, the polling time for the wait command is not configurable: it polls every 15s for 40 attempts (10m).

A deploy can take longer than this. For example the publisher-web deployment this morning was started at 10:01:37 and completed at 10:16:26.

This change allows us to configure the timeout on a deploy.

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -560,6 +560,20 @@ jobs:
         params:
          ECS_SERVICE: frontend
          WORKSPACE: ((workspace))
+    - in_parallel:
+      - &await-deploy-complete
+        task: await-deploy-complete-live
+        file: govuk-infrastructure/concourse/tasks/await-deploy-complete.yml
+        attempts: 4
+        timeout: 5m
+        params: &await-deploy-complete-params
+          ECS_SERVICE: frontend
+          WORKSPACE: ((workspace))
+      - <<: *await-deploy-complete
+        task: await-deploy-complete-draft
+        params:
+          <<: *await-deploy-complete-params
+          ECS_SERVICE: draft-frontend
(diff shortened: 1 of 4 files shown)


sengi (approved):
SGTM. Thanks for fixing this robustness issue. I think it's no bad thing for the await step to be its own task anyway - seems like a clearer separation of concerns.

PR #243, opened 2021-05-05
use new default `main` GitHub branch to build govuk apps

fredericfran-gds:
Some govuk apps have switched to the new default `main` GitHub branch so we modify the relevant concourse resources for these apps:
1. publisher
1. router-api
1. signon
1. static


bilbof (approved):
Thanks for fixing this up. I wonder if we should change it explicitly to `branch: main` to avoid confusion?

PR #245, opened 2021-05-05
Relax container health check

bilbof:
We're seeing new containers get stopped due to
failed health checks. We think could be because
the app web servers are slow to start.

This gives a grace period of 30 seconds to an
app when it starts up, before a failed health
check will count towards the threshold at which
the container will be stopped.

This also increases the number of times we will
retry a failed health check from 3 to 5.


fredericfran-gds (approved):
looks good

PR #247, opened 2021-05-06
Allow Concourse readonly role to read the test Signon secret.

sengi:
It needs this in order to run `terraform plan` unfortunately.

We'll probably want to revisit this, but in the meantime it at least gets the [test](https://github.com/alphagov/govuk-infrastructure/blob/0b45938590bae8a1254898f91a224ce024e68478/concourse/tasks/terraform-plan-govuk-deployment.yml) working again, which was previously causing a failed pre-merge check on every PR.

Tested: [terraform-plan-govuk-deployment test passes](https://cd.gds-reliability.engineering/teams/govuk-ci/pipelines/ci-govuk-infrastructure-main/jobs/terraform-plan-govuk-deployment)

[Trello card](https://trello.com/c/fnrn6FGA/495-fix-broken-test-terraform-plan-govuk-deployment-failing-pre-merge-check)

### DIFF
--- a/terraform/deployments/concourse-iam/main.tf
+++ b/terraform/deployments/concourse-iam/main.tf
@@ -75,6 +75,27 @@ resource "aws_iam_role_policy_attachment" "concourse_readonly" {
   policy_arn = "arn:aws:iam::aws:policy/ReadOnlyAccess"
 }
 
+resource "aws_iam_role_policy" "concourse_terraform_planner_test" {
+  # HACK: In order to run terraform plan (as a pre-merge check), the read-only
+  # Concourse role needs to be able to read certain secrets in the test
+  # account. This is intended for the test account only.
+  # TODO: Consider whether we can run pre-merge checks with the deployer role,
+  # and what controls we should have in place re secrets in the test account.
+  name = "concourse_terraform_planner_test"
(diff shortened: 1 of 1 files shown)


terraform/deployments/concourse-iam/main.tf
```diff
@@ -75,6 +75,27 @@ resource "aws_iam_role_policy_attachment" "concourse_readonly" {
...
+  name = "concourse_terraform_planner_test"
+  role = aws_iam_role.govuk_concourse_terraform_planner.id
+  policy = jsonencode({
+    "Version" : "2012-10-17",
+    "Statement" : [
+      {
+        "Sid" : "TerraformPlannerReadsNonSensitiveTestSecrets",
+        "Effect" : "Allow",
+        "Action" : "secretsmanager:GetSecretValue",
+        "Resource" : "arn:aws:secretsmanager:eu-west-1:430354129336:secret:signon_admin_password_ecs-*"
```
bilbof, line 93: Is it only the `signon_admin_password`? It looks like the tests are now passing but that surprised me!

bilbof (approved):
Thank you for this!

terraform/deployments/concourse-iam/main.tf
sengi: Seems so 🤷

PR #248, opened 2021-05-06
Make concourse-iam environment-agnostic

bilbof:
This will enable the concourse team govuk-integration to deploy apps to integration.

As [in RE Autom8's docs](https://reliability-engineering.cloudapps.digital/continuous-deployment.html#services):

> Each Concourse team’s worker instances have a specific AWS IAM role, this allows Concourse to assume roles in other AWS accounts, if those accounts have a role which Concourse is permitted to assume.

Since we now have a Concourse team `govuk-<env>` for each environment, we need to trust a team-specific Concourse worker IAM role.


fredericfran-gds (approved):
looks good

PR #249, opened 2021-05-06
Disable slack workspace notifications

bilbof:
These are a little noisy. Workspaces shouldn't be expected to always be functional, so this lets one turn off slack messages for a workspace pipeline.

I've also enabled alerts for run-terraform failures.


sengi (approved):
SGTM

PR #250, opened 2021-05-06
Allow enabling or disabling of DB Migrations

smford:
This allows DB migrations to be enabled or disabled per workspace.

We do not want DB migrations to occur in the workspaces we are currently
working upon.  This PR create a new variable, specific to each
workspace, that toggles DB migrations on or off.

A default deploy-parameters.yml file has been defined for use against
the main/default workspace deployment.  Each workspace has its own
subdirectory under govuk-infrastructure/concourse/pipelines which will
now hold the workspaces specific deploy-parameters.yml file

Example Workspace Usage:
```
cd govuk-infrastructure/concourse/pipelines
fly set-pipeline -t cd-govuk -c ./deploy.yml -l ./steve/deploy-parameters.yml -p deploy-apps-steve-test
```


concourse/pipelines/default/deploy-parameters.yml
```diff
@@ -1,2 +1,3 @@
 workspace: default
 govuk_infrastructure_branch: main
+skip_db_migrations: true
```
bilbof, line 3: I would recommend putting this variable in a separate file. Since we want db migrations to run in the test environment, but not in the integration environment.

Then in the `set_pipeline` step I think you can specify this file:

```yaml
      var_files:
        - govuk-infrastructure/concourse/pipelines/deploy-parameters.yml
        - govuk-infrastructure/concourse/parameters/((govuk_environment))/parameters.yml
```

concourse/tasks/run-task.sh
```diff
@@ -22,6 +22,12 @@ root_dir=$(pwd)
 : "${COMMAND:?COMMAND not set}"
 : "${CLUSTER:?COMMAND not set}"
 : "${VARIANT:?VARIANT not set}"
+: "${SKIP_DB_MIGRATIONS:?SKIP_DB_MIGRATIONS not set}"
+
+if [[ "$SKIP_DB_MIGRATIONS" == "true" ]]; then
+  echo "Skipping DB Migrations"
+  exit 0
+fi
```
bilbof, line 30: The `run-task` job is used by other tasks (e.g. smoke tests, bootstrapping secrets, etc), so we'd need this flag to be checked a bit higher up

e.g. you could create a `run_db_migration.yml` file (a copy of the `run_task.yml` file), and have that first check the flag, then run this file.

concourse/pipelines/default/deploy-parameters.yml
smford: So changing from being workspace specific, to being environment specific?

smford: I like this idea!

smford:
This PR has been superseded by: https://github.com/alphagov/govuk-infrastructure/pull/278 and will be closed.  The new PR allows for per-environment configuration

PR #253, opened 2021-05-07
Enable more smoke tests (smokey loop & more apps)

bilbof:
This will run smoke tests after deploys of frontend, publisher, and signon.

We don't yet have smoke test scenarios that will pass for the other apps we've got running so far (authenticating proxy, static, content-store, publishing-api, router, and router-api). For some tests we can publish some content to get them passing (e.g. tests that check a user can reach pages like /robots.txt). Other tests will need rewriting.

This also adds the 'smokey loop' which will run every 5 minutes.

Note: I manually added Smokey's signon user to Secrets Manager.

This should be merged after https://github.com/alphagov/smokey/pull/818


concourse/pipelines/deploy.yml
```diff
@@ -241,6 +241,13 @@ resources:
       location: "Europe/London"
       fire_immediately: true
 
+  - name: every-5-minutes
+    type: cron-resource
+    source:
+      expression: "*/5 * * * 1-5"
```
sengi, line 247: What will Concourse do if the tests start taking longer than 5 minutes?

sengi (approved):
Yay, tests!

concourse/pipelines/deploy.yml
bilbof: It depends on the job [`serial` value](https://concourse-ci.org/jobs.html#schema.job.serial):

> Default false. If set to true, builds will queue up and execute one-by-one, rather than executing in parallel.

In this case, the `all-smoke-tests` job will run smoke tests in parallel if they start taking longer. This is probably fine for now - if we see we start having build ups of tests, we could increase this to every 10mins (and so on).

PR #254, opened 2021-05-07
Fix grafana

fredericfran-gds:
This PR fixes the Grafana deployment that has been broken a while now.
This is because the deployment uses many common modules with
`govuk-publishing-platform` and some changes were not backported to this
deployment.

This deployment uses only the `public-load-balancer` child module
now and to prevent further breakages, we add a GitHub integration
check for this deployment also.

**Testing**
URL of grafana is: https://grafana.ecs.test.govuk.digital
Grafana Concourse URL: https://cd.gds-reliability.engineering/teams/govuk-test/pipelines/monitoring

Ref:
1. [Trello task](https://trello.com/c/twpFac1c/499-fix-grafana-deploy)


sengi (approved):
LGTM. There's a lot of duplicated code here but we can iterate on that later. Thanks for getting this going again.

fredericfran-gds:
yes, the code is duplicated because @richardTowers wanted no dependency between the 2 (monitoring and govuk-platform). I think we should re-visit having Grafana in another cluster. Let's wait for him to come back.

sengi:
Yeah that was mostly about being able to do the spike independently rather than any architectural consideration. We're fine to change that.

PR #255, opened 2021-05-10
Ignore pending rotations when bootstrapping signon tokens

bilbof:
This resolves a race condition which would prevent a SecretsManager secret value being set for a signon bearer token during bootstrapping. A lambda rotation can start before the secret has been set during bootstrapping.

If a secret rotation has started and failed (i.e. the lambda will try again tomorrow) we still want to set a current version of the secret during bootstrapping, regardless of any pending rotations.

Between the bootstrap check for an AWSCURRENT version and the bootstrap set command, the rotation could complete. This would not cause an error, since we need either of the working secrets to be set for the initial run of the pipeline. It doesn't matter which token gets set.

### DIFF
--- a/lib/signon/bootstrap.rb
+++ b/lib/signon/bootstrap.rb
@@ -40,7 +40,8 @@ module Signon
       end
 
       versions = metadata.version_ids_to_stages
-      if versions.any?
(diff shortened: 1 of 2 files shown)


karlbaker02 (approved):
Looks good 👍

PR #257, opened 2021-05-10
Enable Concourse to create bearer tokens during bootstrapping

bilbof:
This fixes two bugs that slipped through in #236:

- Adds scheme (https) to Signon's API URL
- Allow Concourse CIDRs to send requests to Signon's API


fredericfran-gds (approved):
looks good

PR #260, opened 2021-05-11
Bootstrap OAuth applications during Signon deploy

bilbof:
**Motivation**

Currently Signon OAuth applications are created manually using a Signon rake task [`applications:create`](https://github.com/alphagov/signon/blob/138dc124b77f7cf2f6c1431069e049009af0848c/lib/tasks/applications.rake).

This manual process prevents us from spinning up a new cluster automatically.

This automates that manual process to enable us to bring up a new environment without manual intervention.

Trello: https://trello.com/c/elhnwnaf/454-bootstrapping-signon-oauth-app-secrets

**Implementation**

- After Signon is deployed, Concourse will create OAuth applications automatically (if they don't already exist) using a [new Signon API for applications](https://github.com/alphagov/signon/pull/1638).
- Concourse will then put the OAuth ID and Secret into SecretsManager, for use by apps in ECS.
- During bootstrapping, apps that require the env vars `GDS_SSO_OAUTH_ID` or `GDS_SSO_OAUTH_SECRET` will not be deployed until after the required secrets have been generated.

**Further work**

It would be great to rotate app OAuth secrets. This is not currently supported by Doorkeeper (the gem Signon uses for OAuth), so we'd need to implement something in Signon to support this.


spec/lib/signon/bootstrap_spec.rb
```diff
@@ -32,80 +10,228 @@
...
+      end
+    end
+
+    context "when application has been created" do
+      context "when credentials are in secretsmanager" do
+        let(:versions) do
+          { "version1" => %w[AWSCURRENT] }
+        end
+
+        it "will verify the credentials have been set in SecretsManager" do
```
sengi, line 178: "and not create a new application or secret"? (are there length limits on these things?)

spec/lib/signon/client_spec.rb
```diff
@@ -79,8 +83,80 @@
...
+
+      it "does not raise an error" do
+        stub_req(endpoint, method: :get)
+          .to_return(status: 200, body: JSON.generate(res))
+        expect { response }.not_to raise_error
+        expect(response).to eq res
+      end
+    end
+
+    context "when signon request fails" do
```
sengi, line 150: "when an application isn't found"? (the description is currently the same as the 400 case above)

terraform/deployments/govuk-publishing-platform/secrets_manager.tf
```diff
@@ -21,12 +21,6 @@ data "aws_secretsmanager_secret" "frontend_secret_key_base" {
 
 # Content store app secrets
 
-data "aws_secretsmanager_secret" "content_store_oauth_id" {
```
sengi, line 24: 🎉

terraform/deployments/govuk-publishing-platform/signon_bearer_tokens.tf
```diff
@@ -45,7 +36,7 @@ module "publisher_to_publishing_api_bearer_token" {
 
   additional_tags                 = local.additional_tags
   api_user_email                  = local.signon_api_user.publisher
-  app_name                        = local.signon_app.publishing_api
+  app_name                        = local.signon_app.publishing_api.name
```
sengi, line 39: Would it make sense to generate all these module instantiations with module for_each?

terraform/deployments/govuk-publishing-platform/signon_oauth_apps.tf
```diff
@@ -0,0 +1,91 @@
+locals {
+  oauth_applications = [
+    module.oauth_applications["authenticating_proxy"].app_data,
+    module.oauth_applications["content_store"].app_data,
+    module.oauth_applications["draft_content_store"].app_data,
+    module.oauth_applications["publisher"].app_data,
+    module.oauth_applications["publishing_api"].app_data,
+    module.oauth_applications["router_api"].app_data,
+    module.oauth_applications["draft_router_api"].app_data,
+  ]
```
sengi, line 10: Is it feasible to generate this with a [for expression](https://www.terraform.io/docs/language/expressions/for.html)?

terraform/modules/signon_oauth_application/variables.tf
```diff
@@ -0,0 +1,38 @@
...
+
+variable "permissions" {
+  type = list(string)
+}
+
+variable "redirect_path" {
+  type = string
+}
+
+variable "workspace" {
```
sengi, line 36: Is there a reason we're passing workspace around as a var? Is it not the same as terraform.workspace?

sengi (commented):
All looks good except I don't quite get the need for two separate Secrets Manager secrets for ID and password (oauth_secret) - aren't they a pair that we'd want to manage as a single secret? I might be missing something here.

lib/signon/bootstrap.rb
```diff
@@ -6,34 +6,80 @@ module Signon
...
+        return
+      end
+
+      logger.info "Secret (#{app_name}) doesn't have secret values in SecretsManager. Creating them..."
+      secrets = find_or_create_app(signon, app)
+
+      logger.info "Application #{app_name} created or fetched. Putting creds in SecretsManager..."
+      secretsmanager.put_secret_value(
+        secret_id: oauth_id_arn,
+        secret_string: secrets.dig("oauth_id"),
```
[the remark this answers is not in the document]
bilbof: I believe we'd only want to rotate the Secret (the ID can stay the same, it's [not a secret](https://www.oauth.com/oauth2-servers/client-registration/client-id-secret/) - since it's used in URL queries etc).

For rotation we would probably want an app to have multiple secrets, but that wouldn't prevent us combining these into one secet.

gds-sso (the gem we use for OAuth between apps) [expects two environment variables](https://github.com/alphagov/gds-sso/blob/master/lib/gds-sso/config.rb#L13-L19). These could potentially be combined into one secret - but it's probably out of scope for this particular task - I've added https://trello.com/c/uBqmrWBN/312-combine-gdsssooauthid-and-gdsssooauthsecret-into-one-secret to follow up, if that's ok?

lib/signon/client.rb
```diff
@@ -76,5 +104,9 @@ def do_request(request, uri)
         http.request(request)
       end
     end
+
+    def already_exists(res)
+      res.code == "400" && JSON.parse(res.body).dig("error") == "Record already exists"
```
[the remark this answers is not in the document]
bilbof: This is provided by Signon here: https://github.com/alphagov/signon/blob/7bab43419a1e9340273e8fb2e69d2de8b72da2f4/app/helpers/admin_api_helper.rb#L27. It's only used by this client so we can easily change it.

Would a response like `{ "error": "ApplicationAlreadyExistsError"}` feel more robust rather than `{"error": "Record already exists"}`?

If not through parsing the error message, perhaps we can use a `409 Conflict` status code instead. `400` is super broad, which is the reason one must parse the error message. What do you think?

terraform/deployments/govuk-publishing-platform/signon_oauth_apps.tf
bilbof: Ah thank you so much! I thought I was missing something here.

terraform/deployments/govuk-publishing-platform/signon_bearer_tokens.tf
bilbof: Yes, good idea - if it's okay I've stuck in a TODO to follow up and do this in a later commit.

terraform/modules/signon_oauth_application/variables.tf
bilbof: `local.workspace` replaces ["default" with "ecs"](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/govuk-publishing-platform/defaults.tf#L2).

lib/signon/bootstrap.rb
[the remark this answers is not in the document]
sengi: Sure no probs - no need to worry about it in this PR. Sounds like we should ultimately be storing them as one secret with `{oauth_id: foo, oauth_secret: bar}` though.

terraform/modules/signon_oauth_application/variables.tf
sengi: Ah ok, thanks.

sengi (approved):

lib/signon/client.rb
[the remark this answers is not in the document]
sengi: Ah, that's good news then. I reckon either 409 or a more stable-looking message would be an improvement.

lib/signon/bootstrap.rb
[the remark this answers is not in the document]
sengi: I'm assuming that if an app has multiple secrets then each of those would have a separate oauth_id? If that's not the case then I've been talking nonsense all along and you should ignore me 😅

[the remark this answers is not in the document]
bilbof: I think whether an app has multiple IDs will be up to us - we'll need to implement it from scratch, since the [Doorkeeper gem](https://github.com/doorkeeper-gem/doorkeeper) we use for OAuth doesn't support rotation of creds :(

lib/signon/client.rb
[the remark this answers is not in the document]
bilbof: Cool, I'll modify the API and add that in a separate commit.

PR #261, opened 2021-05-11
Retry Concourse terraform destroy task

bilbof:
This task is a bit flaky. Lets see if we can bodge fix it for now with 3 attempts.

Trello: https://trello.com/c/ywDfsw86/467-fix-issues-with-tf-destroy-failing


sengi (approved):
It's sad but SGTM. Terraform (particularly with the AWS provider and its interaction with AWS's APIs and their limitations) is inherently flaky.

PR #262, opened 2021-05-12
Build images serially

bilbof:
See https://concourse-ci.org/jobs.html#schema.job.serial

This ensures that we don't try to build two images for the
same repo concurrently. This can happen if folks merge commits
around the same time.

This can cause a race condition where both builds want to
use the same release tag, which will raise an error, and
the second build will need to be repeated.


fredericfran-gds (approved):
good find, thanks for fixing

PR #263, opened 2021-05-12
Use main branch for Publishing API

bilbof:
Publishing API now uses the main branch for its default branch.


fredericfran-gds (approved):
looks good

PR #264, opened 2021-05-12
fix grafana dns conflict

fredericfran-gds:
**1. fix DNS conflict**

Previously in #254, we imported the DNS zone `ecs.test.govuk.digital`
from `govuk-publishing-platform` to then add the `grafana` CNAME DNS
record. This creates a conflict when `govuk-publishing-platform` is
destroyed nightly before `monitoring`

As a workaround, we create the `grafana` domain in a dns zone
`monitoring.test.govuk.digital`.

So grafana url will be: https://grafana.monitoring.test.govuk.digital

**2. Fix Concourse plan role access to grafana_password**

**3. Fix container healthcheck and image tag**

**4. Fix branch of monitoring pipeline**

### DIFF
--- a/concourse/pipelines/monitoring.yml
+++ b/concourse/pipelines/monitoring.yml
@@ -5,7 +5,7 @@ resources:
   - icon: github
     name: govuk-infrastructure
     source:
-      branch: fix_grafana
+      branch: fix_grafana_zone_conflict
       uri: https://github.com/alphagov/govuk-infrastructure
     type: git
 
--- a/concourse/pipelines/monitoring.yml
+++ b/concourse/pipelines/monitoring.yml
@@ -5,7 +5,7 @@ resources:
   - icon: github
     name: govuk-infrastructure
     source:
-      branch: fix_grafana_zone_conflict
+      branch: main
       uri: https://github.com/alphagov/govuk-infrastructure
     type: git
 
--- a/terraform/modules/monitoring/certs.tf
+++ b/terraform/modules/monitoring/certs.tf
@@ -1,9 +1,23 @@
-resource "aws_route53_zone" "public" {
-  name = local.workspace_external_domain
+data "aws_route53_zone" "public" {
+  name = var.external_app_domain
 }
 
-resource "aws_acm_certificate" "public" {
-  domain_name = "*.${local.workspace_external_domain}"
+resource "aws_route53_record" "monitoring_public_zone_ns" {
+  zone_id = data.aws_route53_zone.public.zone_id
+  name    = local.monitoring_external_domain
+  type    = "NS"
+  ttl     = "300"
+
+  records = aws_route53_zone.monitoring_public.name_servers
+}
+
+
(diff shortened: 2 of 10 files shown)


terraform/modules/monitoring/certs.tf
```diff
@@ -1,9 +1,23 @@
-resource "aws_route53_zone" "public" {
-  name = local.workspace_external_domain
+data "aws_route53_zone" "public" {
+  name = var.external_app_domain
 }
 
-resource "aws_acm_certificate" "public" {
-  domain_name = "*.${local.workspace_external_domain}"
+resource "aws_route53_record" "monitoring_public_zone_ns" {
+  zone_id = data.aws_route53_zone.public.zone_id
```
bilbof, line 6: Would this create a record into a zone created by the `govuk-publishing-platform` deployment?

If so, would that prevent us from destroying the zone without first destroying this record?

concourse/pipelines/monitoring.yml
```diff
@@ -5,7 +5,7 @@ resources:
   - icon: github
     name: govuk-infrastructure
     source:
-      branch: fix_grafana
+      branch: main
```
sengi, line 8: Oops ;) Thanks for fixing. (I could tell you an [amusing story](https://twitter.com/0xakashb/status/485207945626787840?lang=en) on the theme of accidentally leaving temporary/test changes in at review time and them going in unnoticed 😀)

terraform/modules/monitoring/certs.tf
sengi: Is it just ecs.test.govuk.digital that we want to be able to destroy nightly or also test.govuk.digital?

I think this PR basically shifts the issue one level up the DNS hierarchy (which might solve it, if we don't need to tear down test.govuk.digital).

fredericfran-gds: no, it will not. it will create `monitoring.test.govuk.digital` with `test.govuk.digital` being created in govuk-aws.

terraform/modules/monitoring/grafana_task_definition.tf
```diff
@@ -4,7 +4,7 @@ locals {
     essential   = true,
     environment = [for key, value in local.grafana_environment_variables : { name : key, value : tostring(value) }],
     healthCheck = {
-      command     = ["/bin/bash", "-c", "curl -f http://localhost:${var.grafana_port} || exit 1"]
+      command     = ["/bin/bash", "-c", "wget -q -O - http://localhost:3000/api/health || exit 1"]
```
sengi, line 7: Ah, does the base image not come with curl then? I guess we can live with having two different utilities for healthchecks. You might still want to substitute the port variable though.

sengi (approved):

terraform/modules/monitoring/grafana_task_definition.tf
fredericfran-gds: no curl, unfortunately :-(

PR #265, opened 2021-05-12
Make the govuk-publishing-platform environment agnostic

rtrinque:
As we plan to build govuk-publishing-platform in integration we need to make the terraform code usable in any environment
Main things to be aware of:
Backends are now defined per environment in a file named \<env\>.backend, to initialise you now have to run terraform init with the -backend-config=env.backend option. Also you will need to delete the .terraform/terraform.tfstate file which caches your backend config.
The way we use publishing.service.gov.uk has been modified, we never try to deal with workspaces with this domain now.

### DIFF
--- a/terraform/deployments/govuk-publishing-platform/dns.tf
+++ b/terraform/deployments/govuk-publishing-platform/dns.tf
@@ -27,7 +27,7 @@ resource "aws_route53_zone" "internal_private" {
 resource "aws_acm_certificate" "workspace_public" {
   domain_name = "*.${local.workspace_external_domain}"
 
-  subject_alternative_names = local.is_default_workspace ? ["*.${local.workspace}.${var.publishing_service_domain}"] : null
+  subject_alternative_names = local.is_default_workspace ? ["*.${var.publishing_service_domain}"] : null
 
   validation_method = "DNS"
 
--- a/terraform/deployments/govuk-publishing-platform/dns.tf
+++ b/terraform/deployments/govuk-publishing-platform/dns.tf
@@ -1,9 +1,5 @@
-data "aws_route53_zone" "public" {
-  name = var.external_app_domain
-}
-
 resource "aws_route53_record" "workspace_public_zone_ns" {
-  zone_id = data.aws_route53_zone.public.zone_id
+  zone_id = data.terraform_remote_state.infra_root_dns_zones.outputs.external_root_zone_id
   name    = local.workspace_external_domain
   type    = "NS"
   ttl     = "300"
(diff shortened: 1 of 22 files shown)


terraform/deployments/govuk-publishing-platform/integration.backend
```diff
@@ -0,0 +1,4 @@
+bucket  = "govuk-ecs-terraform-integration"
```
karlbaker02, line 1: Should we be consistent on naming here with the `test` environment? So either:

`govuk-terraform-integration`

Or (for `test`)

`govuk-ecs-terraform-test`

karlbaker02 (commented):
Looks good to me, just one minor comment 👍

terraform/deployments/govuk-publishing-platform/integration.backend
rtrinque: Yeah I picked this name to avoid any confusion with the old platform state files in integration, event though govuk-terraform-integration doesn't exist (we use govuk-terraform-state-integration and govuk-terraform-steppingstone-integration)
I feel it would be nice to have a clear delimitation now that the new platform will coexist with the current one

karlbaker02 (approved):
Looks good to me 🥇

terraform/deployments/govuk-publishing-platform/dns.tf
```diff
@@ -31,8 +27,6 @@ resource "aws_route53_zone" "internal_private" {
 resource "aws_acm_certificate" "workspace_public" {
   domain_name = "*.${local.workspace_external_domain}"
 
-  subject_alternative_names = local.is_default_workspace ? ["*.${local.workspace}.${var.publishing_service_domain}"] : null
```
fredericfran-gds, line 34: can you put this line back, just remove the `${local.workspace}`, i.e.
`subject_alternative_names = local.is_default_workspace ? ["*.${var.publishing_service_domain}"] : null`

rtrinque: Isn't it covered in https://github.com/alphagov/govuk-infrastructure/blob/integration/terraform/deployments/govuk-publishing-platform/origins_certificates.tf ?

fredericfran-gds: no, the workspace_public cert is used for the ALB of the origin and passed to the `origin` module. e.g. https://github.com/alphagov/govuk-infrastructure/blob/e87dbc0412b06bce6ca23891f8c859fc58c92d32/terraform/deployments/govuk-publishing-platform/frontends_origins.tf#L62

rtrinque: Ok got it. Fixed as suggested

PR #266, opened 2021-05-13
Disable Smokey loop at night

bilbof:
This prevents the Smokey task running when infrastructure
has been destroyed (or while it is being destroyed).

Cron expression explanation: https://crontab.guru/#*/10_8-22_*_*_1-5

This is useful to do since it saves us money, but also
because smokey ECS Tasks create resources which depend
on the resources managed by the TF deployment
govuk-publishing-platform. Therefore, when smokey is
running this prevents TF destroy from completing at
11pm.

https://trello.com/c/ywDfsw86/467-fix-issues-with-tf-destroy-failing

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -241,10 +241,10 @@ resources:
       location: "Europe/London"
       fire_immediately: true
 
-  - name: every-5-minutes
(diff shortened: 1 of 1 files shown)


fredericfran-gds (approved):
looks good, maybe change name to `every-10-minutes`

PR #267, opened 2021-05-13
Add draft-origin rather than draft-frontend in publisher

fredericfran-gds:
Since draft-router (##225) and draft-origin(#192) are now active in
the draft stack of the publishing platform, publisher should direct
draft-stack views to the the draft-origin.


bilbof (approved):
Thanks 👍

PR #268, opened 2021-05-13
Alert on failing container image builds

bilbof:
We'll be notified when a image build fails in the Slack channel `#govuk-ci-alerts`.

I manually added the secret webhook `govuk_ci_slack_webhook` IT provided.


fredericfran-gds (approved):
looks good

PR #269, opened 2021-05-13
Alert on failing smoke tests

bilbof:
This will alert us when smoke tests fail. This will make the Slack channel #govuk-deploy-alerts a lot noisier - but hopefully it'll prompt us to make the tests/apps more robust.


fredericfran-gds (approved):
looks good

PR #270, opened 2021-05-13
fix publishing-api command

fredericfran-gds:
Publishing-api is run into 2 modes: web and worker.

Therefore in the the app deployment, we need to specify which
mode do we want to run.

This will hopefully fix the issue where the publishing-api workers
are not processing the queue because they are in default web mode.


bilbof (approved):
Thanks for finding and fixing this! Great one.

Looks like the [default web process](https://github.com/alphagov/publishing-api/blob/main/Dockerfile#L25) was running for the worker :)

PR #272, opened 2021-05-14
Ignore changes to bootstrap task definitions

bilbof:
This change will make tf plan/apply diffs less verbose.

These task definitions are only needed during bootstrapping so don't need to be updated after the initial terraform apply.

### DIFF
--- a/terraform/modules/app/task_definition.tf
+++ b/terraform/modules/app/task_definition.tf
@@ -117,6 +117,12 @@ resource "aws_ecs_task_definition" "bootstrap" {
     },
   )
 
+  lifecycle {
+    # Makes tf plan/apply diffs less verbose.
+    # This task definition is needed only for bootstrapping. We don't need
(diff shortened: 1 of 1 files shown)


terraform/modules/app/task_definition.tf
```diff
@@ -117,6 +117,9 @@ resource "aws_ecs_task_definition" "bootstrap" {
     },
   )
 
+  lifecycle {
```
sengi, line 120: Might be worth pasting the PR description (basically the whole thing - it's useful context) here as a comment. It's otherwise not obvious that the ignore_changes here is primarily to reduce noisy/meaningless diffs rather than for correctness or some other function reason.

sengi (approved):

terraform/modules/app/task_definition.tf
bilbof: Good idea, I've added a comment.

PR #273, opened 2021-05-14
Eliminate manual build step for token rotation lambda.

sengi:
Instead of deploying from a manually-built zip file checked into the repo, have Terraform build the zip file.

This still isn't an ideal way to deploy a lambda, but it at least avoids the need to remember to rebuild the zip file manually and check it into the repo every time.

[Trello card](https://trello.com/c/ANYYJxdg/508)

Tested: applied ok in my workspace (after retrying a bunch of times on unrelated issues). Also plans ok in default workspace.

Potential follow-ups:
* I think it should be fairly straightforward to avoid deploying n copies of the lambda to rotate n tokens. The ID of the secret to be rotated is passed in when the lambda is called, so we should be able to have one lambda per type of rotation operation, or per system that it needs password-changing privileges on - rather than a different one for every secret. ([Trello card](https://trello.com/c/6Uaayg1l/304))
* Consider uploading the zip to S3 (`s3_bucket_object`) and having Lambda pull it from there.


bilbof:
> The ID of the secret to be rotated is passed in when the lambda is called

This is [passed in by SecretsManager](https://docs.aws.amazon.com/secretsmanager/latest/userguide/rotating-secrets-lambda-function-overview.html). Unfortunately (as far as I am aware) we're not able to modify the content of the event that SecretsManager sends, so we're not able to add in e.g. the ApiUser email to that call. Would there be another way to pass in this config? I agree having hundreds of lambdas with the same code is crazy.

bilbof (approved):
Thank you so much 🙏

sengi:
No worries - sorry for not spotting this stuff in the earlier review!

> This is passed in by SecretsManager. Unfortunately (as far as I am aware) we're not able to modify the content of the event that SecretsManager sends, so we're not able to add in e.g. the ApiUser email to that call. Would there be another way to pass in this config? I agree having hundreds of lambdas with the same code is crazy.

Ohhhh 😅 Thanks, hadn't realised that was the background to the lots-of-lambdas choice. I think the way we're supposed to get the ApiUser in there is to include it in the Secret (not because it's small-s secret but because it's metadata, similar to a username).

bilbof:
Aha understood. I didn't consider sticking the config in the secret itself 👍

PR #276, opened 2021-05-17
Silence Slack notifications in workspaces

bilbof:
I configured this incorrectly in PR #249. Disable should be set on the 'source configuration' rather than the 'resource configuration' [1].

[1]: https://github.com/cloudfoundry-community/slack-notification-resource


fredericfran-gds (approved):
LGTM

PR #277, opened 2021-05-17
Wait for secrets to be set before deploying

bilbof:
Ensures that apps will wait for SecretsManager credentials to be set before attempting to deploy the app or use the task definition for ad-hoc tasks such as running db migrations.

Will retry 8 times with an exponential backoff in Ruby. Uses 3 attempts in Concourse by default, which is configurable per deployment.

This will mostly be useful during bootstrapping. We'll also see benefit when we add credentials but forget to add them to SecretsManager - it'll make the failure more obvious in Concourse without needing to dig into the AWS console.


fredericfran-gds (approved):
not a ruby expert but looks sensible.

PR #278, opened 2021-05-17
Add ability to swith on/off DB migrations for specific GOV.UK environment

smford:
We do not want DB migrations to be enabled for all GOV.UK environments
in specific scenarios, e.g. testing in integration.  This PR allows us
to toggle that at the environment level via a setting in a parameter
file.

Ref:
1. [trello task](https://trello.com/c/CQ1CdfZV/470-ability-to-turn-off-database-migrations-needed-before-we-point-concourse-at-integration)

### DIFF
--- a/concourse/tasks/run-task.sh
+++ b/concourse/tasks/run-task.sh
@@ -30,6 +30,11 @@ COMMAND=${COMMAND:-"$(cat "run-task-command/run-task-command")"}
 : "${CLUSTER:?CLUSTER not set}"
 : "${VARIANT:?VARIANT not set}"
 
+if [[ "${DISABLE:-false}" == "true" ]]; then
+  echo "Skipping Task"
+  exit 0
+fi
+
 mkdir -p ~/.aws
 
 cat <<EOF > ~/.aws/config
--- a/concourse/README.md
+++ b/concourse/README.md
@@ -1,0 +1,57 @@
+# Configuring Concourse
+
+Our Concourse pipelines are hosted in Big Concourse located at:
+https://cd.gds-reliability.engineering/. You have access to it (and specific
+Concourse teams) by being a member of the appropriate GitHub organisation.
+
+To configure Concourse, you need a Concourse CLI client called `fly` that you
+can download by clicking on the appropriate platform icon at the bottom right of
+the [page](https://cd.gds-reliability.engineering/). You should ideally install
+it in directory `/usr/local/bin/`.
+
+Our Concourse is configured so that each GOV.UK environment (e.g.
+`test` and `staging`) having its own Concourse team, e.g. `govuk-test` concourse
+team for `test` GOV.UK environment.
+
+`fly` uses the concept of `target` to refer to a specific Concourse team. To
+register a new target with `fly`, you do:
+
+```shell
+fly --target <target_name> \
+    login \
+    --team-name <concourse_team> \
+    --concourse-url https://cd.gds-reliability.engineering
+```
+
+where:
+1. `<target_name>` is the name that you want to give to the target. By convention,
+   we use `govuk-<govuk_environment>`, e.g. `govuk_test` for `test` GOV.UK
+   environment
+1. `concourse_team` is the name of an existing Concourse team, which usually
+   follows the same format as `target_name` 
+
+## Deploying Pipelines
+
+Pipelines are deployed in Concourse teams with each GOV.UK environment (e.g.
+`test` and `staging`) having their own team with default variables/secrets
+(see `gds cd secrets ls`).
+
+### Pipeline Deployment
+
+To deploy a [pipeline](./pipelines), you need to run the following commands:
+
+```shell
+cd govuk-infrastructure
+
+fly sp -t <concourse_target> \
+       -p <pipeline_name>
+       -c concourse/pipelines/<pipeline_filename> \
+       -l concourse/parameters/<govuk_environment>/parameters.yml
+```
+
+where:
+1. `<concourse_target>` is the Concourse target that points to the relevant
+   Concourse team for the GOV.UK environment that we want to deploy to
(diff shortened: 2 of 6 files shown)


concourse/tasks/run-task.sh
```diff
@@ -30,6 +30,11 @@ COMMAND=${COMMAND:-"$(cat "run-task-command/run-task-command")"}
 : "${CLUSTER:?CLUSTER not set}"
 : "${VARIANT:?VARIANT not set}"
 
+if [[ "${SKIP_DB_MIGRATIONS:-false}" == "true" ]]; then
+  echo "Skipping DB Migrations"
+  exit 0
+fi
```
karlbaker02, line 36: One question: where is it that we're actually doing the migrations?

karlbaker02 (approved):
Thanks for the discussion, all looks good to me

concourse/parameters/environment/test/parameters.yml
```diff
@@ -0,0 +1 @@
+skip_db_migrations: true
```
bilbof, line 1: I think in the `test` environment we'd like to keep `skip_db_migrations: false`

concourse/tasks/run-task.sh
```diff
@@ -30,6 +30,11 @@ COMMAND=${COMMAND:-"$(cat "run-task-command/run-task-command")"}
 : "${CLUSTER:?CLUSTER not set}"
 : "${VARIANT:?VARIANT not set}"
 
+if [[ "${SKIP_DB_MIGRATIONS:-false}" == "true" ]]; then
+  echo "Skipping DB Migrations"
```
bilbof, line 34: Could we rename this to something more generic, since the task might not be a db migration? This script is also used for running generic tasks in ECS, such as smoke tests.

Perhaps `DISABLE` or `DRY_RUN` would be more appropriate here.

concourse/tasks/run-db-migrations.yml
```diff
@@ -0,0 +1,25 @@
...
+    tag: latest
+    username: ((docker_hub_username))
+    password: ((docker_hub_authtoken))
+inputs:
+  - name: govuk-infrastructure
+    path: src
+  - name: app-terraform-outputs
+  - name: task-definition-arn
+  - name: run-task-command
+    optional: true
```
bilbof, line 15: These two lines are unnecessary since we only provide the CMD for the db migrations tasks through the param `COMMAND` below.

bilbof: Actually, I think you can delete this file, and add `DISABLE` to the params of the current `run-task.yml`  file.

bilbof (changes requested):
Nice work, thanks for doing this! 👍 

Please could you make these changes:
* rename the disable migrations flag to something more generic, e.g. `DISABLE`
* remove the file concourse/tasks/run-db-migrations.yml, and add the `DISABLE` flag to the params there (since we're checking for it in the shared script anyway)
* remove the workspaces params file, since I think we can continue maintaining workspace branches rather than having workspace files committed to `main` (though I'm wondering about this, it might be a nice way to achieve branch deploys).

concourse/parameters/workspace/steve/deploy-parameters.yml
```diff
@@ -0,0 +1,3 @@
+workspace: steve
+govuk_infrastructure_branch: disable_db_migrations_2
+disable_slack_channel_alerts: false
```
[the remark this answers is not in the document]
smford: 👍

concourse/parameters/workspace/steve/deploy-parameters.yml
```diff
@@ -0,0 +1,3 @@
+workspace: steve
+govuk_infrastructure_branch: disable_db_migrations_2
```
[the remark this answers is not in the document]
smford: I was torn on how to implement this myself, there were pros and cons.  I ultimately decided raise a PR and see what others think, so your thoughts on this is much appreciated.

I think that there isn't any single clean way to do it and the choice ultimately comes down to: have the workspace params file captured in git or not.

In git:
- con: we face the problem you mention
- con: junk workspace param files end up in git

Out of git: 
- pro+con: we need to rely on people maintaining it themselves
- con: this means that when we deploy to production (using the `default` workspace), that the workspace config will need to be created, placed in to git, and deployment of pipeline documented, which could circle around to the "in git method" and reintroduce con1, depending on implementation method.

Regardless, I'm cool with whatever method, and I've happily adjusted as per your thoughts.  As we progress we might need to tweak the management of workspace params later.

Concerning ((govuk_environment)) yeah - I agree, that will be the best way forward, but I'll pass on doing that at the moment to get this PR deployed.

concourse/tasks/run-db-migrations.yml
smford: Im following you, yup thats a good idea, it makes it all generic rather than having an ackward almost duplication of run-task.yml file.

concourse/tasks/run-task.sh
smford: aye aye, good idea.

bilbof (approved):
Great stuff, thank you so much. Did you already set `govuk_environment` as a variable on the `govuk-ci`, `govuk-test` and `govuk-integration` teams?

fredericfran-gds:
> Great stuff, thank you so much. Did you already set `govuk_environment` as a variable on the `govuk-ci`, `govuk-test` and `govuk-integration` teams?

oops, forgot about `govuk-ci`, I will do that now

fredericfran-gds:
> > Great stuff, thank you so much. Did you already set `govuk_environment` as a variable on the `govuk-ci`, `govuk-test` and `govuk-integration` teams?
> 
> oops, forgot about `govuk-ci`, I will do that now

done! In govuk-ci, govuk_environment = test

PR #279, opened 2021-05-17
Trigger app smoke tests on all deploys

bilbof:
This ensures app smoke tests are triggered when a new image or new task definition config is deployed, which I expect to be the most frequent trigger of smoke tests.

This was mistakenly omitted in PR #253.


fredericfran-gds (approved):
ok

PR #280, opened 2021-05-17
Update Signon::Client to use new status code

bilbof:
Uses the change in Signon PR #1644. Backwards compatible.


fredericfran-gds (approved):
ok

PR #284, opened 2021-05-18
Move provider config to root module

bilbof:
It's recommended that providers are only defined in root modules, to avoid issues around provider versions.

This should be a no-op.

The motivation for this change is I want to refactor the client of the signon_bearer_token module to use `for_each` on the `module` block. To do this I need to explicitly pass in the parent module's providers to the child module.

https://www.terraform.io/docs/language/modules/develop/providers.html

### DIFF
--- a/terraform/deployments/govuk-publishing-platform/main.tf
+++ b/terraform/deployments/govuk-publishing-platform/main.tf
@@ -17,6 +17,11 @@ terraform {
       source  = "hashicorp/random"
       version = "3.0.0"
     }
+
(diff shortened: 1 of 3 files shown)


fredericfran-gds (approved):
looks good, thanks for doing that.

I guess you meant that the child module now inherits implicitly its providers from the parent deployment.

bilbof:
Yeah exactly for now it is implicit. Later, when I use `for_each` to call this module, I will provide the provider explicitly.

PR #285, opened 2021-05-18
Fixup draft content-stores router-api bearer token

bilbof:
This now gives draft content store access to draft router-api.


fredericfran-gds (approved):
ok

PR #287, opened 2021-05-19
Wait longer for secrets to be set

bilbof:
This change will increase the amount of time
an app waits for secrets to be set before it
attempts to deploy from ~4mins to ~17mins.

We also re-attempt per app in the deploy pipeline

This should stop those errors we saw this morning
when we destroyed and re-ran the deploy pipeline.


fredericfran-gds (approved):
wow, 17 minutes, that's a lot.

bilbof:
Yeah. And it will try 3 times, so 3*17. Now we are starting the bootstrap/deploy of signon at the same time as deploying other apps, so the other apps have to wait for signon to come into existence. Roughly 15-25mins.

PR #289, opened 2021-05-20
Fixup bootstrapping

bilbof:
Paired with @nsabri1 on this one.

Two changes to prevent the deploy-apps pipeline from raising errors in the morning when we recreate the test environment:

- Fixup lambda rotation dependency
- Add secretsmanager waiter task to the concourse tasks (so the change in #287 will have an effect)


nsabri1 (approved):
LGTM

PR #292, opened 2021-05-20
Add default background_image of the GOV.UK cake

richardTowers:
The idea here is that people's individual pipelines can use different
background images, to make them a bit more visually distinct. This should
hopefully avoid someone accidentally doing something in the wrong pipeline, and
also adds fun.

I've use a picture of some cake from GOV.UK's launch (from GDS' Flickr account)
as the default image. Obviously my workspaces will use Doge.

image


bilbof (approved):
So good 👍

PR #293, opened 2021-05-20
Fixup watched paths

bilbof:
These paths don't trigger builds currently since they're incorrect.

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -55,7 +55,7 @@ resources:
     icon: concourse-ci
     source:
       <<: *govuk-infrastructure-source
-      paths: [ concourse/tasks, lib/signon, tasks/secretsmanager.rake, tasks/app_secrets.rake ]
+      paths: [ concourse/tasks, lib/signon, lib/tasks/secretsmanager.rake, lib/tasks/app_secrets.rake ]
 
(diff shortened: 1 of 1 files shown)


concourse/pipelines/deploy.yml
```diff
@@ -55,7 +55,7 @@ resources:
     icon: concourse-ci
     source:
       <<: *govuk-infrastructure-source
-      paths: [ concourse/tasks, lib/signon, tasks/secretsmanager.rake, tasks/app_secrets.rake ]
+      paths: [ concourse/tasks, lib/signon, lib/tasks/secretsmanager.rake, lib/tasks/app_secrets.rake ]
```
richardTowers, line 58: Maybe put the whole of `lib/tasks` here? Will there be any tasks in there that shouldn't create new versions of this resource?

bilbof: There's `lib/tasks/{lint,test}.rake` too. Maybe I should put these tasks into a sub directory or create a `concourse` file.

richardTowers (approved):

concourse/pipelines/deploy.yml
richardTowers: Cool, not that important. Approved 😄

PR #294, opened 2021-05-20
Add documentation on creating environments, workspaces, and secrets

bilbof:
This adds a few different docs on creating a new environment or workspace, a glossary, and signon secrets.

Rendered: https://github.com/alphagov/govuk-infrastructure/tree/6c062026477e477e3335ada503729ca31dae0d44/docs

These are quite quickly written docs, not intended to be exhaustive, since I'd rather have something rough we can add to as we go along.

Of course, if there are things that misleading / incorrect, I'd be happy to correct those.

https://trello.com/c/7xjSjFOW/517-document-bootstrapping-process


rtrinque (approved):
Looks good

PR #295, opened 2021-05-20
put common tfvars variables in common files

fredericfran-gds:
some variables are duplicated and in this work, we put all the common variables together.


terraform/deployments/variables/test/common.tfvars
```diff
@@ -0,0 +1,8 @@
+govuk_aws_state_bucket = "govuk-terraform-steppingstone-test"
+
+govuk_environment             = "test"
+ecs_default_capacity_provider = "FARGATE_SPOT"
+
+publishing_service_domain = "test.publishing.service.gov.uk"
+internal_app_domain       = "test.govuk-internal.digital"
+external_app_domain       = "test.govuk.digital"
```
bilbof, line 8: It seems that apart from `govuk_environment` all of the other variables could be an interpolated `local` e.g.
`locals = { external_app_domain = "${var.govuk_environment}.govuk.digital" }`. Then I guess the content of this file could just be `govuk_environment = "test"`. Would that be possible?

bilbof (approved):
Looks good, thanks for updating the docs

terraform/deployments/variables/test/common.tfvars
fredericfran-gds: we could do it but I think it might be too much optimization. Some people might want to parametize the `govuk.digital` part too.

PR #302, opened 2021-05-21
Fixup container_id in run-task

bilbof:
Fixes a small issue with the run-task script.

Verified it works in my pipeline: https://cd.gds-reliability.engineering/teams/govuk-test/pipelines/deploy-apps-bill/jobs/all-smoke-tests/builds/1207


bilbof:
Ah thanks, great minds, yours is better :)

PR #304, opened 2021-05-21
Fixup bearer token bootstrapping

bilbof:
This fixes an issue seen in the `deploy-apps-towers` pipeline: https://cd.gds-reliability.engineering/teams/govuk-test/pipelines/deploy-apps-towers/jobs/deploy-publisher/builds/1.2.

For some deploy jobs the `await-creds-in-secrets-manager` task blocks the `bootstrap-secrets` task that would actually
create the remaining secrets. Therefore these deploys would stall forever.

The two extra commits fix another issue: since the `bootstrap-secrets` task can occur before the signon API it relies on is ready, it needs to retry for a longer period. Now this task will retry for an extra 17 minutes until it bails. This should be enough additional time for Signon to come up.

I tested this change against the `towers` workspace using the same pipeline and verified this problem was fixed: https://cd.gds-reliability.engineering/teams/govuk-test/pipelines/deploy-apps-towers/jobs/deploy-content-store/builds/4. The build failed for an unrelated reason (commit cf1c27d2 seems to have broken the update-ecs-service script)

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -612,18 +612,6 @@ jobs:
       - get: frontend-image
         trigger: true
     - in_parallel:
-      - &await-secretsmanager-creds
-        task: await-creds-in-secrets-manager
-        file: govuk-infrastructure/concourse/tasks/await-creds-in-secretsmanager.yml
-        attempts: 3
-        params:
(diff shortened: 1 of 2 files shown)


fredericfran-gds (approved):
ok

PR #305, opened 2021-05-24
Don't wait for ECS service stable in update-ecs-service task

bilbof:
We wait for the deploy to complete in the await-deploy-complete task, which is retried with n attempts. The update-ecs-task job is merely for setting the new task definition on the ECS service and thereby triggering the deploy.

This waiter was removed but was re-added accidentally in https://github.com/alphagov/govuk-infrastructure/pull/237/files#diff-32213ce85db7c49f7f2026f29e45a96e298f80b7c85c69e028d666c72971439cR58-R61.


fredericfran-gds (approved):
cool, thanks

PR #306, opened 2021-05-24
Disable refresh on CI plan

bilbof:
This disables the default behavior of synchronizing the
local Terraform state with remote objects before checking
for configuration changes.

This will make the CI planner test faster, since it reduces
the number of API requests from refreshing the state.

However, it makes the test less reliable since it ignores
changes that have been made outside of Terraform, therefore
risks the plan not being complete or correct.


sengi (approved):
Seems a sensible compromise since we're mostly just trying to check that the config is internally consistent. Might also make the test less flaky :)

PR #307, opened 2021-05-24
Fix example command for running terraform apply

sengi:
Now that we're using backend files, we need to specify `-backend-config` in order for tf init to work.

Also add a step for running `terraform workspace select`, as that's an easily avoidable gotcha.

Tested: copy/pasting these steps worked for me as far as running a plan.


bilbof (approved):
Thanks a lot

PR #311, opened 2021-05-24
Commit workspace params files to main branch

bilbof:
Props to @smford for this - he implemented this in an earlier PR.

This includes the pipeline file in the main branch. The benefit is that we can keep our workspaces up to date without needing to rebase frequently. This will reduce the amount of time we need to spend repairing workspaces that fell too far behind.

If we want to work on a separate branch, we can create a new branch and point our pipeline at that branch briefly.

The effect of this PR is that Steve's workspace will be up to date with `main` when he returns from vacation, and won't need to spend time repairing it.

### DIFF
--- a/concourse/parameters/test/steve.yml
+++ b/concourse/parameters/test/steve.yml
@@ -1,0 +1,5 @@
+govuk_infrastructure_branch: main
+workspace: steve
+disable_slack_channel_alerts: true
+skip_db_migrations: false
+background_image: "https://images.unsplash.com/photo-1494022299300-899b96e49893?ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&ixlib=rb-1.2.1&auto=format&fit=crop&w=1350&q=80"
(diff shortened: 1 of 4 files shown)


concourse/parameters/test/steve.yml
```diff
@@ -0,0 +1,5 @@
+govuk_infrastructure_branch: main
+workspace: steve
+disable_slack_channel_alerts: true
+skip_db_migrations: false
+background_image: "https://images.unsplash.com/photo-1494022299300-899b96e49893?ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&ixlib=rb-1.2.1&auto=format&fit=crop&w=1350&q=80"
```
sengi, line 5: I appreciate this is somewhat a pre-existing issue, but hotlinking to image hosting sites like this seems slightly iffy to me. It's a minor thing but probably not the best example to be setting (bearing in mind this is still a public repo). I'd be inclined to remove these or move them to GitHub or something.

sengi (approved):
Having chatted about this I'm still not really getting how this might save effort, but I also don't see any harm if people want to use this workflow.

bilbof:
Closing this PR since it will make the workspaces workflow more complicated. Instead we'll share some guidance on how to use workspaces to ensure one's workspace doesn't break frequently.

PR #313, opened 2021-05-25
Wait longer for deploys to complete

bilbof:
This adds an extra 5 minutes to the max wait time on deploy completions. This should resolve the issues we've seen where a deploy waiter times out, but the deploy was successful a minute or so later.

For example, this morning this task failed: https://cd.gds-reliability.engineering/teams/govuk-test/pipelines/deploy-apps/jobs/deploy-authenticating-proxy/builds/42.

We're learning that it takes up to 25 minutes for ECS to do a rolling deployment of just one task with almost no traffic.

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -674,7 +674,7 @@ jobs:
       - &await-deploy-complete
         task: await-deploy-complete-live
         file: govuk-infrastructure/concourse/tasks/await-deploy-complete.yml
-        attempts: 4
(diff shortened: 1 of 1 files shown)


sengi:
I suspect the slowness is just the variability in admission delay for spot instances (which we're using in the test environment because it's be a waste of money not to, especially with all these personal environments running). 25 minutes still seems like a perfectly reasonable timeout though - just hopefully it won't need to wait that long with on-demand instances!

sengi (approved):

PR #315, opened 2021-05-25
Replace hard coded occurences of account id by variable

rtrinque:
Replace hard coded occurences of account id by variable


fredericfran-gds:
You can use: https://github.com/alphagov/govuk-infrastructure/blob/c66295306d552d869350670b5249dd53a433ebf6/terraform/deployments/concourse-iam/main.tf#L29 to extract the account number

nsabri1 (approved):
LGTM

PR #316, opened 2021-05-25
Add integration environment to ECR allow list

rtrinque:
Add integration environment to ECR allow list


sengi (approved):

bilbof (approved):
LGTM

A nit to save me a `git blame` later. Could you put a comment next to each line to say which account they're for? Or have variables like `var.integration_account_id`?

PR #319, opened 2021-05-26
parametize backend config for `task-runner` deployment

fredericfran-gds:
This is needed as part of the deployment in integration.

### DIFF
--- a/terraform/deployments/task-runner/integration.backend
+++ b/terraform/deployments/task-runner/integration.backend
@@ -1,0 +1,5 @@
+bucket  = "govuk-ecs-terraform-integration"
+key     = "projects/task-runner.tfstate"
+encrypt = true
+region  = "eu-west-1"
+dynamodb_table = "terraform-lock"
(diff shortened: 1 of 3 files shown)


terraform/deployments/task-runner/integration.backend
```diff
@@ -0,0 +1,5 @@
+bucket  = "govuk-ecs-terraform-integration"
+key     = "projects/rake-task.tfstate.tfstate"
```
bilbof, line 2: `rake-task.tfstate.tfstate` looks incorrect. Plus, this cluster is not just for rake tasks, it's for any arbitrary command.

sengi: +1 - task-runner.tfstate would be the obvious choice.

fredericfran-gds: sorry, too much copy, pasting, I will fix

bilbof (approved):

PR #320, opened 2021-05-26
Autogenerate secret_key_base for Frontend app

bilbof:
Trello: https://trello.com/c/HQxqvue6/505-autogenerate-secretkeybase-when-bootstrapping

This autogenerates the `secret_key_base` secret for Rails apps (used for [cookie signing](https://edgeapi.rubyonrails.org/classes/ActionDispatch/Cookies/ChainedCookieJars.html#method-i-signed)). This job is idempotent and will not attempt to create a secret if one already exists in SecretsManager.

I've started with the `frontend` app, it'll be trivial to update the other apps later.

See my pipeline for an example of this running: https://cd.gds-reliability.engineering/teams/govuk-test/pipelines/deploy-apps-bill.

### DIFF
--- a/lib/tasks/secretsmanager.rake
+++ b/lib/tasks/secretsmanager.rake
@@ -67,11 +67,11 @@ namespace :secretsmanager do
     config.fetch("arns").each do |secret_arn|
       metadata = secrets_client.describe_secret(secret_id: secret_arn)
       versions = metadata.version_ids_to_stages
-      is_unset = versions.nil? || versions.values.none? do |stages|
+      already_set = !versions.nil? && versions.values.any? do |stages|
         stages.include?("AWSCURRENT")
       end
 
-      unless is_unset
+      if already_set
         puts "Secret #{secret_arn} already set; skipping. 🔒"
         next
       end
--- a/lib/tasks/secretsmanager.rake
+++ b/lib/tasks/secretsmanager.rake
@@ -49,4 +49,42 @@ namespace :secretsmanager do
       exit 1
     end
   end
+
+  desc "Autogenerate secret_key_base for Rails apps"
+  task :autogenerate_secret_key_base do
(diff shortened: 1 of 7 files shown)


lib/tasks/secretsmanager.rake
```diff
@@ -49,4 +49,42 @@ namespace :secretsmanager do
...
+      end
+
+      unless is_unset
+        puts "Secret #{secret_arn} already set; skipping. 🔒"
+        next
+      end
+
+      secrets_client.put_secret_value(
+        secret_id: secret_arn,
+        secret_string: SecureRandom.hex(64),
```
richardTowers, line 81: I was going to say this is probably longer than it needs to be, but actually this is what rails itself does!

https://github.com/rails/rails/blob/main/railties/lib/rails/tasks/misc.rake#L6

lib/tasks/secretsmanager.rake
```diff
@@ -49,4 +49,42 @@ namespace :secretsmanager do
...
+    )
+
+    config.fetch("arns").each do |secret_arn|
+      metadata = secrets_client.describe_secret(secret_id: secret_arn)
+      versions = metadata.version_ids_to_stages
+      is_unset = versions.nil? || versions.values.none? do |stages|
+        stages.include?("AWSCURRENT")
+      end
+
+      unless is_unset
```
sengi, line 74: For clarity I'd consider avoiding the double-negative here. i.e. I'd reverse the sense of this bool, call it `is_set` (or perhaps `already_set`) and change the unless to an if. (There's a great [testing on the toilet](https://testing.googleblog.com/2007/01/introducing-testing-on-toilet.html) episode for this but I can't find it. It's similar to [this blog post](https://derickbailey.com/2014/03/27/double-negatives-in-code-dont-not-avoid-double-negative-boolean-logic/) though.)

sengi (approved):

PR #321, opened 2021-05-27
fix content store db name

fredericfran-gds:
In `test` GOV.UK environment, we were using `live_content_store_production` as the db name for the live content store but in other environment, it is `content_store_production`. This PR resolves this unexplained anomaly.

Ref:
1. [trello card](https://trello.com/c/lZQJKP1h/485-spin-up-in-integration)


sengi (approved):
👍

PR #322, opened 2021-06-03
Add outbound traffic to draft content store security group

karlbaker02:
This PR allows all outbound traffic to the draft content store security group. This rule already exists for the (origin) content store security group, and without it draft content store is unable to talk to MongoDB, resulting in timeouts which propagate up the stack to CloudFront serving an `upstream request timeout` error.


fredericfran-gds (approved):
looks good.

PR #324, opened 2021-06-04
Redeploy apps when secrets are rotated

bilbof:
Depends on https://github.com/alphagov/paas-s3-resource/pull/10
Trello: https://trello.com/c/LKgHd7ph/523-redeploy-apps-when-their-secrets-are-rotated

This adds an s3-iam resource that gives access to S3 buckets
in our AWS account (as opposed to the AWS account owned by
the team behind Big Concourse).

The frontend-deploy-event-trigger Concourse job will redeploy
the frontend app whenever a new object is placed in the frontend
bucket in `deploy-events-((workspace))`.

The intent is for the Secret Rotation Lambda to place a timestamped
file into the bucket whenever a secret an app uses is rotated,
triggering the app to be redeployed and the env var secrets for the
app's tasks to be refreshed from SecretsManager.

Screenshots:
-------------

Lambda places a file in an S3 bucket after rotating a secret:

Concourse resources checks for objects in the S3 bucket, triggering the deploy jobs:

Screenshot 2021-06-04 at 15 47 21


concourse/parameters/test/deploy.yml
```diff
@@ -1,6 +1,6 @@
-govuk_infrastructure_branch: main
+govuk_infrastructure_branch: bilbof/rotate-triggers-redeploy
 concourse_deployer_role_arn: arn:aws:iam::430354129336:role/govuk-concourse-deployer
-workspace: default
-disable_slack_channel_alerts: false
-skip_db_migrations: false
+workspace: bill
+disable_slack_channel_alerts: true
+skip_db_migrations: true
```
richardTowers, line 5: This wants scrubbing before we merge.

concourse/parameters/test/deploy.yml
```diff
@@ -1,6 +1,6 @@
-govuk_infrastructure_branch: main
+govuk_infrastructure_branch: bilbof/rotate-triggers-redeploy
```
sengi, line 1: Don't forget to revert this before merging :)

terraform/deployments/govuk-publishing-platform/signon_bearer_tokens.tf
```diff
@@ -25,65 +25,89 @@ locals {
...
 
     dcs_to_draft_router_api = {
-      api_user = local.signon_api_user.draft_content_store
-      app      = local.signon_app.draft_router_api.name
+      api_user   = local.signon_api_user.draft_content_store.email
+      app        = local.signon_app.draft_router_api.name
+      client_app = local.signon_api_user.draft_content_store.app_name
     }
 
     pub_to_pub_api = {
```
karlbaker02, line 75: Are we potentially missing draft Publisher to Publishing API?

terraform/modules/signon_bearer_token/iam.tf
```diff
@@ -113,3 +113,28 @@ resource "aws_iam_policy" "vpc" {
...
+  path        = "/"
+  description = "Allow lambda to write events to S3 deploy-events bucket"
+
+  policy = jsonencode({
+    Version = "2012-10-17",
+    Statement = [
+      {
+        Action = [
+          "s3:PutObject",
+          "s3:PutObjectAcl"
```
karlbaker02, line 133: Just wondering what `PubObjectAcl` is needed for specifically?

karlbaker02 (commented):
Looks good, a few small comments 🙂

terraform/deployments/govuk-publishing-platform/signon_bearer_tokens.tf
bilbof: Nah, there's no draft publisher app, we only have draft instances on the frontend side of things https://docs.publishing.service.gov.uk/manual/content-preview.html#design

terraform/modules/signon_bearer_token/iam.tf
bilbof: This is a [required permission](https://github.com/concourse/s3-resource#required-iam-permissions) for the concourse resource. 

When putting an object you can set the ACL on it, I think the default is [`private`](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html#canned-acl).

```
- put: release
  params:
    file: path/to/release-*.tgz
    acl: public-read
```

concourse/pipelines/deploy.yml
```diff
@@ -171,6 +178,89 @@ resources:
...
       initial_version: "0"
 
+  - &deploy-event-trigger
+    name: frontend-deploy-event-trigger
+    type: s3-iam
+    icon: file
+    source: &deploy-event-trigger-source
+      aws_role_arn: ((concourse_deployer_role_arn))
+      bucket: deploy-event-((workspace))
+      region_name: eu-west-1
```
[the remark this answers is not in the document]
bilbof: Hm, not yet - I think we're only in eu-west-1 right now, we don't set a variable for this in Terraform either.

bilbof:
Thanks all for your reviews. I've reset `concourse/parameters/test/deploy.yml` to what is on `main`.

terraform/modules/signon_bearer_token/iam.tf
karlbaker02: Interesting, thanks!

karlbaker02 (approved):
Great work Bill! 🥇

PR #326, opened 2021-06-04
Allow draft-router-api from draft-router-api security group

karlbaker02:
This PR allows all outbound traffic from the draft-router-api security group


fredericfran-gds (approved):
looks good

PR #327, opened 2021-06-07
Exclude traffic to services outside the cluster from App Mesh.

sengi:
This is to (hopefully) prevent the intermittent TCP failures that we've been seeing occasionally on connections to MongoDB. It's a hard-to-reproduce problem so it makes sense to just exclude all this traffic from App Mesh preemptively, given that there's no benefit to having it go through Envoy. We don't plan to use any of the App Mesh features for this traffic anyway so there's no reason for it to.

See https://trello.com/c/6yn0REmI/368.

### DIFF
--- a/terraform/modules/envoy-configuration/main.tf
+++ b/terraform/modules/envoy-configuration/main.tf
@@ -72,6 +72,14 @@ output "proxy_properties" {
     # additional addresses."
     EgressIgnoredIPs = "169.254.170.2,169.254.169.254"
 
+    # https://trello.com/c/6yn0REmI/368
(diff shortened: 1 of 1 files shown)


sengi:
On second thoughts we don't want to merge this to main for a bit - makes sense to run this on a workspace for a while and see how it differs from main.

sengi:
Rather than put time and effort into reproducing this tricky, intermittent problem, let's just exclude all the traffic to external services now and then see if the problem reoccurs. If the App Mesh / Envoy / iptables stuff was the culprit then it won't, and if it does then we know it warrants further investigation.

bilbof (approved):
LGTM, thanks!

PR #328, opened 2021-06-07
Log Smokey output to AWS

karlbaker02:
This PR updates the logging for Smokey to be sent to AWS, rather than Splunk. This is necessary so that developers can view the entire output of a particular Smokey job when debugging issues, as logs sent to Splunk are broken up by log line and means navigating multiple search results pages to see the output of a single job.

Trello: https://trello.com/c/9QipS0X0


concourse/tasks/run-task.sh
```diff
@@ -88,8 +88,12 @@ aws ecs wait tasks-stopped --tasks $task_id --cluster $CLUSTER
 echo "task finished."
 task_results=$(aws ecs describe-tasks --tasks $task_id --cluster $CLUSTER)
 
-container_id=$(echo $task_results | jq '.tasks[0].containers[]  | select(.name=="app") | .runtimeId')
-echo "Check Splunk for logs: https://gds.splunkcloud.com/en-GB/app/gds-006-govuk/search?q=search%20index%3D%22govuk_replatforming%22%20container_id%3D$container_id"
+if [[ "${LOG_TO_SPLUNK:-true}" == "true" ]]; then
```
bilbof, line 91: Could we override the log config in `containerOverrides` so that all tasks run with this `run-task` script use the `awslogs` log driver?

I think the code would be simpler and one wouldn't need to make additional changes (such as setting the `LOG_TO_SPLUNK` flag) to get logs to output to Concourse.

karlbaker02: I think after our recent conversation that we've determined that it's not possible to override the `logConfiguration` in container overrides - indeed [this is an issue on the ECS containers roadmap](https://github.com/aws/containers-roadmap/issues/74).

bilbof (approved):
Thanks for looking into this. Feel free to merge. It's a good demonstration that Splunk might not be a good fit for us, since we are working around it rather than working with it.

I wonder if instead we should go directly back to using `awslogs` for all containers, and have a look at using fluentd and a new kibana stack. I think it might save us from some complexity by not supporting two logging drivers. What do you think?

karlbaker02:
> Thanks for looking into this. Feel free to merge. It's a good demonstration that Splunk might not be a good fit for us, since we are working around it rather than working with it.
> 
> I wonder if instead we should go directly back to using `awslogs` for all containers, and have a look at using fluentd and a new kibana stack. I think it might save us from some complexity by not supporting two logging drivers. What do you think?

Fair points. I think in terms of this PR I'll keep as using `awslogs` only for Smokey - I think it we switch back to `awslogs` for everything then we might want to do that in a separate PR with reasoning attached.

PR #329, opened 2021-06-07
Exclude smoke tests not for current environment

karlbaker02:
This PR excludes smoke tests not meant for the current environment when they're run by Concourse. To facilitate this, any tests not meant for a particular environment should be marked up as `@replatformingnot<environment>` in Smokey, where `<environment>` is to be replaced with the name of the environment in which they should not be run.

The reason for this change is to allow us to run smoke tests for publishing content, ensuring that they do not run in production.

Trello: https://trello.com/c/9QipS0X0

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -618,7 +618,7 @@ jobs:
         ASSUME_ROLE_ARN: ((concourse_deployer_role_arn))
         APPLICATION: smokey
         VARIANT: default
-        COMMAND: bundle exec cucumber --profile test --strict-undefined -t @replatforming -t \"not @notreplatforming\"
+        COMMAND: bundle exec cucumber --profile test --strict-undefined -t @replatforming -t \"not @notreplatforming and not @not((govuk_environment))\"
     on_failure:
       <<: *notify-slack-failure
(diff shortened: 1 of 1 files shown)


concourse/pipelines/deploy.yml
```diff
@@ -618,7 +618,7 @@ jobs:
         ASSUME_ROLE_ARN: ((concourse_deployer_role_arn))
         APPLICATION: smokey
         VARIANT: default
-        COMMAND: bundle exec cucumber --profile test --strict-undefined -t @replatforming -t \"not @notreplatforming\"
+        COMMAND: bundle exec cucumber --profile test --strict-undefined -t @replatforming -t \"not @notreplatforming and not @replatformingnot((govuk_environment))\"
```
bilbof, line 621: Could we use `@not((govuk_environment))` instead since there's an existing `@notproduction` tag?

karlbaker02: We could, absolutely, especially if this `@replatforming` tag is going to go away?

bilbof (approved):

bilbof:
Nice I didn't know that the tags expression could contain a logical AND.

PR #330, opened 2021-06-07
Upgrade terraform version to 0.15.5

bilbof:
This upgrades us from version 0.14.6 to the current version.

There are many minor improvements and bug fixes, nothing particularly notable for us. See the change log: https://github.com/hashicorp/terraform/blob/v0.15.5/CHANGELOG.md


terraform/modules/container-definition/variables.tf
```diff
@@ -19,7 +19,7 @@ DESC
 }
 
 variable "dependsOn" {
-  type        = list(any)
+  type        = list(object({ containerName = string, condition = string }))
```
karlbaker02, line 22: In this case though, there's only a single list element (type) of `object` - is the problem that this object contains `string` properties?

karlbaker02 (commented):
One small comment, everything looks pretty sensible 🥇

karlbaker02 (approved):
LGTM, after discussing in more detail with Bill

PR #331, opened 2021-06-08
Add logging to CloudFront distribution to Splunk

OllieJC:
**Pending**
 https://github.com/alphagov/cyber-security-terraform/pull/356 and deployment of IAM permissions to our central CloudFront bucket

**What?** 
Adds logging from CloudFront to central S3 bucket which ends up in Splunk.
Uses prefix with `var.workspace` to align to apparent GOV.UK usage.

**Why?**
Increased visibility when CDN bypassed through direct access or failover event.


sengi:
We're no longer using CloudFront or indeed ECS, so let's clean this up now. (Thanks for your help though!)

PR #333, opened 2021-06-10
Allow router-api from draft-router-api security group

karlbaker02:
Similarly to https://github.com/alphagov/govuk-infrastructure/pull/326/commits/b3d3cd2c06ecf47636d09ec6a32154e86920d83d, this PR allows all outbound traffic from the router-api security group, which is necessary for Router API to be able to commit routes to MongoDB.


sengi (approved):
On the one hand it's overly broad, but on the other it's the same as we already have in prod, so let's go with it at least for now.

PR #334, opened 2021-06-10
Fix instructions for creating workspace Concourse pipelines.

sengi:
If you forget to commit and/or push the deploy-parameters.yml changes, the pipeline config will look correct when you create it but will update-pipeline itself back to the config from the main branch as soon as it runs.


bilbof (approved):
Thank you!

PR #336, opened 2021-06-14
Add missing LOG_TO_SPLUNK param to Smokey runs

karlbaker02:
This PR adds the missing `LOG_TO_SPLUNK` param to Smokey runs and sets it to `false` so that we can inspect Smokey job output in Concourse. This should have been added in https://github.com/alphagov/govuk-infrastructure/pull/328 but seems to have been missed out.


nsabri1 (commented):
LGTM

bilbof (approved):

PR #337, opened 2021-06-15
Retry failed smoke tests

bilbof:
This will retry any failed smoke test 3 times if it fails.

It should make our flaky smoke tests a bit more reliable.

For example, the [most recent](https://cd.gds-reliability.engineering/teams/govuk-test/pipelines/deploy-apps/jobs/all-smoke-tests/builds/2037) smokey invocation failed due
to the browser mob proxy failing to start up in enough time.
The job succeeded when manually retried.

We should also look at making these tests more robust, this
is a quick fix.


fredericfran-gds (approved):
ok, though I think the root cause should be fixed soon.

PR #343, opened 2021-06-18
Be more explicit about which tests to run

richardTowers:
There are some new accounts tests which are relevant to @app-frontend,
but don't pass in the replatforming subset of apps (yet). We could mark
these as `@notreplatforming`, but it probably makes more sense for us to
only run tests which have been explicitly marked as okay for
`@replatforming`.

The tests we actually want to run in smokey are already marked
explicitly:

```
smokey ▶ main ▶ % ▶ git grep '@replatforming'
features/frontend.feature:@aws @replatforming @app-frontend
features/publisher.feature:@app-publisher @replatforming
features/publishing_tools.feature:  @app-publisher @replatforming
features/signon.feature:@app-signon @replatforming
```

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -856,7 +856,7 @@ jobs:
     - <<: *run-smoke-test
       params:
         <<: *smoke-tests-params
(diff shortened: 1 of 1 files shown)


karlbaker02 (approved):
Looks like a sensible thing to do 👍

bilbof (approved):
Thank you!

PR #344, opened 2021-06-18
Enable execute command

richardTowers:
This enables ECS Exec for all apps, and documents how to use it (along with one workaround for when it's broken).

Basically the best UX we can manage without writing any code is:

```
task_arn="$(aws ecs list-tasks --cluster govuk-ecs --service-name frontend --query 'taskArns[0]' --output text)"
aws ecs execute-command \
  --cluster govuk-ecs \
  --task "$task_arn" \
  --container app \
  --interactive \
  --command "bash"
 ```
 
 Which is not great, but not terrible.

We didn't do this previously because we were sticking to a principle of creating separate tasks for ad hoc commands. That's a nice idea because it avoids the situation where an operator does something bad in a container that's busy serving production traffic. But it's very inconvenient, particularly while ECS Exec is so buggy.

### DIFF
--- a/docs/execute-commands.md
+++ b/docs/execute-commands.md
@@ -1,6 +1,6 @@
 # Execute commands
 
-You can run commands inside containers use Amazon ECS Exec. This is a lot
+You can run commands inside containers using Amazon ECS Exec. This is a lot
 like SSH, although what happens under the hood is quite different.
 
 To use this feature, you need to install the AWS CLI and [install the Session Manager plugin][].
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: docs/execute-commands.md [hunk 1]
```diff
@@ -0,0 +1,74 @@
+# Execute commands
+
+You can run commands inside containers use Amazon ECS Exec. This is a lot
```
bilbof, line 3:
```suggestion
You can run commands inside containers using Amazon ECS Exec. This is a lot
```

docs/execute-commands.md
```diff
@@ -0,0 +1,74 @@
...
+
+```
+An error occurred (InvalidParameterException) when calling the ExecuteCommand
+operation: The execute command failed because execute command was not enabled
+when the task was run or the execute command agent isn’t running. Wait and try
+again or run a new task with execute command enabled and try again.
+```
+
+This sometimes happens in the `app` container, sometimes in the `envoy`
+container, sometimes in both.
```
bilbof, line 51:
```suggestion
container, sometimes in both. It can also occur if the task is not yet in a `READY` state, i.e. it just started, so ensure that the container is ready before troubleshooting further.
```

bilbof (approved):
Thank you so much 👍

bilbof (approved):
⭐

PR #346, opened 2021-06-20
Fix freudian slip in ECS Exec docs

richardTowers:
It's ECS Exec which is the buggy new feature, not the whole of ECS 🙃


sengi (commented):
😅

sengi (approved):
Oops, thought I'd approved but apparently not 🙈 - sorry about the wait!

PR #348, opened 2021-07-12
Add environment variable GOVUK_ENVIRONMENT_NAME for apps

karlbaker02:
This PR adds the new environment variable `GOVUK_ENVIRONMENT_NAME`, which will be used as the canonical source of truth to determine an app's environment for the purposes of setting the `environment_label` and `environment_style` properties which are used by apps still requiring the deprecated `govuk_admin_template` gem.

Once this change is in, apps using said deprecated gem in the `test` account will correctly show the current environment and display the correct colour scheme relating to this environment (currently this is missing because we (correctly) [don't create a specific initializer file during app deploy for `test`](https://github.com/alphagov/govuk-app-deployment/blob/c52a9c767dff658a912aa62cea5adde6d28ea94d/recipes/govuk_admin_template.rb)).

Trello: https://trello.com/c/0Z2F9lKU


fredericfran-gds (approved):
looks good, might to be good to add a little comment in the code itself but this is up to you.

PR #349, opened 2021-07-13
Remove obsolete IAM user for ECR pull

fredericfran-gds:
Following #347 where Concourse registry-image resource was upgraded to
support AWS credentials via EC2 metadata and IAM roles chain. We no
longer need this now obsolete IAM user for ECR pulls.

Additional work needed is to remove credentials IAM user in Concourse
secrets.


nsabri1 (approved):
Lgtm

PR #350, opened 2021-07-13
fix unsafe use of count

fredericfran-gds:
There are some potential unsafe uses of count found:
1. `alb-listener-ip-restriction-rules` module:
    this module is deleted as it is not used currently in the platform.
    Its functionality has been replaced by a WAF.

Other uses of count in the repository are for whether to apply a resource
or not and therefore, should be enough.

Ref: [trello](https://trello.com/c/iIsJYdHs/574-fix-unsafe-use-of-terraform-count-for-non-identical-objects)


terraform/modules/alb-listener-ip-restriction-rules/main.tf
```diff
@@ -21,22 +8,19 @@ resource "aws_lb_listener_rule" "allow_unrestricted_cidrs" {
...
+      values = var.fully_trusted_source_ips
     }
   }
 }
 
 resource "aws_lb_listener_rule" "restrict_all_other_ips" {
-  count = length(var.restricted_path_patterns)
-
   listener_arn = var.aws_lb_listener_arn
-  priority     = length(local.rules) + 2 + count.index
```
bilbof, line 39: I didn't realise priority is optional on ALB listener rules!

But leaving out an explicit priority on rules might lead to surprising behaviour. From [the docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb_listener_rule):

> Priority (Optional) ... Leaving it unset will automatically set the rule with next available priority after currently existing highest rule. A listener can't have multiple rules with the same priority.

The priority is important, since as there are two rules, 1) requests to a restricted path from a fully trusted source IP are forwarded on; 2) requests to a restricted path from everyone are forbidden. If we reverse the order of these two rules, requests from trusted IPs will be blocked.

terraform/modules/alb-listener-ip-restriction-rules/main.tf
```diff
@@ -1,18 +1,5 @@
-locals {
-  # Note: Needed due to an AWS quota restriction (limit of 5 conditions per rule)
-  # In future we may need to request a service quota increase of
-  # "Rules per Application Load Balancer" from AWS.
-  rules = setproduct(var.restricted_path_patterns, var.fully_trusted_source_ips)
```
bilbof, line 5: You note in the commit message using `count` is inappropriate as one rule can support multiple path patterns and  source IPs. That's true, but there is a limitation to this that the Terraform docs don't capture - and isn't caught by a `terraform plan`.

We used `count` here to handle an AWS limit: if I recall correctly, AWS sets a limit of 5 elements in the values list for `path_pattern` and `source_ip`.

I think this constraint is documented as  "Match evaluations per rule: 5" in the [AWS docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-limits.html).

We hit this constraint as we have more than 5 trusted IPs.

bilbof (changes requested):
Couple of issues with this.

I think we should go for the solution in the comment, and let clients assign the priority, since that's the dangerous part:

```
# TODO: Uses count: This may lead to unintuitive behaviour when adding/removing
# restricted_path_patterns and fully_trusted_source_ips. Impact could be
# reduced by letting client assign priority to rules.
```

terraform/modules/alb-listener-ip-restriction-rules/main.tf
fredericfran-gds: yes, I think I kept the order correct by making the `denial` rule depends on (install after) the `allow` rule.

terraform/modules/alb-listener-ip-restriction-rules/main.tf
fredericfran-gds: cool, I overlook that fact. I think if this is the case,we might be better off using a WAF.

bilbof (approved):

PR #351, opened 2021-07-13
Skip unnecessary image downloads

fredericfran-gds:
It is often not necessary to download the full images in Concourse since we only want to get the latest tag/digest to pass to ECS task definition.

The only time that we want to download the full image is when we want to extract files, e.g. static assets to be uploaded to S3 bucket. This applies only to a small subset of apps currently: frontend, static, signon and publisher.


bilbof (approved):
Great change, thanks for doing this 👍

PR #352, opened 2021-07-19
Use main branch for frontend image

bilbof:
The default branch for the frontend repo has changed to main from master.


fredericfran-gds (approved):
looks good

PR #353, opened 2021-07-19
Add dependabot for Docker images and Ruby gems

bilbof:
We'll get notified when our gems / images need to be updated.


fredericfran-gds (approved):
looks good

PR #358, opened 2021-07-20
Run smokey builds serially

bilbof:
We saw the error:
Updates were rejected because the tag already exists in the remote

This happened because two smokey jobs ran in parallel. Running the builds serially will prevent this happening again.


fredericfran-gds (approved):
good catch

PR #359, opened 2021-07-21
Push infra-concourse-task image to ECR

bilbof:
The infra-concourse-task image is used to run Concourse tasks in the deploy pipeline. The image contains the files and dependencies required by the concourse tasks such as ruby gems.

This creates a build job in the build-images pipeline, triggered by modifications to files required by the image on main.

You can see it running successfully at https://cd.gds-reliability.engineering/teams/govuk-ci/pipelines/build-images/jobs/infra-concourse-task.

Since this repository holds the Dockerfiles for multiple images, the git commit tag is prefixed with the ECR repository name: `infra-concourse-task_1.1.0-release`. The image tag will be `latest` or `1.2.3-release`, which is consistent with other images.

### DIFF
--- a/concourse/pipelines/build-images.yml
+++ b/concourse/pipelines/build-images.yml
@@ -35,6 +35,13 @@ resources:
       uri: git@github.com:alphagov/frontend.git
 
   - <<: *git-repo
+    name: infra-concourse-task-repo
+    source:
+      <<: *git-repo-source
+      uri: git@github.com:alphagov/govuk-infrastructure.git
+      paths: [ docker/infra-concourse-task, Gemfile*, .ruby-version, concourse, lib, Rakefile ]
+
+  - <<: *git-repo
     name: publisher-repo
     source:
       <<: *git-repo-source
@@ -112,6 +119,12 @@ resources:
       tag: latest
 
   - <<: *docker-image
+    name: infra-concourse-task-image
+    source:
+      <<: *docker-image-source
+      repository: infra-concourse-task
+
+  - <<: *docker-image
     name: publisher-image
     source:
       <<: *docker-image-source
@@ -184,6 +197,12 @@ resources:
     source:
       <<: *semver-version-source
       key: frontend-version
+
+  - <<: *semver-version
+    name: infra-concourse-task-version
+    source:
+      <<: *semver-version-source
+      key: infra-concourse-task-version
 
   - <<: *semver-version
     name: publisher-version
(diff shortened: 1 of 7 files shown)


sengi:
I don't understand the bit about the version tag - that seems very strange. Can you explain the rationale behind it?

bilbof:
@sengi you're right, thanks for catching that, I've explained the rationale below. I'm very open to changing this to something less strange! 😅  

The image tag has the format `infra-concourse-task:<semver>-release`.

The git commit tag has the format `<image_name>_<semver>-release`. We tag git commits so we can map commits to images.

We can remove the [`-release` suffix](https://github.com/alphagov/govuk-infrastructure/pull/359/files#diff-24a5d7f153b73273aa38613919f1d28ea1fccc61b8da73be5b0189770d88c8e6R380) from both tags. This suffix is included in image tags to ensure we only deploy images built by the build-images pipeline. We use the `variant: release` filter on [registry-image](https://github.com/concourse/registry-image-resource#source-configuration) resources in the deploy pipeline ([example](https://github.com/alphagov/govuk-infrastructure/blob/153f139f4f6661d67445501dabd8c3b286b7d303/concourse/pipelines/deploy.yml#L273)) which ensures we only use those images.

The `-release` suffix isn't needed now that we've moved to ECR. At the time we chose this format, we still used DockerHub, and repositories such as `govuk/frontend` contained images from branches and releases built by Jenkins (in the format `release_<n>`).

Another oddity is the `<image_name>_` prefix in the commit tag. I've added this prefix since there are multiple Dockerfiles in this repository in `docker/images/`, and adding `<image_name>` to commit tags lets us differentiate release tags between images. We could drop adding commit tags, use another format, or move the images to other repos, what do you think?

sengi:
> Another oddity is the <image_name>_ prefix in the commit tag. I've added this prefix since there are multiple Dockerfiles in this repository in docker/images/, and adding <image_name> to commit tags lets us differentiate release tags between images. We could drop adding commit tags, use another format, or move the images to other repos, what do you think?

It's this bit that I'm concerned about. Container registries aren't supposed to work that way. Consider what would happen if you were to pull `:latest`, for example. If there are images for different things (as opposed to different versions of the same thing) then they're gonna need distinct names otherwise things are gonna get very weird.

bilbof:
Ah sorry I misspoke there.

I incorrectly wrote the "image tag has the format `infra-concourse-task:<version>-release`" above. What I meant to say is the registry image _tag_ is `latest` or (e.g.) `1.2.3-release`. The image (and registry) name is `infra-concourse-task`.

My confusion came from the docker CLI. The `--tag` flag for `docker build` is for specifying both the name and the tag, i.e. `docker build -t name:tag`. I assumed that the name + tag was called the _tag_. 😄

See a screenshot for the registry for the new image below:

Screenshot 2021-07-23 at 15 49 33

You also mentioned you are concerned with the _git commit_ tag format being `infra-concourse-task_<version>-release`, as seen at https://github.com/alphagov/govuk-infrastructure/tags. This tag is purely cosmetic and used to help us map image tags to git commit tags. I'm completely open to changing how we do this though, if there's a better way?

sengi:
Sorry, I misread your [comment][#issuecomment-884815490]. If we're using the version tag correctly then it all looks fine to me.

sengi (approved):
Sorry forgot to actually approve

concourse/pipelines/build-images.yml
```diff
@@ -34,6 +34,14 @@ resources:
       <<: *git-repo-source
       uri: git@github.com:alphagov/frontend.git
 
+  - <<: *git-repo
+    name: infra-concourse-task-repo
+    source:
+      <<: *git-repo-source
+      uri: git@github.com:alphagov/govuk-infrastructure.git
+      paths: [ docker/infra-concourse-task, Gemfile*, .ruby-version, concourse, lib, Rakefile ]
+      branch: bilbof/build-infra-concourse-task
```
sengi, line 43: Oh don't forget to revert this

PR #360, opened 2021-07-23
Provide Content Schemas to Publishing API in ECS

bilbof:
The Publishing API app depends on [content-schemas](https://github.com/alphagov/govuk-content-schemas) being available at the path `/govuk-content-schemas`.

We use Capistrano to copy the content-schema files to EC2. As a temporary fix in ECS we have baked the content schemas into the publishing-api image. However, publishing-api releases should continue to be decoupled from content schemas versions in ECS.

In ECS we will push content-schemas to S3, trigger redeployments of publishing-api when new content-schemas versions are pushed to S3, and have publishing-api ECS Tasks copy content-schemas from S3 to the task file system for use by the publishing-api without requiring changes to the app.

This solution implements @karlbaker02's [design doc](https://docs.google.com/document/d/1nxJzdbTK__jWDFjUj9C5sH4yP0caIHhXp6fx1yKGHwE/edit) (internal-only).

Changes:

- Push content-schemas to S3 bucket in a Concourse job
- Updates to content-schemas triggers a deploy of publishing-api
- Publishing API can still be deployed without updating content-schemas
- Content Schemas will be deployed during bootstrapping but do not need to be deployed after every run-terraform job
- Added create-infrastructure job (inverse of destroy-terraform) to be run at 6am daily - triggers run-terraform, and enables us to run some jobs when bootstrapping but not on every run-terraform job.

The pipeline is available to view at https://cd.gds-reliability.engineering/teams/govuk-test/pipelines/deploy-apps-bill.
https://trello.com/c/6PvcoCAE/540-implement-content-schemas

### DIFF
--- a/terraform/deployments/govuk-publishing-platform/content_schemas.tf
+++ b/terraform/deployments/govuk-publishing-platform/content_schemas.tf
@@ -1,5 +1,18 @@
+locals {
+  content_schemas = {
+    mount_point = {
+      sourceVolume  = "content_schemas",
+      containerPath = "/govuk-content-schemas",
+      readOnly      = false
+    }
+
+    s3_bucket = "govuk-${var.govuk_environment}-${local.workspace}-content-schemas"
+    volume    = { name = "content_schemas" }
+  }
+}
+
 resource "aws_s3_bucket" "content_schemas" {
-  bucket = "govuk-${var.govuk_environment}-${local.workspace}-content-schemas"
+  bucket = local.content_schemas.s3_bucket
 
   # Unless we're in staging or production, it's okay to delete this bucket
   # even if it still contains objects.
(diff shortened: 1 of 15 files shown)


terraform/deployments/govuk-publishing-platform/content_schemas.tf
```diff
@@ -1,5 +1,18 @@
+locals {
+  content_schemas = {
+    mount_point = {
+      sourceVolume  = "content_schemas",
+      containerPath = "/govuk-content-schemas",
```
karlbaker02, line 5: The `containerPath` is set by the environment variable `GOVUK_CONTENT_SCHEMAS_PATH` if I'm not mistaken. If this is indeed the case you might want to consider something like the following which I did in my spike: https://github.com/alphagov/govuk-infrastructure/pull/314/files#diff-59692f4e1dde43d64027a70e11c6c279842b47de85a8b1a86576619c69b2ffe6R40

terraform/deployments/govuk-publishing-platform/content_schemas.tf
```diff
@@ -9,9 +22,53 @@ resource "aws_s3_bucket" "content_schemas" {
...
     aws_environment = var.govuk_environment
   }
 }
 
-# TODO: ECS must be able to cp objects in this bucket.
+module "content_schemas_container_definition" {
+  source              = "../../modules/container-definition"
+  image               = "430354129336.dkr.ecr.eu-west-1.amazonaws.com/govuk/ecs-cli:latest"
+  aws_region          = data.aws_region.current.name
+  command             = ["/bin/sh", "-c", "aws s3 cp s3://${local.content_schemas.s3_bucket} /govuk-content-schemas --recursive"]
```
karlbaker02, line 34: Same comment as here regarding the content schemas path.

karlbaker02 (commented):
Very comprehensive, you've certainly gone to town on this. Great work! 🥳

terraform/deployments/govuk-publishing-platform/content_schemas.tf
bilbof: Thanks, I've changed the env vars to reference the local variable `local.content_schemas.mount_point.containerPath`.

terraform/deployments/govuk-publishing-platform/content_schemas.tf
bilbof: I've changed this to use `$GOVUK_CONTENT_SCHEMAS_PATH`.

terraform/deployments/govuk-publishing-platform/content_schemas.tf
karlbaker02: Would you also not need to do the same thing with this path?

karlbaker02 (approved):
Looks great Bill! 💯

PR #362, opened 2021-07-26
Point test deploy-apps pipeline at default tf workspace

bilbof:
This issue was introduced accidentally in a previous commit.


karlbaker02 (approved):
👍

PR #363, opened 2021-07-26
Fix dynamic vars for tasks

bilbof:
This provides vars to task step files, fixing an error introduced in https://github.com/alphagov/govuk-infrastructure/pull/359.

When vars such as ((my-var)) are defined in a task step file i.e. my-task.yaml, then callers of that step file need to specify the value of those vars in the `vars` param for that task.

If this field is not provided you'll see an error that looks like: "failed to interpolate task config: undefined vars: my-var".

Other task step file vars such as `((docker_hub_username))` do not need `vars` specified at the pipeline level, since these are set 'globally' at the team level.

See the vars field here for more detail:
https://concourse-ci.org/jobs.html#schema.step.task-step.file

### DIFF
--- a/concourse/pipelines/deploy.yml
+++ b/concourse/pipelines/deploy.yml
@@ -787,6 +787,10 @@ jobs:
         terraform-outputs: secret_key_bases
       params:
         ASSUME_ROLE_ARN: ((concourse_deployer_role_arn))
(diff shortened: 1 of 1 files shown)


karlbaker02 (approved):
👍

PR #364, opened 2021-07-27
Permit integration AWS account to pull from prod ECR

bilbof:
This change should enable the integration concourse deployer to pull images from production ECR.


fredericfran-gds (approved):
good find

PR #365, opened 2021-07-27
Ensure apps are deployed in correct order during bootstrapping

bilbof:
### Problem and context

We've seen the `frontend-smoke-test` job fail when bootstrapping the infrastructure daily ([example]()). The error message is usually something like "503 Service Unavailable". The job fails because frontend cannot reach content-store, which is still to finish deploying when the smoke test job runs.

The root cause of the problem is that we deploy applications before their dependencies are available.

We saw the same problem when bootstrapping apps dependent on Signon. Signon generates credentials for other apps and has an API required by other apps (both during deployment and at runtime). Our solution was the `await-credentials` job. Currently we start all deploy-app jobs more or less simultaneously and have some jobs poll secretsmanager for their required secrets.

This is works okay, but it feels like a Concourse anti-pattern. A secretsmanager Concourse resource might be useful here, but that wouldn't capture all of the dependencies between apps.

### Solution

I've added the bootstrap trigger to apps, so that apps will be bootstrapped after their dependencies have been deployed.

For example, frontend won't be deployed until signon, static, and content-store are deployed (fixing the above issue):

```yaml
  - name: deploy-frontend
    plan:
    - in_parallel:
      ...
      - <<: *bootstrap-trigger
        passed: [deploy-signon,deploy-static,deploy-content-store]
```

I've also modified the pipeline [groups](https://concourse-ci.org/pipelines.html#schema.pipeline.groups) to include the dependencies of apps, which should make it easier to debug an app.

Frontend group

### Reflections

The bootstrapping process should be more robust, but it will probably take longer to complete.

The pipeline also looks messier (see whole pipeline: https://cd.gds-reliability.engineering/teams/govuk-test/pipelines/deploy-apps-bill?group=all). There are more lines (showing dependencies) which can make it harder to follow the pipeline as a whole. I think this is OK given we're going to have at least 40 more deploy-<app> jobs which will make the pipeline extremely difficult to look at in its entirety. I've added dependencies to groups to make it a bit easier to view parts of the pipeline.

### DIFF
--- a/concourse/parameters/test/deploy.yml
+++ b/concourse/parameters/test/deploy.yml
@@ -1,8 +1,8 @@
-govuk_infrastructure_branch: main
+govuk_infrastructure_branch: bilbof/deploy-dependencies
 ecr_registry_id: "172025368201"
 concourse_deployer_role_arn: arn:aws:iam::430354129336:role/govuk-concourse-deployer
 concourse_ecr_role_arn: arn:aws:iam::172025368201:role/pull_images_from_ecr_role
-workspace: default
-disable_slack_channel_alerts: false
-skip_db_migrations: false
+workspace: bill
+disable_slack_channel_alerts: true
+skip_db_migrations: true
 background_image: "https://live.staticflickr.com/7486/15537665259_a2d796c7d4_k_d.jpg"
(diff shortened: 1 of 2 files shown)


bilbof:
Closing as we're doing a big pivot to EKS and we may be doing things pretty differently.

PR #367, opened 2021-07-28
Create EKS cluster

fredericfran-gds:
This PR does the following:
1. create a `govuk` EKS cluster
2. create necessary resources for AWS ALB ingress controllers
3. install some apps using basic k8s yaml using kubectl (not yet
converted into helm chart). You can access frontend
via: https://www-origin.eks.test.govuk.digital/
4. add our team members as admin so they can use their AWS admin role to
interact with the cluster using kubectl

Ref:
1. [trello
tasks](https://trello.com/c/9PWkSvuN/589-reconfigure-govuk-infrastructure-to-support-eks)

Co-authored-by: William Flanklin <william.franklin@digital.cabinet-office.gov.uk>

### DIFF
--- a/terraform/deployments/govuk-eks/helm_auth.tf
+++ b/terraform/deployments/govuk-eks/helm_auth.tf
@@ -1,0 +1,11 @@
+provider "helm" {
+  kubernetes {
+    host                   = aws_eks_cluster.govuk.endpoint
+    cluster_ca_certificate = base64decode(aws_eks_cluster.govuk.certificate_authority[0].data)
+    exec {
+      api_version = "client.authentication.k8s.io/v1alpha1"
+      args        = ["eks", "get-token", "--cluster-name", aws_eks_cluster.govuk.name]
+      command     = "aws"
+    }
+  }
+}
--- a/terraform/deployments/govuk-eks/www_origin.tf
+++ b/terraform/deployments/govuk-eks/www_origin.tf
@@ -1,0 +1,76 @@
+# TODO: This is a placeholder service that will be replaced by `router` service.
+# The service will still be deployed later by using `kubectl` or `helm`.
+# The k8s ingress needs the service(s) to be present before creation or else
+# it will stuck in creation.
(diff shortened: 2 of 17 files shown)


terraform/deployments/govuk-eks/www_origin.tf
```diff
@@ -0,0 +1,76 @@
+# TODO: This is a placeholder service that will be replaced by `router` service.
+# The service will still be deployed later by using `kubectl` or `helm`.
+# The k8s ingress needs the service(s) to be present before creation or else
+# it will stuck in creation.
+# It is easier to have k8s ingress created in terraform since we can add a CNAME
+# to the "random" FQDN of the AWS ALB created to represent the k8s ingress.
```
sengi, line 6: We won't want to use Terraform to manage k8s objects on the application side like this, otherwise we'll lose a lot of the testability and portability benefits of k8s. A much better way to make a CNAME to the ALB is to use [external-dns](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/aws.md) and keep everything with k8s.

fredericfran-gds: cool, this adds to my knowledge. I only search whether the ingress controller could do it with annotations.

terraform/deployments/govuk-eks/helm_auth.tf
```diff
@@ -0,0 +1,11 @@
+provider "helm" {
```
kerin, line 1: Something to be aware of here - providers defined with attribute values from other provider's resources are [not predictable](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#stacking-with-managed-kubernetes-cluster-resources) and can lead to hard to reproduce issues, so the recommendation is to handle EKS and Helm/Kubernetes resources in completely separate terraform states and apply phases.

fredericfran-gds:
This is superseded by:
1. https://github.com/alphagov/govuk-infrastructure/pull/371
2. https://github.com/alphagov/govuk-infrastructure/pull/372

PR #368, opened 2021-07-30
Use Redis engine version 6.x

bilbof:
This upgrades us to use the latest available Redis engine.

3.2.10 is EOL and we are planning to upgrade to 6.x in the
current infrastructure soon.


fredericfran-gds (approved):
looks good

PR #371, opened 2021-08-03
Create GOV.UK tmp EKS Cluster

smford:
This creates a temporary EKS cluster for use by GOV.UK replatforming.
This is named govuk-tmp because there is a currently running cluster
which we are working with.  In time both of these clusters will be
cleaned up.

The PR does:
- creates an eks cluster
- creates a node group
- defines worker node sizing
- grants access to the cluster to the replatforming administrators
- grants access to the cluster to allow the k8s provider to configure
  iam role mapping to k8s users

Pair: @karlbaker02 & @smford

### DIFF
--- a/terraform/deployments/eks/k8s_auth.tf
+++ b/terraform/deployments/eks/k8s_auth.tf
@@ -1,0 +1,13 @@
+data "aws_eks_cluster_auth" "auth" {
+  name = "govuk"
+}
+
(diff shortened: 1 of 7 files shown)


karlbaker02 (approved):
👍  Looks good

smford:
Done as pair, merging.

PR #372, opened 2021-08-09
Use terraform-aws-modules/eks for the EKS cluster.

sengi:
This avoids us having to maintain a lot of fiddly resources just to create a basic cluster. E.g. it takes care of the various boilerplate IAM roles and policies which every EKS cluster needs. The module looks fairly well-maintained and customisable; didn't see any red flags on a quick read-through.

Also:

 - Fix name of state file in test.backend so it matches the module name.
 - Remove stuff that's not strictly related to creating the EKS cluster.
 - Removed no-longer-used vars, DNS remote state, k8s provider.
 - Add vars for min and max ASG size for managed node group, with reasonable defaults.

Tested: tf apply succeeds and creates a working cluster.

[Trello card](https://trello.com/c/dqG4nXn0/591)

The module has a feature to create the aws_auth configmap, but this is a bad idea because of the [race condition between TF providers](https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#stacking-with-managed-kubernetes-cluster-resources) so it's turned off here. We've yet to implement [generating the aws_auth configmap](https://trello.com/c/ONhtNSTg/610).


terraform/deployments/eks/main.tf
```diff
@@ -22,51 +17,23 @@ provider "aws" {
...
+  manage_aws_auth = false
+
+  worker_groups = [
     {
-      "Effect": "Allow",
-      "Principal": {
-        "Service": "eks.amazonaws.com"
-      },
-      "Action": "sts:AssumeRole"
+      instance_type        = "m5.xlarge"
```
fredericfran-gds, line 32: We could parameterize this one but might be worth doing later.

fredericfran-gds (approved):
looks good

terraform/deployments/eks/main.tf
sengi: Ooh, good thing you mentioned that. Turns out we already have a var for it but I didn't substitute it here. Fixed.

sengi:
Re-tested; applied cleanly (against the existing stuff still there from last time).

PR #375, opened 2021-08-11
Add EKS terraform module ADR

kerin:


sengi (approved):
SGTM. Matches what we've implemented in #372. Thanks for documenting!

PR #376, opened 2021-08-11
Add EKS split Terrafrom state ADR

kerin:


docs/architecture/decisions/0003-split-terraform-state-into-separate-aws-cluster-and-kubernetes-resource-phases.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+As Terraform providers are not fully handled as nodes in the TF plan graph it cannot be guaranteed that AWS resources will be fully created at the point that Terraform configures the `kubernetes` provider, which will lead to non-deterministic behaviour, race conditions, etc. Because of this, [Terraform providers should only use config values that are known ahead of time](https://www.terraform.io/docs/language/providers/configuration.html#provider-configuration-1).
+
+## Decision
+
+Split Terraform resources into two separate states and phases (actual names TBD):
+
+1. cluster-infrastructure - EKS cluster and associated AWS resources. No `kubernetes` provider should be defined.
+2. cluster-services - Kubernetes resources (e.g. ingress controllers) and in-cluster configuration (e.g. `aws-auth` ConfigMap), with the `kubernetes` provider configured via TF remote data sources, after `cluster-infrastructure` has been applied
```
richardTowers, line 36: remote data sources, or remote state?

Personally, I think remote state would be nicer. Maybe you could avoid having the aws provider in the cluster-services state at all (instead reading all the stuff it needs from the cluster-infrastructure's state).

docs/architecture/decisions/0003-split-terraform-state-into-separate-aws-cluster-and-kubernetes-resource-phases.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Consequences
+
+The EKS module's `manage_aws_auth` input variable must be set to `false` to prevent the module attempting to create aws auth `ConfigMap` objects.
+
+This should also prevent potential issues when tearing down an EKS cluster with deployed workloads. As ingress controllers and k8s `Service` objects commonly provision AWS load balancers themselves, Terraform does not know about these LBs and therefore does not include them in the plan graph, which can lead to Terraform attempting to delete a cluster with references to detatched AWS resources.
+
+This decision also introduces a clear boundary and separation of concerns between k8s cluster provisioning and in-cluster resources and configuration. This is an important and useful divison to enforce, as it keeps the largely generic cluster infrastructure separate from the GOV.UK-specific configuration and supporting services. This in principle provides a path to switching from an EKS cluster to any other Kubernetes cluster in any environment, as well as providing a baseline EKS cluster that could in future be used for other non-GOV.UK purposes.
```
fredericfran-gds, line 44: Not strongly related to this AD: if we delete a k8s ingress based on AWS LB, will k8s mark the ingress as deleted after all the dependent resources in AWS have been verified to be deleted? This is because k8s is based on the `desired` state.

docs/architecture/decisions/0003-split-terraform-state-into-separate-aws-cluster-and-kubernetes-resource-phases.md
```diff
@@ -0,0 +1,46 @@
...
+## Decision
+
+Split Terraform resources into two separate states and phases (actual names TBD):
+
+1. cluster-infrastructure - EKS cluster and associated AWS resources. No `kubernetes` provider should be defined.
+2. cluster-services - Kubernetes resources (e.g. ingress controllers) and in-cluster configuration (e.g. `aws-auth` ConfigMap), with the `kubernetes` provider configured via TF remote data sources, after `cluster-infrastructure` has been applied
+
+## Consequences
+
+The EKS module's `manage_aws_auth` input variable must be set to `false` to prevent the module attempting to create aws auth `ConfigMap` objects.
```
fredericfran-gds, line 40: might be worth linking to the module GitHub [repo](https://github.com/terraform-aws-modules/terraform-aws-eks)?

docs/architecture/decisions/0003-split-terraform-state-into-separate-aws-cluster-and-kubernetes-resource-phases.md
kerin: Sorry yes, that was the intent - I've clarified that, with a nod towards the potential need to access AWS info via remote data sources in some cases if necessary.

docs/architecture/decisions/0003-split-terraform-state-into-separate-aws-cluster-and-kubernetes-resource-phases.md
kerin: Yes, if an LB-backed k8s `Ingress` or `Service` object is deleted then k8s will take care of deleting the LB as well. But if a terraform-managed EKS cluster which contains an LN-backed Ingress is deleted, the Ingress object isn't explicitly deleted, so the LB is left behind in a detached state.

Splitting the state into AWS and Kubernetes resources helps prevent scenarios like that - although that said, we're more likely to run into that kind of issue with subsequent application-centric phases, e.g. Helm charts containing LB-backed Ingresses deployed via Github Actions, which Terraform will know nothing about. The overall principle of having distinct ordered phases will protect against that though, for now at least.

docs/architecture/decisions/0003-split-terraform-state-into-separate-aws-cluster-and-kubernetes-resource-phases.md
kerin: 👍

sengi (approved):
+1

Already implemented, apart from the names. I definitely like your names better than my temporary ones so I'll sort that out shortly.

PR #377, opened 2021-08-11
Move aws-auth ConfigMap generation into a separate root module.

sengi:
Terraform cannot reliably manage Kubernetes objects in the same run as creating the cluster. This is because Terraform does not handle dependencies between provider configurations: see https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs#stacking-with-managed-kubernetes-cluster-resources

We therefore need a separate deployment in order to reliably deploy k8s objects into the cluster. Configuration about the EKS cluster is passed from the first-stage deployment to the second-stage deployment via Terraform remote state.

Also clean up the variables, e.g. remove some defaults which were specific to the test environment.

Based on some earlier proof-of-concept work by @fredericfran-gds and @bilbof.

[Trello card](https://trello.com/c/dqG4nXn0/591)

#### Alternatives considered

Considered generating the aws-auth ConfigMap outside Terraform using a simple script (run terraform output from govuk-aws, pipe into transformation script to output the configmap, pipe into kubectl apply) and CDing it from Concourse. Decided against this because we're going to need this second-stage Terraform deployment anyway because we want to use the Helm provider to install/manage the cluster addons and therefore it makes sense to use the same setup for this.

#### Testing

Successfully applied the `eks` module followed by the `eks-stage2` module in the test account. Resulting aws-auth configmap looks good and a user other than the creator of the cluster was able to authenticate to the apiserver.

How to test:
```
cd eks
terraform init -backend-config test.backend -reconfigure
terraform apply -var-file=../variables/test/common.tfvars
cd ../eks-stage2
terraform init -backend-config test.backend -reconfigure
terraform apply -var-file=../variables/test/common.tfvars
aws eks update-kubeconfig --name govuk-tmp
kubectl auth can-i \* \*
kubectl describe cm aws-auth -n kube-system
```

#### Rollout

Depends on alphagov/govuk-aws#1465.

#### Caveats / follow-up work

We're currently just assigning `system:masters` to all the users. This is not good practice and we'll need to improve on that very soon. It's tolerable for just getting us going in the test environment for now.

Co-authored-by: Karl Baker <karl.baker@digital.cabinet-office.gov.uk>

### DIFF
--- a/terraform/deployments/eks-stage2/aws_auth_configmap.tf
+++ b/terraform/deployments/eks-stage2/aws_auth_configmap.tf
@@ -1,0 +1,48 @@
+# Generate the aws-auth ConfigMap, which defines the mapping between AWS IAM
+# roles and k8s RBAC. The authoritative ACLs are defined in
+# https://github.com/alphagov/govuk-aws-data/blob/master/data/infra-security/
+# and read here via Terraform remote state.
+#
+# The aws-auth ConfigMap is documented at
+# https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html
+#
+# Generally, it is unwise to manage k8s objects directly from Terraform (as
+# opposed to using Helm or kubectl or other tooling designed to work with
+# k8s). This is a rare exception to that rule of thumb.
+
+locals {
+  default_configmap_roles = [
+    {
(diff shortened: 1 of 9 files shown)


karlbaker02 (approved):
Approving having paired on this 👍 💯

PR #378, opened 2021-08-12
Add Concourse pipeline for terraforming EKS

karlbaker02:
This PR creates a pipeline for terraforming EKS, consisting of two jobs `terraform-eks-cluster` and `terraform-eks-cluster-addons`, which terraforms the EKS cluster and the addons respectively. This is necessary for the reasons as detailed in commit c228db6.

We also change the way we assume the `govuk-concourse-deployer` role.  Instead of configuring the role once per Terraform provider per top-level module (`deployments/*`), we do the assume-role as part of the terraform-apply.sh script (by running aws sts assume-role). This results in less overall complexity. (Config for the k8s TF provider to assume a role is different from the equivalent config for the AWS provider.)

An alternative would be to keep all the existing assume-role configs and pass `--role-arn` to the `aws` command from the k8s TF provider config. Our approach however means we have one place and one way of assuming the role, rather than two different ways and multiple places (one per provider per top-level module).

We also add the `govuk-concourse-deployer` role to the `aws-auth` configmap so that Concourse can apply changes to the Kubernetes cluster via the Kubernetes Terraform provider. This is only strictly necessary when Concourse is modifying an existing cluster _not_ created by Concourse (i.e. not created by the `govuk-concourse-deployer` role), but we're including this for robustness and to make the required permissions explicit.

We're running Terraform in automation mode per the best practices described in https://learn.hashicorp.com/tutorials/terraform/automate-terraform

Pair: @sengi 

Trello: https://trello.com/c/dqG4nXn0

Tested: successfully updates the existing govuk-tmp cluster (see [Concourse pipeline](https://cd.gds-reliability.engineering/teams/govuk-test/pipelines/eks)).

Screenshot 2021-08-12 at 14 37 45

### DIFF
--- a/concourse/parameters/test/eks.yml
+++ b/concourse/parameters/test/eks.yml
@@ -1,0 +1,3 @@
+aws_region: eu-west-1
+concourse_deployer_role_arn: arn:aws:iam::430354129336:role/govuk-concourse-deployer
+govuk_infrastructure_branch: main
--- a/concourse/pipelines/eks.yml
+++ b/concourse/pipelines/eks.yml
@@ -1,0 +1,84 @@
+---
+resource_types:
+  - name: slack-notification
+    type: docker-image
+    source:
+      repository: cfcommunity/slack-notification-resource
+      tag: latest # TODO - don't use latest (once we've worked out a policy for third party images)
+      username: ((docker_hub_username))
+      password: ((docker_hub_authtoken))
+
+resources:
+  - icon: github
+    name: govuk-infrastructure
+    source:
+      branch: ((govuk_infrastructure_branch))
+      uri: https://github.com/alphagov/govuk-infrastructure
+    type: git
+
+  - name: deploy-slack-channel
+    type: slack-notification
+    icon: bell-ring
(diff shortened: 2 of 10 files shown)


sengi:
> Thanks for this work.
> 
> Did you consider using the flag `--role-arn` when setting up the k8s provider [here](https://github.com/alphagov/govuk-infrastructure/blob/c1ab48d2cab8893ad9d5b40e8112ba98827b08bf/terraform/deployments/eks-stage2/k8s_auth.tf#L11)?
> 
> Just trying to see what is the main benefit of setting the credentials using sts and having to deal with jq.

Thanks - we hadn't actually twigged that we could do this(!), but we think the solution we came up with is less complex and easier to understand than duplicating the config lots of times in two different ways for the two providers in every top-level module. We've updated the commit message and added a comment to signpost this.

sengi (approved):
@karlbaker02 and I paired on this and we've also reviewed each other's contributions (quite thoroughly!). Just had final read-through and we think it looks good so I'm approving 👍

PR #381, opened 2021-08-13
Add ADR for choice of EKS node group

kerin:


sengi (approved):
+1 for managed node group. Definitely the safest option to get started with.

PR #385, opened 2021-08-16
Add additional IPs to office_cidr_list list

fredericfran-gds:
GDS `Bring Your Own Device` network has 2 new IPs:
1. 213.86.153.211
2. 213.86.153.231

that we want to add to the list of `office_cidr_list` so that these devices can continue to have access to the GOV.UK services.

Ref:
1. [trello task](https://trello.com/c/0La4yrYZ/613-add-byod-vpn-egress-ips-to-allowlists)


sengi (approved):
Didn't know we still had this duplicated here. Thanks for updating it. Maybe stick a TODO to nuke it and use remote-state though?

PR #387, opened 2021-08-16
Associate admin users with `cluster-admins` group

karlbaker02:
This PR associates all current admin users with the existing in-built `cluster-admin` role via a new ClusterRoleBinding. By making this association, we can remove the direct association between these users and the `system:masters` group, meaning that we have greater visibility over users' permissions as well as laying some groundwork for when we more tightly scope user roles.

Trello: https://trello.com/c/dqG4nXn0


sengi (approved):
Cool

PR #388, opened 2021-08-16
Add ADR for use of Helm for application packaging

kerin:


fredericfran-gds (approved):
looks great, thanks as usual for providing so much useful context to support the ADR.

sengi (commented):
+1 for Helm - it's so widely used that it'd be hard to argue for starting with anything else.

Agree that text templates and package versioning are potential tarpits to be mindful of.

PR #389, opened 2021-08-16
Remove the Concourse deployer IAM role from the aws-auth ConfigMap.

sengi:
Including this was necessary in order for Concourse to manage objects in a cluster which Concourse didn't originally create, but it introduces a risk of locking ourselves out of the cluster entirely if we break the aws-auth ConfigMap.

If we ever did need Concourse to "adopt" a cluster which it didn't create, we could still do this by adding its role to the ConfigMap at that time. Unless/until that's needed, it's safer to omit it so that we can always access the cluster via the cluster creator IAM role (which from this point on will be the govuk-concourse-deployer role).

See also this Amazon knowledge centre article which mentions the issue: https://aws.amazon.com/premiumsupport/knowledge-center/amazon-eks-cluster-access/

### DIFF
--- a/terraform/deployments/eks-stage2/aws_auth_configmap.tf
+++ b/terraform/deployments/eks-stage2/aws_auth_configmap.tf
@@ -19,11 +19,7 @@ locals {
     },
   ]
 
-  concourse_worker_role = {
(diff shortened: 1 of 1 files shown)


fredericfran-gds (approved):
good work to add resilience to the platform, looks good to me.

PR #390, opened 2021-08-16
Add ADR for EKS application logging

kerin:


sengi (approved):
+1 for fluentbit if we can get away with it (sounds like it's much more capable than when I last looked a few years ago)

Is 'EDR' a typo in the PR title / commit message?

bilbof (approved):
SGTM, thanks! 👍

kerin:
> +1 for fluentbit if we can get away with it (sounds like it's much more capable than when I last looked a few years ago)

I've not yet used it personally, so this is a bit speculative on my part - it is a direct drop-in replacement for Fluentd though and there'd be very minimal friction in changing horses, so it seems well worth a go.

> Is 'EDR' a typo in the PR title / commit message?

It is!

PR #391, opened 2021-08-17
add readonly users to aws-auth config map

fredericfran-gds:
This commit adds `readonly` users (as defined in
[govuk-aws-data](https://github.com/alphagov/govuk-aws-data)) are added
to the cluster.

These new users are mapped to the k8s group `readonly` with the default
[view](https://kubernetes.io/docs/reference/access-authn-authz/rbac/)
clusterrole is associated to that group.

Since Concourse CI in the `test` environment needs a user to verify
commits via terraform plans, we added a new readonly user
`govuk-ci-concourse` to the cluster in the `test` environment.

We add 2 additional privileges to the `readonly` k8s group: read
clusterroles and clusterrolebindings since this necessary for
`govuk-ci-concourse` to be able to plan the `cluster-services`
deployment.

Ref:
1. [trello card](https://trello.com/c/ONhtNSTg/406-iam-eks-configure-fulladmin-and-readonly-access-levels)

### DIFF
--- a/terraform/deployments/cluster-services/aws_auth_configmap.tf
+++ b/terraform/deployments/cluster-services/aws_auth_configmap.tf
@@ -27,6 +27,19 @@ locals {
       groups   = ["cluster-admins"]
     }
   ]
+
+  ci_planner_username_and_rolearn = {
+    "govuk-ci-concourse" = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/govuk-ci-concourse"
+  }
+
+  readonly_roles_and_arns = merge(data.terraform_remote_state.infra_security.outputs.user_roles_and_arns, local.ci_planner_username_and_rolearn)
+  readonly_configmap_roles = [
+    for user, arn in local.readonly_roles_and_arns : {
+      rolearn  = arn
+      username = user
+      groups   = ["readonly"]
+    }
+  ]
 }
 
 resource "kubernetes_config_map" "aws_auth" {
@@ -41,6 +54,7 @@ resource "kubernetes_config_map" "aws_auth" {
       distinct(concat(
         local.default_configmap_roles,
         local.admin_configmap_roles,
+        local.readonly_configmap_roles,
       ))
     )
   }
@@ -58,6 +72,50 @@ resource "kubernetes_cluster_role_binding" "cluster_admins" {
   subject {
     kind      = "Group"
     name      = "cluster-admins"
+    api_group = "rbac.authorization.k8s.io"
+  }
+}
+
+resource "kubernetes_cluster_role_binding" "cluster_readonly" {
+  metadata {
+    name = "cluster-readonly"
+  }
+  role_ref {
+    api_group = "rbac.authorization.k8s.io"
+    kind      = "ClusterRole"
+    name      = "view"
+  }
+  subject {
+    kind      = "Group"
+    name      = "readonly"
+    api_group = "rbac.authorization.k8s.io"
+  }
+}
+
+resource "kubernetes_cluster_role" "read_crs_and_crbs" {
+  metadata {
+    name = "read-crs-and-crbs"
+  }
+
+  rule {
+    api_groups = ["rbac.authorization.k8s.io"]
+    resources  = ["clusterrolebindings", "clusterroles"]
+    verbs      = ["get", "list", "watch"]
+  }
+}
+
+resource "kubernetes_cluster_role_binding" "read_crs_and_crbs" {
+  metadata {
+    name = "read-crs-and-crbs"
+  }
+  role_ref {
+    api_group = "rbac.authorization.k8s.io"
+    kind      = "ClusterRole"
+    name      = kubernetes_cluster_role.read_crs_and_crbs.metadata.0.name
+  }
+  subject {
+    kind      = "Group"
+    name      = "readonly"
     api_group = "rbac.authorization.k8s.io"
   }
 }


sengi (commented):
I'd much prefer to keep it simple for now until we have a definite use case for this "planners" group being separate. It's fine for the readonly role to be able to see CRs and CRBs so that you can use it for your CI job.

terraform/deployments/cluster-services/aws_auth_configmap.tf
```diff
@@ -61,3 +76,47 @@ resource "kubernetes_cluster_role_binding" "cluster_admins" {
...
+    name      = "view"
+  }
+  subject {
+    kind      = "Group"
+    name      = "readonly"
+    api_group = "rbac.authorization.k8s.io"
+  }
+}
+
+resource "kubernetes_cluster_role" "additional_cluster_readonly" {
```
sengi, line 96: Can we call this something more informative? (and update the k8s object name to match)
```suggestion
resource "kubernetes_cluster_role" "read_crs_and_crbs" {
```

terraform/deployments/cluster-services/aws_auth_configmap.tf
```diff
@@ -27,6 +27,20 @@ locals {
       groups   = ["cluster-admins"]
     }
   ]
+
+  # only the test environment has a CI planner user to check commits before merge into main
```
sengi, line 31: Why is this specific to the test account? Surely we'd want CI to be able to show the plan for all the environments?

terraform/deployments/cluster-services/variables.tf
```diff
@@ -7,3 +7,8 @@ variable "cluster_infrastructure_state_bucket" {
   type        = string
   description = "Name of the S3 bucket for the cluster-infrastructure module's Terraform state. Must match the name of the bucket specified in the backend config file."
 }
+
+variable "govuk_environment" {
```
sengi, line 11: It doesn't seem right to me that this module would need to know which "environment" it's in.

terraform/deployments/cluster-services/aws_auth_configmap.tf
sengi: In fact even if it didn't need to talk to the other account, there's still no need to make this environment-specific. The actual access is controlled via IAM - this is just the mapping, which is fine to exist in all the clusters.

terraform/deployments/cluster-services/aws_auth_configmap.tf
fredericfran-gds: yes, I can do that.

terraform/deployments/cluster-services/aws_auth_configmap.tf
fredericfran-gds: currently, the concourse planner is in test environment only: https://github.com/alphagov/govuk-infrastructure/blob/df299404ebb36f99c1e1c7671b5696ae58e88dc6/terraform/deployments/concourse-iam/main.tf#L70

The concourse CI planner that checks commits runs in the test environment only.

I didn't want to add users that will "never" be used in the other environments.

sengi: Please keep the environments the same. This module shouldn't need to know the environment.

fredericfran-gds: Ok, I can do that.

terraform/deployments/cluster-services/aws_auth_configmap.tf
```diff
@@ -61,3 +75,47 @@ resource "kubernetes_cluster_role_binding" "cluster_admins" {
...
+  subject {
+    kind      = "Group"
+    name      = "readonly"
+    api_group = "rbac.authorization.k8s.io"
+  }
+}
+
+resource "kubernetes_cluster_role" "read_crs_and_crbs" {
+  metadata {
+    name = "read_crs_and_crbs"
```
sengi, line 97: You have dashes in the object name for the rolebinding but underscores here - best keep them consistent.
```suggestion
    name = "read-crs-and-crbs"
```

terraform/deployments/cluster-services/aws_auth_configmap.tf
```diff
@@ -61,3 +75,47 @@ resource "kubernetes_cluster_role_binding" "cluster_admins" {
...
+  rule {
+    api_groups = ["rbac.authorization.k8s.io"]
+    resources  = ["clusterrolebindings", "clusterroles"]
+    verbs      = ["get", "list", "watch"]
+  }
+}
+
+resource "kubernetes_cluster_role_binding" "read_crs_and_crbs" {
+  metadata {
+    name = "read_crs_and_crbs"
```
sengi, line 109: (same suggestion as above, for convenience)
```suggestion
    name = "read-crs-and-crbs"
```

sengi (approved):
Just a tiny naming inconsistency to fix please - otherwise looks great. Thanks.

terraform/deployments/cluster-services/aws_auth_configmap.tf
fredericfran-gds: ok will do

terraform/deployments/cluster-services/aws_auth_configmap.tf
fredericfran-gds: ok will do

PR #394, opened 2021-08-17
Rename steps within EKS pipeline

karlbaker02:
This PR renames the steps (which apply the Terraform) within the `terraform-cluster-infrastructure` and `terraform-cluster-services` jobs of the `eks` pipeline to be called `terraform-apply`. This removes the duplication between the job and step name that currently exists and also makes our naming consistent with the existing `deploy` pipeline that we have been using to deploy to ECS.

### DIFF
--- a/concourse/pipelines/eks.yml
+++ b/concourse/pipelines/eks.yml
@@ -39,11 +39,11 @@ jobs:
       trigger: true
       passed:
       - update-pipeline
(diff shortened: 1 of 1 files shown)


bilbof (approved):
LGTM

karlbaker02:
Thanks @bilbof 👍

PR #395, opened 2021-08-17
Document how to run Terraform and connect to the cluster.

sengi:
Replaces the old ECS docs.

Tested the steps on my workstation by copy/paste.

Probably easiest to review this by viewing the new file rather than reading the diff.

### DIFF
--- a/terraform/docs/applying-terraform.md
+++ b/terraform/docs/applying-terraform.md
@@ -1,140 +1,44 @@
 # Applying Terraform
 
-> **Note:** This document may become outdated. Please update this document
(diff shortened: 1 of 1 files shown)


karlbaker02 (approved):
Looks good & paired when writing, so approving 👍

fredericfran-gds (commented):
Thank you for updating the docs.

I have 2 comments:
1. most times you use `gds aws` etc and one time you use `eval` with kubectl, maybe converge one method.
2. there is a lack of description with the commands, especially with `aws eks update-kubeconfig`

PR #397, opened 2021-08-20
Install the AWS Load Balancer Controller as part of cluster turnup.

sengi:
Add the [AWS Load Balancer Controller](https://docs.aws.amazon.com/eks/latest/userguide/aws-load-balancer-controller.html) and its prerequisites to the Terraform cluster turnup automation.

- Enable AWS [IAM Roles for Service Accounts][IRSA] by enabling creation of the OIDC provider in the `eks` module.
- Add the boilerplate IAM policy for the LB controller (which Amazon  inexplicably hasn't made a managed policy).
- Add outputs for the IAM role ARN and corresponding k8s service account name, passed via remote-state to the cluster-services module where we deploy the Helm chart for the LB controller.
- Remove the AWS provider from the cluster-services module (per [adr-3]).
- Move provider configs into main.tf and remove some dead code.
- Add the Helm provider to the cluster-services module.
- Install the AWS Load Balancer Controller Helm chart.

We're using IRSA because it's Amazon's recommended way of granting the LB controller access to AWS APIs. The old way was to authorise the nodes, but then any pod running on those nodes could change ALB/NLB configs.

We pass the ServiceAccount name and [IRSA] annotation to the Helm chart so that it can create the ServiceAccount along with everything else, rather than adding another dodgy Terraform-managed k8s object.

[Trello card](https://trello.com/c/hzCu1FLy/618)

Pair: @karlbaker02 

### Testing

- `tf init -backend-config test.backend -reconfigure && tf apply -var-file=../variables/test/common.tfvars` succeeded for both root modules (cluster-infrastructure and cluster-services).
- Deployed the [EchoServer] example (updated to use `apiVersion: networking.k8s.io/v1` rather than `extensions/v1beta1`), confirmed that pods were receiving traffic, ALB was created and configured as expected and the info logs from the controller looked good. The ALB was also deleted as expected when deleting the example Ingress.

### Not in this PR
- We manually added tags to the existing govuk-aws subnets which the `govuk` cluster is temporarily using, in order to test the [subnet discovery feature of the LB controller](https://docs.aws.amazon.com/eks/latest/userguide/alb-ingress.html). We won't be using those subnets for long; the cluster will soon have its own dedicated subnets so we'll label those via Terraform when we make that change (soon).

[IRSA]: https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html
[adr-3]: https://github.com/alphagov/govuk-infrastructure/blob/main/docs/architecture/decisions/0003-split-terraform-state-into-separate-aws-cluster-and-kubernetes-resource-phases.md
[EchoServer]: https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/examples/echo_server/


terraform/deployments/cluster-services/main.tf
```diff
@@ -1,23 +1,38 @@
...
-      source  = "hashicorp/aws"
-      version = "~> 3.33"
-    }
+provider "kubernetes" {
+  host                   = data.terraform_remote_state.cluster_infrastructure.outputs.cluster_endpoint
+  cluster_ca_certificate = base64decode(data.terraform_remote_state.cluster_infrastructure.outputs.cluster_certificate_authority_data)
+  exec {
+    # aws eks get-token (at least up to aws-cli 2.2.30) still uses the v1alpha1 API.
+    # Amazon fluffed the migration to v1beta1 and had to revert it:
+    # https://github.com/aws/aws-cli/pull/6309
```
bilbof, line 19: Nit, I wonder if this comment could be rewritten as a TODO? Perhaps

```suggestion
    # TODO: use v1beta1 API once `aws eks get-token` supports this
    # https://awscli.amazonaws.com/v2/documentation/api/latest/reference/eks/get-token.html
```

terraform/deployments/cluster-services/aws_lb_controller.tf
```diff
@@ -0,0 +1,23 @@
...
+# The AWS IAM resources for the AWS LB Controller are in
+# ../cluster-infrastructure/aws_lb_controller_iam.tf. See
+# https://github.com/alphagov/govuk-infrastructure/blob/main/docs/architecture/decisions/0003-split-terraform-state-into-separate-aws-cluster-and-kubernetes-resource-phases.md#decision for rationale.
+
+resource "helm_release" "aws_lb_controller" {
+  name       = "aws-load-balancer-controller"
+  repository = "https://aws.github.io/eks-charts"
+  chart      = "aws-load-balancer-controller"
+  version    = "1.2.6" # TODO: Dependabot or equivalent so this doesn't get neglected.
+  namespace  = "kube-system"
```
bilbof, line 13: Forgive my ignorance, is `kube-system` appropriate for user-defined objects, or is this LB controller part of the Kubernetes system?

bilbof (approved):
Thanks for a really detailed PR description, this helped a lot!

You'll want k8s experts to review this, but from a k8s beginner perspective this all looks really good. Thanks for adding helpful inline comments.

terraform/deployments/cluster-services/aws_lb_controller.tf
kerin: @bilbof that is a good question, and the picture on that seems a bit unclear. The [Kubernetes docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#viewing-namespaces) state that`kube-system` is intended for objects created by Kubernetes itself, which implies that we shouldn't deploy anything into it - but then [Amazon's docs](https://docs.aws.amazon.com/eks/latest/userguide/aws-load-balancer-controller.html) say to deploy it to `kube-system`, as do the [AWS LB Controller docs](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/deploy/installation/).

So, not clear at all really, and I think you can argue it either way, but right now I'd say that we should use a separate namespace for supporting services like this, `cluster-services` or similar, as we'll have other components like Prometheus, Fluentbit etc. which should logically be organised into the same NS which don't seem to naturally belong in `kube-system` - that way we should avoid any unforeseen consequences of having third-party objects in `kube-system`, and allow us to apply RBAC controls to the `cluster-services` namespace if necessary, which might be a problem with `kube-system`.

bilbof: That makes sense, thanks for explaining!

sengi: Great question. kube-system is basically for cluster API machinery. Objects which are managed by Kubernetes itself (i.e. the master node) certainly belong in there (for example coredns, kube-proxy, cloud provider's CNI plugin and so on). Deploying stuff into kube-system is definitely not something you'd do as a user of the cluster. In this case it's a bit of a fuzzy area because the LB controller is a basic service that's pretty much an essential part of the cluster, and it's the cloud provider's supported ingress controller addon. Both the [maintainer SIG's install guide](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/deploy/installation/) and [Amazon's](https://docs.aws.amazon.com/eks/latest/userguide/aws-load-balancer-controller.html) say to use `kube-system`, and it seems reasonable in this case - so we went with that.

sengi: I'll be surprised if we end up with anything else which warrants going in kube-system, btw.

terraform/deployments/cluster-services/main.tf
sengi: Yeah that's a good shout, thanks. Done.

terraform/deployments/cluster-services/aws_lb_controller.tf
sengi: FWIW I can't think of any way that it could cause any practical problem by being in kube-system. I'd probably just be guided by the principle of least surprise on this one, so both manuals suggesting kube-system suggests that's a reasonable choice (and it's the first place I'd look if I wanted to grab the logs for the cloud provider's ingress controller). 

If there are strong feelings about this then I'm happy to bung it in its own namespace and re-test. It'd be another thing for people to have to discover if they're debugging though (because all the offical docs have kube-system in their examples).

bilbof: SGTM, I'm happy for us to follow the AWS docs and keep this in the `kube-system` namespace.

kerin: No, no strong feelings on this from me - as I say, this can clearly be argued either way. So yeah, stick with `kube-system` - if there are any unintended consequences then it should be trivial to re-deploy to a different namespace.

terraform/deployments/cluster-infrastructure/aws_lb_controller_iam.tf
```diff
@@ -0,0 +1,263 @@
...
+  # have to strip the "https://" when we construct the trust policy
+  # (assume-role policy). The subscript [0] is not an off-by-one; it really
+  # does means the first capturing group and not the entire matched pattern.
+  cluster_oidc_issuer = regex("https://(.*)", module.eks.cluster_oidc_issuer_url)[0]
+}
+
+resource "aws_iam_role" "aws_lb_controller" {
+  name        = "AmazonEKSLoadBalancerControllerRole"
+  description = "Role for the AWS Load Balancer Controller. Corresponds to ${local.aws_lb_controller_service_account_name} k8s ServiceAccount."
+  tags        = {} # TODO: tagging
```
kerin, line 240: Just a note on this - should we make use of [provider default tags](https://www.hashicorp.com/blog/default-tags-in-the-terraform-aws-provider) so that everything gets consistent tags?

terraform/deployments/cluster-services/aws_lb_controller.tf
sengi: The equivalent controller in GKE lives in kube-system FWIW (as did all our official cluster addons), though those were all properly managed by the product — we didn't make the customers install them themselves with a handy [30+ step manual process](https://docs.aws.amazon.com/eks/latest/userguide/aws-load-balancer-controller.html) 😛).

terraform/deployments/cluster-infrastructure/aws_lb_controller_iam.tf
```diff
@@ -0,0 +1,263 @@
+# IAM role and policy to allow the AWS Load Balancer Controller to talk to AWS
```
kerin, line 1: General question on structure - should we define everything as local TF modules? I personally quite like that approach as it means [you can see the high-level relationships](https://github.com/ministryofjustice/analytics-platform-ops/blob/main/infra/terraform/platform-base/main.tf) between logical components within the main.tf file. It does mean more work as you have to define more variables and outputs, but that does also mean you end up with something like an actual defined interface for each component.

kerin (approved):
LGTM 👍 - just got a couple of questions on Terraform structure, not specific to this PR.

terraform/deployments/cluster-infrastructure/aws_lb_controller_iam.tf
sengi: Hrm that's an interesting one. I do see your point of view, in fact I started out with that rule of thumb on the ECS project (which was at least an order of magnitude more complex so definitely some use of submodules) but I've since changed my mind based on that experience. My advice would be to use submodules only where TF code actually needs to be reused, because the verbosity and complexity that the module calls introduce is very expensive to maintain and really hurts readability.

I'd love to make things more modular and testable in principle, but the HCL/Terraform language imposes such a high maintenance penalty on doing that that I've come to the conclusion that it's only worth it where reuse is needed, rather than for structure. Thoughtful use of separate files within the same module can achieve most of the same benefits of surfacing the structure and responsibilities without the maintenance penalty.

sengi: How strongly do you feel about this one? 😅

kerin: Not that strongly really, certainly not at this stage - this just comes from personal experience of having a fairly large terraform project where it became really hard to see what was going on because so much of terraform's behaviour is implicit, and moving to an explicit input/output structure with modules really helped introduce some sanity. But, we're a long way away from that scenario at the moment, plus I don't want to impose my opinions on everyone - this kind of thing is a team decision in my mind.

terraform/deployments/cluster-infrastructure/aws_lb_controller_iam.tf
sengi: Oh that's a good call, thanks. I'll raise another PR for that shortly in case people have thoughts about the tag values or whatnot.

sengi: Remove the empty tags maps and TODOs for now since I suspect we won't need to add anything here if we're setting defaults at the provider.

terraform/deployments/cluster-infrastructure/aws_lb_controller_iam.tf
sengi: Totally happy to keep an eye on how this goes and refactor if it starts getting hairy again.

Maybe one day if I get a bunch of spare time I'll write a Dhall frontend for Pulumi 😁

kerin: ...only thing I'd add is that in the absence of locally defined modules we should probably all be clear on how we structure things - I'd vote for exactly what you've got here, all resources for a given logical component grouped together in either a single file or multiple files with a common prefix (`aws_lb_controller_iam.tf`, `aws_lb_controller_ec2.tf` if there was EC2 stuff for example), rather than a resource type-based structure like `iam.tf`, `s3.tf`, where any notion of logical components goes right out the window.

sengi: 💯

terraform/deployments/cluster-infrastructure/aws_lb_controller_iam.tf
sengi: #399 is my attempt at setting some default tags.

PR #402, opened 2021-08-24
Give the EKS cluster its own subnets and associated resources.

sengi:
The VPC itself and the Internet gateway are still defined in alphagov/govuk-aws for now, since the current plan is to use the existing VPC and move/import it here after the migration to Kubernetes.

This follows the guidance in [Cluster VPC considerations](https://docs.aws.amazon.com/eks/latest/userguide/network_reqs.html) in the EKS user guide.

In each of three AZs, we have a public /28 for the control plane (i.e. the master ENIs) as recommended, a public /24 for Internet-facing load balancers and a private /22 for nodes and pods.

Separating the node and pod subnets is possible but involves some [hairy configuration of Amazon's CNI plugin][cni-custom-network], so it doesn't seem worth it without a specific need.

Also some small fixes while we're there (see commit messages).

[cni-custom-network]: https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html

[Trello card](https://trello.com/c/mtzlqNa5/622-use-separate-subnets-for-the-eks-cluster)

#### Testing
* Terraform plan/apply of cluster-infrastructure and cluster-services works in a separate workspace (with a different cluster name).
* Resulting cluster comes up and nodes are able to join.
* AWS LB Controller ingress still works.

#### Alternatives considered

Looked at using a library module to create all this lot, but the main one seems to be terraform-aws-modules/terraform-aws-vpc and it doesn't have an option to use an existing VPC.

Was tempted to factor the {subnet, route_table, route_table_association, default route} set of resources into a local module, but resisted that temptation for now as it's probably premature abstraction given that we know we're going to be importing the remaining bits of VPC gubbins from govuk-aws at some point and given how much friction TF causes in refactoring anything that involves moving resources into/out of/between modules.

(I added a TODO mentioning both of the above.)

### DIFF
--- a/terraform/deployments/cluster-infrastructure/variables.tf
+++ b/terraform/deployments/cluster-infrastructure/variables.tf
@@ -14,6 +14,21 @@ variable "cluster_name" {
   default     = "govuk"
 }
 
+variable "eks_control_plane_subnets" {
+  type        = map(object({ az = string, cidr = string }))
+  description = "Map of {subnet_name: {az=<az>, cidr=<cidr>}} for the public subnets for the EKS cluster's apiserver."
+}
+
+variable "eks_private_subnets" {
+  type        = map(object({ az = string, cidr = string }))
+  description = "Map of {subnet_name: {az=<az>, cidr=<cidr>}} for the private subnets for the EKS cluster's nodes and pods."
+}
+
+variable "eks_public_subnets" {
+  type        = map(object({ az = string, cidr = string }))
+  description = "Map of {subnet_name: {az=<az>, cidr=<cidr>}} for the public subnets where the EKS cluster will create Internet-facing load balancers."
+}
+
 variable "workers_instance_type" {
   type        = string
   description = "Instance type for the managed node group."
--- a/terraform/deployments/cluster-infrastructure/vpc.tf
+++ b/terraform/deployments/cluster-infrastructure/vpc.tf
@@ -1,0 +1,161 @@
+# vpc.tf manages the subnets for the EKS cluster and their associated
+# paraphernalia such as NAT gateways and route tables. The VPC itself is
+# defined in https://github.com/alphagov/govuk-aws/ while we transition to
+# Kubernetes.
+#
+# https://docs.aws.amazon.com/eks/latest/userguide/network_reqs.html
+
+# TODO: Once the EC2 environment is gone, move the aws_vpc from
+# alphagov/govuk-aws to here.
+
+# TODO: Consider factoring out the (subnet, route_table,
+# route_table_association, route) resources into a local module (or maybe two,
+# for public/private), perhaps after we've imported the VPC resource from
+# govuk-aws. Also consider using terraform-aws-modules/terraform-aws-vpc to
+# replace the whole lot. (That module makes some dubious use of count() but
+# it's ok because we're very unlikely to add or remove availability zones.)
+
+locals {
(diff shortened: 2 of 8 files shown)


terraform/deployments/cluster-infrastructure/variables.tf
```diff
@@ -3,11 +3,32 @@ variable "govuk_aws_state_bucket" {
...
+}
+
 variable "cluster_log_retention_in_days" {
   type        = number
   description = "Number of days to retain cluster log events in CloudWatch."
 }
 
+variable "eks_control_plane_subnets" {
+  type        = map(object({ az = string, cidr = string }))
+  description = "Map of {subnet_name: {az=<az>, cidr=<cidr>}} for the public subnets for the EKS cluster's apiserver."
```
kerin, line 19: This is probably something to address later, but I think we probably want to consider keeping the API server endpoint private, given the potential consequences of a leaked EKS credential or security issue with k8s/EKS, or whether relying on CIDR/IP whitelisting is good enough - that obviously raises the question of what a non-public network route to the API endpoint would be (VPN, bastion, etc.). Similarly we probably should make use of the private API endpoint for in-cluster API calls, if we're not already.

([public/private API endpoint docs](https://docs.aws.amazon.com/eks/latest/userguide/cluster-endpoint.html))

kerin (approved):
LGTM 👍 - just one note on API server access, which is something for another day/further discussion

terraform/deployments/cluster-infrastructure/variables.tf
sengi: Yeah those are both good points. I've made a note to look into both (private apiserver and whether we need to do anything about the routing of traffic between nodes and apiserver).

My hunch is that with out current AWS setup, making the apiserver private would just move the problem rather than provide any actual security improvement - e.g. we'd end up sshing into a bastion VM or something ugly like that. We could stick an IP allowlist on the apiserver (and we'll almost certainly need to do so for prod - that'd be roughly equivalent to what we have now).

sengi: Private apiserver endpoint wasn't enabled before but it is now - thanks for picking up on that 👍

sengi:
Ok I think this is RFAL, at long last!

terraform/deployments/cluster-infrastructure/vpc.tf
```diff
@@ -0,0 +1,151 @@
...
+# Public subnets and associated resources. The public subnets are used by
+# controllers in the cluster (such as aws-load-balancer-controller) for
+# creating Internet-facing load balancers.
+
+resource "aws_subnet" "eks_public" {
+  for_each          = var.eks_public_subnets
+  vpc_id            = data.terraform_remote_state.infra_vpc.outputs.vpc_id
+  cidr_block        = each.value.cidr
+  availability_zone = each.value.az
+  tags = {
```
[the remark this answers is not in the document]
sengi: Thankfully the repetitive merge() pattern is no longer needed, except on ASGs (and the eks module sorta takes care of that, albeit in a slightly funky way).

sengi:
Rebased to squash a very small fix: apply the route create timeout workaround consistently to all routes, not just the IGW one. Should make `tf apply` slightly less flaky and brings us into line with what terraform-aws-modules/terraform-aws-vpc does.

PR #415, opened 2021-09-01
Install external-secrets and AWS SM ClusterSecretStore

bilbof:
This integrates the Kubernetes cluster with AWS SecretsManager.
It enables developers to create a SecretsManager secret and a
corresponding Kubernetes ExternalSecret, so that AWS SM secrets
can be referenced in Helm deployments.

I've implemented https://external-secrets.io/ as described in
ADR 8. The external-secrets-operator handles creating CRDs for
mapping AWS SM resources to Kubernetes objects, and syncing
Kubernetes secrets with AWS SM secrets.

The ClusterSecretStore should be created by cluster-services.
However, since this depends on the CRDs created by external-secrets
we must install external-secrets first. I looked at using the
kubernetes_manifest (alpha, experimental Terraform provider feature)
but it was unreliable.
I also tried creating a chart for both installing external-secrets
and the ClusterSecretStore. This didn't work since Helm couldn't
handle creating dependencies first. It turns out the Helm community
does not have a consensus for how to handle CRDs. I've decided to
go for the second option detailed in the Helm docs [1]: create
two charts and install them sequentially, achieved with Terraform
depends_on (later could be Concourse if we don't want to use TF).

Still to implement/out of scope:
* managing secret lifecycle in AWS SecretsManager
* using secrets in app deployments

Trello: https://trello.com/c/YE7PGg4y/593-deploy-external-secrets-to-cluster

[1]: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-2-separate-charts


terraform/deployments/cluster-infrastructure/external_secrets_iam.tf
```diff
@@ -0,0 +1,50 @@
+# IAM role and policy to enable the k8s external-secrets operator to talk to
+# AWS APIs to access and manage SecretsManager secrets.
+#
+# The k8s side of the autoscaler is in
+# ../cluster-services/external_secrets.tf.
+
+locals {
+  external_secrets_service_account_namespace = "secrets"
```
kerin, line 8: So there's a question here around which namespaces we should have and what resources go where. A specific namespace for external-secrets feels a bit too granular to me, but I also wouldn't want it to go in `default` unless required (which I don't think it is). I'd suggest having a `cluster-services` namespace for all supporting services like secrets, DNS, etc, with the possible exception of monitoring and logging, as there will be a lot of resources there so we might want those in a `monitoring` namespace just to keep things clean.

This is up for grabs though, so comments gratefully received (@sengi ?)

terraform/deployments/cluster-infrastructure/external_secrets_iam.tf
```diff
@@ -0,0 +1,50 @@
...
+    effect = "Allow"
+    actions = [
+      "secretsmanager:GetResourcePolicy",
+      "secretsmanager:GetSecretValue",
+      "secretsmanager:DescribeSecret",
+      "secretsmanager:ListSecretVersionIds"
+    ]
+
+    resources = [
+      "arn:aws:secretsmanager:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:secret:*",
```
kerin, line 47: Should we use a prefix for our secrets (`govuk/` or such), just to make sure they're separate from anything else which may end up in the account's Secrets Manager?

kerin (commented):
A couple of queries on namespace and AWS secrets structure, otherwise LGTM

docs/kubernetes-external-secrets.md
```diff
@@ -0,0 +1,39 @@
+# Kubernetes external secrets
+
+This project implements [external-secrets][] as described in [ADR 8][].
```
sengi, line 3: nit: markdown syntax
```suggestion
This project implements [external-secrets] as described in [ADR 8].
```

docs/kubernetes-external-secrets.md
```diff
@@ -0,0 +1,39 @@
...
+
+Not discussed here is how secrets are added to SecretsManager. It would be nice
+if we could avoid having Terraform config for every SecretsManager secret, and
+have this config live entirely in Kubernetes. But the lifecycle of secrets in
+SecretsManager is not the responsibility of ESO, so we need to figure out how
+to manage SecretsManager secrets. Also not discussed is how we’ll rotate secrets,
+this is also out of scope of the ESO.
+
+[external-secrets]: https://external-secrets.io/
+[ADR 8]: ./architecture/decisions/0008-use-external-secrets-for-secrets-storage.md
```
sengi, line 39:
```suggestion
[ADR 8]: architecture/decisions/0008-use-external-secrets-for-secrets-storage.md
```

sengi (approved):
Cool!

terraform/deployments/cluster-infrastructure/external_secrets_iam.tf
```diff
@@ -0,0 +1,50 @@
...
+
+locals {
+  external_secrets_service_account_namespace = "secrets"
+  external_secrets_service_account_name      = "external-secrets"
+}
+
+# The module block external_secrets_iam_role is taken from
+# https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/examples/irsa/irsa.tf,
+# which is Apache-2.0 licenced:
+# https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/LICENSE
```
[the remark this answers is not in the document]
bilbof: I agree, thanks, bit of a Ship of Theseus question :)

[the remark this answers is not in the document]
sengi: Ha, yes exactly 😁

terraform/deployments/cluster-infrastructure/external_secrets_iam.tf
sengi: I'm easy on this. `cluster-services` and maybe `monitoring` SGTM. I'd also be ok with separate ones (though if we were to go for that I'd mildly prefer `external-secrets` as the name).

terraform/deployments/cluster-infrastructure/external_secrets_iam.tf
sengi: Prefixing with the `${cluster_name}/` sounds sensible to me. Any downsides?

kerin: Only that it explicitly binds a given set of secrets to a specific cluster, but that's likely no bad thing.

kerin: ...for now anyway. _Maybe_ at some stage we'll have multiple application environments in a single (non-prod) cluster, but as always we can burn this bridge when we come to it.

docs/kubernetes-external-secrets.md
```diff
@@ -0,0 +1,39 @@
...
+to manage a secret in SecretsManager. It holds references to secrets which hold
+credentials to access the SecretsManager API. The ClusterSecretStore
+'aws-secretsmanager' uses a short-lived Service Account token to authenticate
+with AWS.
+
+The ESO reconciles each ExternalSecret with secrets in AWS SecretsManager (using
+creds provided by ClusterSecretStore), and creates a `Secret` (and keeps the
+Kubernetes Secret in sync with the SecretsManager Secret).
+
+![external-secrets syncing AWS SM secrets to Kubernetes secrets](./img/k8s-external-secrets.png)
```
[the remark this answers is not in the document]
bilbof: It's arguable that this doc isn't needed, given it duplicates information in various parts of the external-secrets docs. Perhaps we should delete it?

I've updated the `ClusterSecretStore` - you're right, it's intended to represent the aws-secretsmanager object.

The lines into and out of the ClusterSecretStore is an attempt to summarise the authn/authz that this object provides to the ESO, interpreted from the docs: https://external-secrets.io/provider-aws-secrets-manager/#eks-service-account-credentials). The docs seem to suggest the store provides credentials, but the diagram suggests it acts as a client for the ESO 🤔 

The numbers intended to show the [rough steps](https://external-secrets.io/api-overview/#behavior) the ESO takes to reconcile an ExternalSecret with an AWS SM Secret to create/update a Kubernetes Secret. I removed the numbers as they're a bit confusing.

terraform/deployments/cluster-infrastructure/external_secrets_iam.tf
bilbof: Thanks, I've changed this to `cluster-services`.

bilbof:
Thanks both, I've made the recommended changes.

docs/kubernetes-external-secrets.md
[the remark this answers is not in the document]
sengi: I'd support removing the doc if it's mostly duplicated and you're happy to. Duplicated docs = more stuff to maintain for little value. If it were me I'd probably just summarise anything that's specific to our usage of it in comments in the TF file and maybe link to their docs.

Thanks for the extra info though. Sorry - I should probably have just read the external-secrets docs a bit more 😅

PR #417, opened 2021-09-02
Specify the subnet CIDR blocks for the integration account.

sengi:
Also improve the formatting of the equivalent config for the test account.

No diff. (`tf plan -var-file ../variables/test/common.tfvars -lock=false`)


kerin (approved):
LGTM

PR #418, opened 2021-09-02
Delete old code related to the ECS project.

sengi:
Remove old code related to the ECS project. The latest version of all the ECS work is on the `ecs` branch and the Concourse pipelines (which still exist for now) are now pointing at that branch.

Also remove ECS-specific stuff from docs and update the setting-up-a-new-environment doc to reflect the switch to EKS (d747f17).

The rspec test for Terraform version consistency between terraform/.terraform-version and the Concourse pipeline has to go away because the Terraform version is no longer specified in the Concourse pipeline config.

Probably easiest to review commit-by-commit. Pretty much just straight deletions apart from d747f17 (see [rendered version](https://github.com/alphagov/govuk-infrastructure/blob/df3035b98889e95bb0955065df8ce58ed1a614c1/docs/create-a-new-environment.md)).

[Trello card](https://trello.com/c/b3GJgOAW/631)

PR #419, opened 2021-09-02
Enable tflint and fix errors in cluster-services module.

sengi:
Follows from #418, where it transpired that we're only linting Terraform modules which contain a `.tflint.hcl`.

Enable tflint for the `cluster-services` module (with same config as the other modules) and fix lint errors:

* Unused variable (`fluentbit_service_account_name`)
* Missing provider version constraints.
* Missing `outputs.tf`. Considered disabling this one, but it's not possible to disable just this specific rule; unfortunately it's conflated with other rules about enforcing the standard module structure which we do kinda want - so let's just put up with it insisting on an empty file for now.

It'd be good to sort out the lint job so that we can have one base config file (ideally still with the ability to override per module) instead of having to remember to put this hidden `.tflint.hcl` file in every root module, but I'll leave that to figure out later.

[Trello card](https://trello.com/c/b3GJgOAW/631)

Tested: `tf init -backend-config test.backend -upgrade` and `tf apply -var-file=../variables/test/common.tfvars` succeeded. The `-upgrade` is in theory only necessary because I'm running with existing state in `.terraform`; CD should continue to work without it. (Famous last words.)

### DIFF
--- a/terraform/deployments/cluster-services/.tflint.hcl
+++ b/terraform/deployments/cluster-services/.tflint.hcl
@@ -1,0 +1,18 @@
+plugin "aws" {
+  enabled = true
+}
+
+config {
+  varfile = [
+    "../variables/common.tfvars",
+    "../variables/test/common.tfvars",
+  ]
+}
+
+rule "terraform_comment_syntax" { enabled = true }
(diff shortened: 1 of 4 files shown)


kerin (approved):
👍

PR #420, opened 2021-09-03
Update Terraform and terraform-aws-eks module versions.

sengi:
[Terraform changelog](https://github.com/hashicorp/terraform/blob/v1.0.5/CHANGELOG.md)

[terraform-aws-eks changelog](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/CHANGELOG.md#v1780---2021-09-03)

There is one notable change here, which is that `ebs_optimized` should now be enabled by default (for instance types which support it) in the managed node group launch template. See PR 1459 in https://github.com/terraform-aws-modules/terraform-aws-eks/ (not linking directly because I want to be a good neighbour and not have GH auto-spam the PR with references)

Testing: mostly need to just merge this and check that it works, but I did check that `cluster-infrastructure` still plans. (`tf init -backend-config test.backend && tf plan -var-file=../variables/test/common.tfvars`) No diff, interestingly: it seems EBS optimisation is already enabled on our nodes 🤷 

[Trello card](https://trello.com/c/My3xZicq/632)


kerin (approved):
LGTM

PR #421, opened 2021-09-03
Adding Nginx sidecar to Rails app Helm charts.

nsabri1:
Adding Nginx reverse proxy to Rails apps as sidecar.

We're not 100% sure what behaviour these will need in the future. Possibly they'll be involved in serving static assets  such ad CSS, JS etc.

It will certainly be needed in some applications such asset manager, which uses ‘x-accel-redirect’.

We are not mandating the use of Nginx sidecar for every Rails app, though we believe it will be the case.

If it turns out that the Nginx sidecar have no behaviour, we may consider removing them at a later date.

Some of the reasons to why we are adding Nginx sidecar:
 
We want apps to be able to work as independently as possible.

Helm chart should be a self-contained encapsulated artefact.

Nginx sidecar will just be doing reverse proxy, static serving and possibly some x-accel-redirect stuff to s3 for the few apps that need this.

Security benefit of having Nginx sit between inbound http traffic and Unicorn.

Unicorn can't deal with slow clients efficiently.

Nginx sidecar Config:

The Nginx config itself will need to be updated to support functionality we require in the future, 

Some reference for NGINX requirements can be found here:
https://docs.google.com/document/d/1hOgkeimIn20O5zrD0wTflap1_DTVatb7NOnb223NhSE/
https://trello.com/c/Lkb8N313/381-epic-nginx-and-varnish-changes-needed

Values naming changes:

Rails application image to: appImage and 
Application container port to: appPort

Nginx sidecar image to - nginxImage
Nginx container portto: nginxPort

### DIFF
--- a/terraform/deployments/cluster-services/.tflint.hcl
+++ b/terraform/deployments/cluster-services/.tflint.hcl
@@ -1,0 +1,18 @@
+plugin "aws" {
+  enabled = true
+}
+
+config {
+  varfile = [
+    "../variables/common.tfvars",
+    "../variables/test/common.tfvars",
+  ]
+}
+
+rule "terraform_comment_syntax" { enabled = true }
+rule "terraform_deprecated_index" { enabled = true }
+rule "terraform_required_providers" { enabled = true }
+rule "terraform_standard_module_structure" { enabled = true }
+rule "terraform_typed_variables" { enabled = true }
+rule "terraform_unused_declarations" { enabled = true }
+rule "terraform_unused_required_providers" { enabled = true }
--- a/terraform/deployments/cluster-services/aws_auth_configmap.tf
+++ b/terraform/deployments/cluster-services/aws_auth_configmap.tf
@@ -111,7 +111,7 @@ resource "kubernetes_cluster_role_binding" "read_crs_and_crbs" {
   role_ref {
     api_group = "rbac.authorization.k8s.io"
     kind      = "ClusterRole"
-    name      = kubernetes_cluster_role.read_crs_and_crbs.metadata.0.name
+    name      = kubernetes_cluster_role.read_crs_and_crbs.metadata[0].name
   }
   subject {
     kind      = "Group"
--- a/terraform/deployments/cluster-services/logging.tf
+++ b/terraform/deployments/cluster-services/logging.tf
@@ -2,8 +2,7 @@
 # NOTE: Kibana and Elasticsearch will in future be replaced by Logit.
 
 locals {
-  fluentbit_service_account_name = "fluentbit"
-  logging_namespace              = "logging"
+  logging_namespace = "logging"
 
   fluentbit_output = <<-OUTPUT
   [OUTPUT]
--- a/terraform/deployments/cluster-services/main.tf
+++ b/terraform/deployments/cluster-services/main.tf
@@ -8,6 +8,22 @@
 
 terraform {
   backend "s3" {}
+  required_providers {
+    kubernetes = {
+      source  = "hashicorp/kubernetes"
+      version = "~> 2.4"
+    }
+    helm = {
+      source  = "hashicorp/helm"
+      version = "~> 2.3"
+    }
+    # The AWS provider is only used here for remote state in remote.tf. Please
+    # do not add AWS resources to this module.
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 3.57"
+    }
+  }
 }
 
 provider "kubernetes" {


sengi:
Having a proper look now but in the meantime please fix the lint errors ;)

helm/content-store/templates/configmap.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-nginx-conf
+data:
+  nginx.conf: |-
+    user  nginx;
+    worker_processes  1;
```
sengi, line 8: This is already the default and we're not likely to need to change it, so I'd probably omit it.
```suggestion
```

helm/content-store/values.yaml
```diff
@@ -10,17 +10,21 @@ service:
...
+appImage:
   repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/content-store
   pullPolicy: IfNotPresent
   tag: latest
 
-containerPort: 80
+nginxImage:
+  repository: nginx
+  pullPolicy: IfNotPresent
+  tag: latest
```
sengi, line 24: I'd recommend using stable as a starting point. Latest can be the development branch with nginx: see their [supported tags](https://hub.docker.com/_/nginx). (Ultimately we might want to pin the versions and have something like Dependabot update them in a managed way, but that's way in the future.)
```suggestion
  tag: stable
```

helm/content-store/values.yaml
```diff
@@ -10,17 +10,21 @@ service:
...
-image:
+appImage:
   repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/content-store
   pullPolicy: IfNotPresent
   tag: latest
 
-containerPort: 80
+nginxImage:
+  repository: nginx
+  pullPolicy: IfNotPresent
```
kerin, line 23: Something to be aware of here - if `imageTag` is `latest` and `pullPolicy` is `IfNotPresent` then Kubelet on the node will pull the image the first time then never pull it again, unless either the tag or pull policy is changed at a later date. This will also apply to other unrelated Pods or Deployments that land on the same worker that also use the same `image:tag` and `pullPolicy: IfNotPresent` combo, as from Kubernetes' point an image tagged with `latest` is indeed already `Present`.

This is one reason why it's recommended to [not use `latest`](https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy), because it's non-deterministic; or, to specify `imagePullPolicy: Always`, so that the latest image is always pulled on deploy. There are pros and cons to both pinning a specific tag and always pulling `latest`, but generally I'd lean towards being specific, so that we don't get an unexpected breaking change from a new major version one day.

tl;dr - I'd change `tag` to `1.21.1`

sengi, line 23: If you switch from `latest` to `stable` (or anything else for that matter) then you'll no longer need to specify this (i.e. it'll default to `IfNotPresent`).
```suggestion
```

kerin (changes requested):
There's one sneaky issue lurking here ready to surprise us in the future, otherwise LGTM 👍

helm/content-store/templates/configmap.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-nginx-conf
+data:
+  nginx.conf: |-
+    user  nginx;
```
sengi, line 7: Formatting nits (sorry 🙈) - the indentation of the values looks a bit funky here (through to the rest of the file). It looks like they were once lined up with spaces but have since gone wonky 😁

I'd probably just go with one space between the key and the value throughout and save the faff of trying to line things up (I don't think making all the values line up helps readability and it leads to unnecessary diffs and merge conflicts if it's enforced).

helm/content-store/values.yaml
sengi: Snap :) If you go with a specific version then we'll need an update mechanism though (which we'll want eventually but it'll probably be a while). We could go with `stable` in the interim so at least we'll get security updates.

sengi: Sorry I'm being an idiot - that'll have the same issue. The interim thing until we have an update mechanism would be to keep pullPolicy: Always in conjunction with using `stable`.

sengi: Sorry didn't think this through properly - if we're using any label that can move around we don't really want IfNotPresent because there's no way for it to know to re-pull when the label moves.

kerin: Yeah, either approach is fine in this case I think, particularly as a specific tag can always be provided at `helm install` time - a default of `stable` with `Always` seems reasonable 👍

sengi: So it's either `pullPolicy: Always` and `stable` or `1.20.1` and no/default PullPolicy and a big scary TODO to figure out a Dependabot-like thing to update the pinned versions.

kerin: This is very much something for another day, but as any Docker tag can arbitrarily move around the only 100% safe usage of `IfNotPresent` is when using a container's image digest (`image@sha256:`) rather than a tag - definitely not something we need to get into now though.

sengi: > particularly as a specific tag can always be provided at helm install time
Ah yeah good point

kerin: Yeah go with `stable / Always` - it's a reasonable default when nothing specific has been requested at `helm install` time.

PR #423, opened 2021-09-03
Remove ECS-specific Terraform vars.

sengi:
Leaving `govuk_environment` for now even though it's a bit of an abstraction leak, as it's still used in the concourse-iam and terraform-lock root modules.

[Trello card](https://trello.com/c/b3GJgOAW/631)

Tested: all five root modules still plan and apply without errors. No diffs. (`tf init -backend-config test.backend -reconfigure && tf apply -var-file=../variables/test/common.tfvars -var-file=../variables/common.tfvars`)


smford (approved):
LGTM

PR #424, opened 2021-09-06
Fix External Secrets pod Crashing

fredericfran-gds:
[External Secrets](https://github.com/external-secrets/external-secrets) was installed via Helm in #415. Unfortunately, its pod kept crashing due to some auth issue.

Further investigation shows that the service account information pass to the helm chart misses the namespace of the service account, this can be seen in the tests run in [here](https://github.com/external-secrets/external-secrets/blob/8827e3ab92cdd87800f44d8f8bd995f0ea749476/pkg/provider/aws/auth/auth_test.go#L342)


fredericfran-gds:
superseded by https://github.com/alphagov/govuk-infrastructure/pull/428

PR #427, opened 2021-09-06
Removing the maintainer validation helm lint check

nsabri1:
Adding validate-maintainers: false to the ct.yml 

This change will stop Helm Lint from failing if `maintainer:` has not been added to Chart.yaml

### DIFF
--- a/ct.yml
+++ b/ct.yml
@@ -1,4 +1,5 @@
 remote: origin
 target-branch: main
+validate-maintainers: false
 chart-dirs:
   - helm


kerin (approved):
👍

smford:
I think this is the right decision disabling the check, we don't want to use a service account on github just to make this check happy given that it is an optional thing in a helm chart; nor do we want to maintain a list of maintainers.

The check can be made to pass if we put any old random github account in there too - which would be bad.

The reason for this check is documented [here](https://github.com/helm/chart-testing/issues/192#issuecomment-661901643) and is basically around public helm charts on the helm official repo needing a maintainer to defined.

PR #428, opened 2021-09-06
Fix external-secrets controller

kerin:
The external-secrets controller service account's namespace must be explicitly specifed when using ClusterSecretStore. This does not appear to be documented anywhere, but the assumption that the namespace is specified can be seen in the source: https://github.com/external-secrets/external-secrets/blob/867006996753dbd4b96d2e8bf3e22c3d583fbcbc/pkg/provider/aws/auth/auth.go#L163


fredericfran-gds (approved):
looks good, thanks for looking this up

PR #429, opened 2021-09-07
Removal of two steps from our Github actions workflow:

nsabri1:
Create EKS cluster and run Helm chart testing have been removed.

Currently we do not need the above steps.

There is a need to come up with a testing solution which will be added to Trello as a new card and linked to the below Trello card.

Trello card: https://trello.com/c/RL7522ud/638-removal-of-create-eks-cluster-and-run-helm-chart-testing-steps-from-github-actions


fredericfran-gds (approved):
looks good

PR #430, opened 2021-09-07
Adding Nginx reverse proxy to Rails apps as sidecar.

nsabri1:
We're not 100% sure what behaviour these will need in the future. Possibly they'll be involved in serving static assets  such ad CSS, JS etc.

It will certainly be needed in some applications such asset manager, which uses ‘x-accel-redirect’.

We are not mandating the use of Nginx sidecar for every Rails app, though we believe it will be the case.

If it turns out that the Nginx sidecar have no behaviour, we may consider removing them at a later date.

**Some of the reasons to why we are adding Nginx sidecar:**

We want apps to be able to work as independently as possible.

  Helm chart should be a self-contained encapsulated artefact.
  
Nginx sidecar will just be doing reverse proxy, static serving and possibly some x-accel-redirect stuff to s3 for the few apps that need this.
  
Unicorn can't deal with slow clients efficiently.

**Nginx sidecar Config:**

The Nginx config itself will need to be updated to support functionality we require in the future,

Some reference for NGINX requirements can be found here

https://docs.google.com/document/d/1hOgkeimIn20O5zrD0wTflap1_DTVatb7NOnb223NhSE/
https://trello.com/c/Lkb8N313/381-epic-nginx-and-varnish-changes-needed

**Values naming changes:**

  Rails application image to: appImage
  Rails application container port to: appPort

  Nginx sidecar image to - nginxImage
  Nginx sidecar container port to: nginxPort

### DIFF
--- a/helm/content-store/templates/deployment.yaml
+++ b/helm/content-store/templates/deployment.yaml
@@ -16,9 +16,9 @@ spec:
     spec:
       containers:
         - name: {{ .Release.Name }}
-          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          image: "{{ .Values.appImage.repository }}:{{ .Values.appImage.tag }}"
           ports:
-            - containerPort: {{ .Values.containerPort }}
+            - containerPort: {{ .Values.appPort }}
           env:
             - name: DEFAULT_TTL
               value: "{{ .Values.appDefaultTtl }}"
@@ -45,7 +45,7 @@ spec:
             - name: PLEK_SERVICE_SIGNON_URI
               value: "{{ .Values.appPlekServiceSignonUri | default (printf "http://signon.%s.%s" .Release.Namespace .Values.clusterDomain) }}"
             - name: PORT
-              value: "{{ .Values.containerPort }}"
+              value: "{{ .Values.appPort }}"
             - name: RAILS_ENV
               value: "{{ .Values.appRailsEnv }}"
             - name: SECRET_KEY_BASE
@@ -58,12 +58,24 @@ spec:
           livenessProbe:
             httpGet:
               path: /healthcheck/live
-              port: {{ .Values.containerPort }}
+              port: {{ .Values.appPort }}
             initialDelaySeconds: 30
             periodSeconds: 5
           readinessProbe:
             httpGet:
               path: /healthcheck/ready
-              port: {{ .Values.containerPort }}
+              port: {{ .Values.appPort }}
             initialDelaySeconds: 30
             periodSeconds: 5
+        - name: {{ .Release.Name }}-nginx
+          image: "{{ .Values.nginxImage.repository }}:{{ .Values.nginxImage.tag }}"
+          ports:
+            - containerPort: {{ .Values.nginxPort }}
+          volumeMounts:
+          - name: {{ .Release.Name }}-nginx-conf
+            mountPath: /etc/nginx/nginx.conf
+            subPath: nginx.conf
+      volumes:
+      - name: {{ .Release.Name }}-nginx-conf
+        configMap:
+          name: {{ .Release.Name }}-nginx-conf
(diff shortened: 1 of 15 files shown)


helm/content-store/templates/deployment.yaml
```diff
@@ -58,12 +58,24 @@ spec:
...
             initialDelaySeconds: 30
             periodSeconds: 5
           readinessProbe:
             httpGet:
               path: /healthcheck/ready
-              port: {{ .Values.containerPort }}
+              port: {{ .Values.appPort }}
             initialDelaySeconds: 30
             periodSeconds: 5
+        - name: {{ .Release.Name }}-nginx
```
kerin, line 70: We should include basic liveness and readiness probes for the Nginx container - I'd go with a TCP check rather than HTTP, as per [the Nginx helm chart](https://github.com/bitnami/charts/blob/f88d9fae4eaeba918df2aec7a76efb268e3d0318/bitnami/nginx/templates/deployment.yaml#L165), as any HTTP response means that Nginx is alive and ready.

helm/content-store/templates/deployment.yaml
```diff
@@ -16,9 +16,9 @@ spec:
     spec:
       containers:
         - name: {{ .Release.Name }}
-          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          image: "{{ .Values.appImage.repository }}:{{ .Values.appImage.tag }}"
           ports:
-            - containerPort: {{ .Values.containerPort }}
+            - containerPort: {{ .Values.appPort }}
```
kerin, line 21: I'd recommend always specifying a port name ([example](https://github.com/bitnami/charts/blob/f88d9fae4eaeba918df2aec7a76efb268e3d0318/bitnami/nginx/templates/deployment.yaml#L157)), so that health check, Services etc. that reference the port don't need to be updated if the containerPort is changed.

helm/content-store/values.yaml
```diff
@@ -10,39 +10,48 @@ service:
...
   port: 80
-  targetPort: 80
-
 
 replicaCount: 1
 
-image:
+appImage:
   repository: 172025368201.dkr.ecr.eu-west-1.amazonaws.com/content-store
   pullPolicy: IfNotPresent
```
kerin, line 18: This should be `Always` if the default tag is `latest`.

helm/content-store/values.yaml
```diff
@@ -10,39 +10,48 @@ service:
...
-  # memory: 1024M
-  requests:
-  # cpu:
-  # memory: 1024M
-
-
+nginxImage:
+  repository: nginx
+  pullPolicy: IfNotPresent
+  tag: stable
```
kerin, line 24: `stable` is a moving tag, much like `latest`, so pullPolicy should be `Always`.

helm/static/templates/service.yaml
```diff
@@ -14,7 +14,7 @@ spec:
   type: {{ .Values.service.type }}
   ports:
     - port: {{ .Values.service.port }}
-      targetPort: {{ .Values.containerPort}}
+      targetPort: {{ .Values.nginxPort}}
       protocol: TCP
       name: https
```
kerin, line 19: This should be `http`, as we're terminating TLS at the load balancer.

kerin (changes requested):
A couple of minor things, LGTM otherwise

helm/content-store/templates/deployment.yaml
nsabri1: Whats your thoughts on the following name - this port will be local port for the Rails app conatiner ( - name: {{ .Release.Name }}-port )

helm/content-store/templates/deployment.yaml
nsabri1: i will look into this and then run some test after implementation.

helm/content-store/templates/deployment.yaml
nsabri1: in a way it is a http port - on the rails side so could also be ( -name : http ) baring it uses port 3000 but i could change the port to 8080

kerin: It should be `http` as that's the protocol the port is exposing.

kerin (approved):
LGTM 👍

PR #431, opened 2021-09-07
Ship logs to Logit

bilbof:
This sends logs to Logit from Filebeat.

Filebeat is the recommended log shipper to use with Kubernetes and Logit: https://logit.io/sources/configure/kubernetes.

I replaced Fluent-Bit as we would need a corresponding Logstash filter in each stack:

```
filter {
  if [message] =~ /^{.*}/ or [type] == "json" {

    mutate { gsub => ["message", "app.kubernetes.io", "app_kubernetes_io"] }
    json { source => "message" }
 }
}
```

I manually added the host credential to SecretsManager for the Logit host, under `govuk/logit-host`. This is stored in the form `{"host":"foo"}`.

https://trello.com/c/zjJxCwDk/627-ship-logs-to-logit

### DIFF
--- a/terraform/deployments/cluster-services/external_secrets.tf
+++ b/terraform/deployments/cluster-services/external_secrets.tf
@@ -35,3 +35,13 @@ resource "helm_release" "cluster_secret_store" {
     serviceAccountName = data.terraform_remote_state.cluster_infrastructure.outputs.external_secrets_service_account_name
   })]
 }
+
+resource "helm_release" "cluster_secrets" {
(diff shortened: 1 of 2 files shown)


sengi (commented):
Cool. Do you have a link to a working stack so we can see how the fields are being mapped and so on?

sengi (approved):
Fields look sensible from what I could see on the test rig. I didn't find any JSON logs to see how those come out, but we can easily add to the config later if need be.

PR #432, opened 2021-09-08
Use govuk ruby for apps

karlbaker02:
This PR updates the `build-images` pipeline to specify the `govuk-ruby-2.7.X` image as the base image for building app images.

There are a [number of different ways to do this](concourse/oci-build-task#14) but this PR uses makes use of the [`IMAGE_ARG_*` param as part of the `vito/oci-build-task` resource](concourse/oci-build-task#52) which allows us to use image resources already defined in the pipeline within our Dockerfiles, where the base image is served from a registry local to the pipeline. It does this by bringing in these images as resources, namespaced under separate ECR repositories, then passing in the appropriate image to the build process via the `IMAGE_ARG_base_image` param (reflected in app `Dockerfile`s); the `FROM` command in said `Dockerfile`s then point to the correct ECR repository for the `govuk-ruby-2.7.X` base image (rather than to Docker Hub).

Trello: https://trello.com/c/dZQwFSKK

### DIFF
--- a/concourse/pipelines/build-images.yml
+++ b/concourse/pipelines/build-images.yml
@@ -127,6 +127,20 @@ resources:
       tag: latest
 
   - <<: *docker-image
+    name: govuk-ruby-2.7.2
+    source:
+      <<: *docker-image-source
+      repository: govuk-ruby-2.7.2
+      tag: latest
+
+  - <<: *docker-image
+    name: govuk-ruby-2.7.3
+    source:
+      <<: *docker-image-source
+      repository: govuk-ruby-2.7.3
+      tag: latest
+
+  - <<: *docker-image
     name: infra-concourse-task-image
     source:
       <<: *docker-image-source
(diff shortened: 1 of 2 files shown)


concourse/parameters/ci/ci.yml
```diff
@@ -2,4 +2,4 @@ aws_region: eu-west-1
 ecr_registry_id: "172025368201"
 concourse_ci_role_arn: arn:aws:iam::430354129336:role/govuk-ci-concourse
 concourse_ecr_role_arn: arn:aws:iam::172025368201:role/pull_images_from_ecr_role
-govuk_infrastructure_branch: main
+govuk_infrastructure_branch: use-govuk-ruby-base-image
```
rtrinque, line 5: Why do we need to use a specific branch? Isn't the goal to use main once everything is merged?

rtrinque (approved):
Looks good

concourse/pipelines/build-images.yml
```diff
@@ -126,6 +132,20 @@ resources:
       repository: frontend
       tag: latest
 
+  - <<: *docker-image
+    name: govuk-ruby-2.7.2
+    source:
+      <<: *docker-image-source
+      repository: govuk-ruby
```
karlbaker02, line 139: Change this to `govuk-ruby-2.7.2`

PR #435, opened 2021-09-09
Ensure unique names for IRSA IAM roles within the account.

sengi:
- Ensure that the IAM role for external-secrets contains the cluster name, so that we don't break the ability to turn up extra clusters in the test account.
- Construct the role names consistently for all of the cluster addons. I've chosen to include the cluster name in the IAM role names to satisfy the uniqueness constraint but omit it from the k8s ClusterRole names so that the ClusterRole names stay the same (and so that they don't contain a spurious "govuk-" prefix).
- Add descriptions to the IAM roles and include the name of the k8s ClusterRole.

[Trello card](https://trello.com/c/hqCr1ozs/625)


kerin (approved):
LGTM

PR #436, opened 2021-09-09
Add the external-dns cluster addon.

sengi:
- Install the addon using Bitnami's Helm chart.
- Create a Route53 hosted zone to be managed by the addon, taking the name of the parent domain from the canonical source in alphagov/govuk-aws.
- Create an IAM role and service account to allow it to manage the zone.
- Add a var for passing force_destroy to resource in non-prod, so that we can still run `terraform destroy`
in the test account without manual faff.

Records will not be automatically deleted when Ingresses/Services are deleted. It's possible to change this by setting `policy=sync` instead of the default `upsert-only`, but there's a good reason for the default behaviour: e.g. if we were to set `policy=sync` and someone were to remove and then immediately reinstall a Helm chart, it could cause significant downtime beyond the actual window where the record didn't exist (because of negative caching).

TTL defaults to 5 minutes, which seems fine for now so I've left it alone.

Considered using https://registry.terraform.io/modules/lablabs/eks-external-dns/aws/ but that would only fit if we decided to merge the cluster-infrastructure and cluster-services modules.

[Trello card](https://trello.com/c/hqCr1ozs/625)

Tested:
- Tried the [Service example from the docs](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/aws.md#verify-externaldns-works-service-example) with `nginx-test.eks.test.govuk.digital`. Works.
- Checked for any significant warnings on [Zonemaster](https://zonemaster.net/result/73385cfed48e262f) - nothing major.

### DIFF
--- a/terraform/deployments/cluster-infrastructure/external_dns.tf
+++ b/terraform/deployments/cluster-infrastructure/external_dns.tf
@@ -1,0 +1,56 @@
+# external_dns.tf defines a Route53 zone and an IAM role and policy to allow
+# the k8s external-dns addon to manage the zone.
+#
+# The k8s side of the external-dns addon is in
+# ../cluster-services/external_dns.tf.
+
+locals {
+  external_dns_service_account_name = "external-dns"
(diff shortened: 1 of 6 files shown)


kerin (approved):
LGTM 👍

PR #438, opened 2021-09-09
Set default IngressClass and add DNS for Prometheus stack

kerin:
- Adds an `ingress-class` Helm chart to create a default `IngressClass` resource for AWS LB Controller
- Removes ingress class annotation on Prometheus ingress resources
- Adds Ingress pathType to Grafana ingress for consistency
- Specifies hostnames to Ingress resources for external-dns


nsabri1 (approved):
LGTM

sengi:
Thanks for sorting out the default ingress class, been meaning to get around to that 👍

PR #442, opened 2021-09-10
Add Publisher Helm Chart

fredericfran-gds:
This PR adds the Publisher Helm Chart.

Compared to the previous apps, publisher works in 2 modes:
web and worker. This is accounted for in the chart.

We also include a Kubernetes job that does the db migration if enabled
in the values.yaml file on each loading of the chart in Kubernetes.

Ref:
1. [trello card](https://trello.com/c/bnJygqQl/640-add-database-migration-job-to-helm-charts)

### DIFF
--- a/helm/publisher/templates/common/configmap.yaml
+++ b/helm/publisher/templates/common/configmap.yaml
@@ -1,0 +1,30 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-env
+data:
+  ASSETS_PREFIX: "{{ .Values.appAssetsPrefix }}"
+  BASIC_AUTH_USERNAME: "{{ .Values.appBasicAuthUsername }}"
+  DEFAULT_TTL: "{{ .Values.appDefaultTtl }}"
+  EMAIL_GROUP_BUSINESS: "{{ .Values.appEmailGroupBusiness }}"
+  EMAIL_GROUP_CITIZEN: "{{ .Values.appEmailGroupCitizen }}"
+  EMAIL_GROUP_FORCE_PUBLISH_ALERTS: "{{ .Values.appEmailGroupForcePublishAlerts }}"
+  FACT_CHECK_SUBJECT_PREFIX: "{{ .Values.appFactCheckSubjectPrefix }}"
+  FACT_CHECK_USERNAME: "{{ .Values.appFactCheckUsername }}"
+  GOVUK_APP_DOMAIN: "{{ .Values.appGovukAppDomain | default (printf "%s.eks.%s.%s" .Release.Namespace .Values.govukEnvironment .Values.clusterDomain) }}"
+  GOVUK_APP_DOMAIN_EXTERNAL: "{{ .Values.appGovukDomainExternal | default (printf "%s.eks.%s.%s" .Release.Namespace .Values.govukEnvironment .Values.govukDomainExternal) }}"
+  GOVUK_APP_NAME: "{{ .Values.appGovukAppName | default .Release.Name }}"
+  GOVUK_APP_ROOT: "{{ .Values.appGovukAppRoot }}"
+  GOVUK_APP_TYPE: "{{ .Values.appGovukAppType }}"
+  GOVUK_ASSET_ROOT: "{{ .Values.appGovukAssetRoot | default (printf "https://assets.%s.eks.%s.%s" .Release.Namespace .Values.govukEnvironment .Values.govukDomainExternal) }}"
+  ASSET_HOST: "{{ .Values.appAssetHost | default (printf "https://www-origin.%s.eks.%s.%s" .Release.Namespace .Values.govukEnvironment .Values.govukDomainExternal) }}"
+  GOVUK_ENVIRONMENT: "{{ .Values.govukEnvironment }}"
+  GOVUK_WEBSITE_ROOT: "{{ .Values.appGovukWebsiteRoot | default (printf "https://www-origin.%s.eks.%s.%s" .Release.Namespace .Values.govukEnvironment .Values.govukDomainExternal) }}"
+  PLEK_SERVICE_CONTENT_STORE_URI: "{{ .Values.appPlekServiceContentStoreUri | default (printf "http://content-store.%s.%s" .Release.Namespace .Values.clusterDomain) }}"
+  PLEK_SERVICE_DRAFT_ORIGIN_URI: "{{ .Values.appPlekServiceContentStoreUri | default (printf "https://draft-origin.%s.eks.%s.%s" .Release.Namespace .Values.govukEnvironment .Values.govukDomainExternal) }}"
+  PLEK_SERVICE_PUBLISHING_API_URI: "{{ .Values.appPlekServicePublishingApiUri | default (printf "http://publishing-api-web.%s.%s" .Release.Namespace .Values.clusterDomain) }}"
+  PLEK_SERVICE_SIGNON_URI: "{{ .Values.appPlekServiceSignonUri | default (printf "https://signon.%s.eks.%s.%s" .Release.Namespace .Values.govukEnvironment .Values.govukDomainExternal) }}"
+  PLEK_SERVICE_STATIC_URI: "{{ .Values.appPlekServiceStaticUri | default (printf "http://static.%s.%s" .Release.Namespace .Values.clusterDomain) }}"
+  PORT: "{{ .Values.appPort }}"
+  RAILS_ENV: "{{ .Values.appRailsEnv }}"
+  REDIS_URL: "{{ .Values.appRedisUrl }}"    # TODO: decide on Redis for EKS
--- a/helm/publisher/templates/dbmigrationjob.yaml
+++ b/helm/publisher/templates/dbmigrationjob.yaml
@@ -1,0 +1,37 @@
+{{- if .Values.dbMigrationEnabled -}}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ .Release.Name }}-dbmigrate-{{ now | unixEpoch }}
+spec:
+  template:
+    metadata:
+      name: "{{ .Release.Name }}"
+      labels:
+        app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+        app.kubernetes.io/instance: {{ .Release.Name | quote }}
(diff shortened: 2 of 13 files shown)


helm/publisher/templates/web/deployment.yaml
```diff
@@ -0,0 +1,86 @@
...
+      annotations:
+        # TODO: if Helm library is used, `$.Template.BasePath` needs to be replaced 
+        # by chart name, see https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments
+        checksum/configmap: {{ include (print $.Template.BasePath "/common/configmap.yaml") . | sha256sum }}
+        checksum/secrets: {{ include (print $.Template.BasePath "/common/secrets.yaml") . | sha256sum }}
+    spec:
+      containers:
+        - name: app
+          image: "{{ .Values.appImage.repository }}:{{ .Values.appImage.tag }}"
+          command: ["foreman"]
```
kerin, line 25: This is a bit outside of my lane, but I don't think we want to use foreman here, as it's not intended for production use and introduces an extra level of indirection between the container's entrypoint and the actual Rails process. That is primarily an issue for the container/Dockerfile though, which we should likely look at separately from this PR.

helm/publisher/templates/web/ingress.yaml
```diff
@@ -0,0 +1,54 @@
...
+metadata:
+  name: {{ .Values.ingress.name }}
+  labels:
+    {{- include "publisher.labels" . | nindent 4 }}
+  {{- with .Values.ingress.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  {{- if .Values.ingress.tls }}
```
kerin, line 26: Just a note - with AWS Load Balancer's integration with AWS Certificate Manager we should never need to configure TLS specifically as it will discover and attach the certificate based on `Ingress.spec.rules.host`. It's a common convention to have Ingress TLS configurable in a Helm chart though, so it's fine to leave this here even if we're unlikely to use it, in the spirit of not binding our application charts to our specific EKS configuration.

helm/publisher/templates/web/service.yaml
```diff
@@ -0,0 +1,21 @@
...
+    {{- range $key, $value := .Values.service.annotations }}
+    {{ $key }}: {{ $value }}
+    {{- end }}
+{{- end }}
+spec:
+  type: {{ .Values.service.type }}
+  ports:
+    - port: {{ .Values.service.port }}
+      targetPort: {{ .Values.nginxPort }}
+      protocol: TCP
```
kerin, line 18: TCP is the default protocol so this can be safely be omitted.

helm/publisher/templates/web/service.yaml
```diff
@@ -0,0 +1,21 @@
...
+      name: https
```
kerin, line 19: HTTPS will be terminated at the load balancer, with nginx exposing plain HTTP, so I'd change this to `http` to match.

helm/publisher/templates/worker/deployment.yaml
```diff
@@ -0,0 +1,90 @@
...
+        # by chart name, see https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments
+        checksum/configmap: {{ include (print $.Template.BasePath "/common/configmap.yaml") . | sha256sum }}
+        checksum/secrets: {{ include (print $.Template.BasePath "/common/secrets.yaml") . | sha256sum }}
+    spec:
+      containers:
+        - name: app
+          image: "{{ .Values.appImage.repository }}:{{ .Values.appImage.tag }}"
+          command: ["foreman"]
+          args: ["run", "worker"]
+          ports:
```
kerin, line 27: As with the DB migration Job, the worker process presumably doesn't need to expose a port?

helm/publisher/templates/worker/deployment.yaml
```diff
@@ -0,0 +1,90 @@
...
+            periodSeconds: 5
+          readinessProbe:
+            exec:
+              command:
+              - /bin/sh
+              - -c
+              - exit 0
+            initialDelaySeconds: 30
+            periodSeconds: 5
+        - name: nginx
```
kerin, line 62: Presumably a worker process also doesn't need an nginx proxy container?

helm/publisher/values.yaml
```diff
@@ -0,0 +1,118 @@
...
+#     value: world
+
+
+ingress:
+  enabled: true
+  name: publisher
+  host:
+  path: "/"
+  pathType: "Prefix"
+  className: alb
```
kerin, line 94: Ingress class name doesn't need to be specified, as alb ingress is the default in our cluster.

kerin (changes requested):
A couple of minor things to address and some notes FYI - LGTM otherwise though.

helm/publisher/templates/common/configmap.yaml
```diff
@@ -0,0 +1,30 @@
...
+  PLEK_SERVICE_CONTENT_STORE_URI: "{{ .Values.appPlekServiceContentStoreUri | default (printf "http://content-store.%s.%s" .Release.Namespace .Values.clusterDomain) }}"
+  PLEK_SERVICE_DRAFT_ORIGIN_URI: "{{ .Values.appPlekServiceContentStoreUri | default (printf "https://draft-origin.%s.eks.%s.%s" .Release.Namespace .Values.govukEnvironment .Values.govukDomainExternal) }}"
+  PLEK_SERVICE_PUBLISHING_API_URI: "{{ .Values.appPlekServicePublishingApiUri | default (printf "http://publishing-api-web.%s.%s" .Release.Namespace .Values.clusterDomain) }}"
+  PLEK_SERVICE_SIGNON_URI: "{{ .Values.appPlekServiceSignonUri | default (printf "https://signon.%s.eks.%s.%s" .Release.Namespace .Values.govukEnvironment .Values.govukDomainExternal) }}"
+  PLEK_SERVICE_STATIC_URI: "{{ .Values.appPlekServiceStaticUri | default (printf "http://static.%s.%s" .Release.Namespace .Values.clusterDomain) }}"
+  PORT: "{{ .Values.appPort }}"
+  RAILS_ENV: "{{ .Values.appRailsEnv }}"
+  REDIS_URL: "{{ .Values.appRedisUrl | default (printf "redis://redis.ecs.%s.%s:6379" .Values.govukEnvironment .Values.govukDomainInternal) }}" #TODO: use EKS Redis cluster
```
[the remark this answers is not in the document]
fredericfran-gds: we are still using the ECS one, we need to spin up our own one for EKS.

helm/publisher/templates/dbmigrationjob.yaml
```diff
@@ -0,0 +1,40 @@
...
+        app.kubernetes.io/instance: {{ .Release.Name | quote }}
+        helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: publisher-dbmigrate
+        image: "{{ .Values.appImage.repository }}:{{ .Values.appImage.tag }}"
+        command: ["bundle"]
+        args: ["exec", "rails", "db:migrate"]
+        ports:
```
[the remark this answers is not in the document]
fredericfran-gds: Umm, good question, I will try it without. I guess not since it is not receiving requests in?

helm/publisher/templates/web/deployment.yaml
```diff
@@ -0,0 +1,86 @@
...
+      app: {{ .Release.Name }}-web
+  template:
+    metadata:
+      labels:
+        app: {{ .Release.Name }}-web
+      annotations:
+        # TODO: if Helm library is used, `$.Template.BasePath` needs to be replaced 
+        # by chart name, see https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments
+        checksum/configmap: {{ include (print $.Template.BasePath "/common/configmap.yaml") . | sha256sum }}
+        checksum/secrets: {{ include (print $.Template.BasePath "/common/secrets.yaml") . | sha256sum }}
```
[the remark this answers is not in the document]
fredericfran-gds: cool, I did make a note about it in the PR description. I agree that if the contents rather than declaration of the secrets change, then there will be rolling.

helm/publisher/templates/web/deployment.yaml
fredericfran-gds: this is how we ran it in ECS, I'm no ruby expert too: https://github.com/alphagov/govuk-infrastructure/blob/fe361ded9480b71014fe3dc4c6912a04f2a2f370/terraform/deployments/govuk-publishing-platform/app_publisher.tf#L86

helm/publisher/templates/web/ingress.yaml
fredericfran-gds: ok cool

helm/publisher/templates/web/service.yaml
fredericfran-gds: ok, I will do.

helm/publisher/templates/web/service.yaml
fredericfran-gds: ok, I will modify that, I guess it is the same in other services merged.

helm/publisher/values.yaml
fredericfran-gds: ok, I will remove that.

helm/publisher/templates/worker/deployment.yaml
fredericfran-gds: I can remove it, I guess maybe we don't need.

helm/publisher/templates/worker/deployment.yaml
fredericfran-gds: I can remove it, I guess maybe we don't need.

helm/publisher/templates/common/configmap.yaml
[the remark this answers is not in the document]
fredericfran-gds: I will decide what to do in the next Trello card as discussed on slack.

helm/publisher/templates/web/deployment.yaml
kerin: Yeah that will work, but it's not really a production-grade setup. We can dig into that separately from this PR though.

kerin (approved):

PR #443, opened 2021-09-13
TLS and HTTPS redirect for Ingress

sengi:
Create a wildcard TLS cert for the external-dns domain, discoverable by the [Amazon Load Balancer Controller](https://kubernetes-sigs.github.io/aws-load-balancer-controller/).

Also tweak the NCACHE time on the DNS zone where external-dns creates names for Ingresses and Services\[type=LoadBalancer], so that we don't get hosed by negative caching when turning up a new name.

Beware that [some of the aws-lb-controller documentation](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/tasks/ssl_redirect/) is out of date with regard to configuring the HTTPS redirect. The new way as of v2.2 is `alb.ingress.kubernetes.io/ssl-redirect` whereas the old, hacky way is `alb.ingress.kubernetes.io/actions.ssl-redirect`. There are also still a lot of examples floating around using the deprecated v1beta1 Ingress API; those are also best avoided.

#### Testing

Using Amazon's [2048 example](https://github.com/kubernetes-sigs/aws-load-balancer-controller/blob/main/docs/examples/2048/2048_full_latest.yaml), checked that:

* The LB is created and destroyed as expected.
* The expected TLS cert is presented to the client.
* The name of the LB can be set.
* HTTPS redirect works.
* TLS <1.2 can be disabled.

```
---
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: ingress-2048
  annotations:
    alb.ingress.kubernetes.io/scheme: internet-facing
    alb.ingress.kubernetes.io/target-type: ip
    alb.ingress.kubernetes.io/load-balancer-name: 2048game-test
    alb.ingress.kubernetes.io/listen-ports: '[{"HTTP": 80}, {"HTTPS": 443}]'
    alb.ingress.kubernetes.io/ssl-redirect: "443"
    alb.ingress.kubernetes.io/ssl-policy: ELBSecurityPolicy-TLS-1-2-Ext-2018-06  # No TLS 1.0 or 1.1.
spec:
  rules:
  - host: 2048game.eks.test.govuk.digital
    http:
      paths:
      - pathType: Prefix
        path: "/"
        backend:
          service:
            name: service-2048
            port:
              number: 80
```

```
$ curl -sv http://2048game.eks.test.govuk.digital/  >/dev/null
*   Trying 54.194.48.85...
* TCP_NODELAY set
* Connected to 2048game.eks.test.govuk.digital (54.194.48.85) port 80 (#0)
> GET / HTTP/1.1
> Host: 2048game.eks.test.govuk.digital
> User-Agent: curl/7.64.1
> Accept: */*
> 
< HTTP/1.1 301 Moved Permanently
< Server: awselb/2.0
< Date: Tue, 14 Sep 2021 11:30:12 GMT
< Content-Type: text/html
< Content-Length: 134
< Connection: keep-alive
< Location: https://2048game.eks.test.govuk.digital:443/
```

I was a bit wary of the redundant `:443` but it seems at least Firefox and Chrome elide it so it doesn't cause any user-visible ugliness.

#### Follow-up work / not in this PR

I plan to update the existing Helm charts to use these annotations imminently in a separate PR.

I'd like to find a way to set some of these annotations as defaults via IngressClass (especially `target-type: ip`, since we'd never want the default spray-to-nodes sideways-forwarding comedy show, and `ssl-policy`). It doesn't look like that's implemented yet but it seems like it ought to be doable.

The validation DNS record for the `*.(env.)?publishing.service.gov.uk` SAN is not currently handled automatically; added a TODO for this because we can't easily fix it right now.

### DIFF
--- a/terraform/deployments/cluster-infrastructure/external_dns.tf
+++ b/terraform/deployments/cluster-infrastructure/external_dns.tf
@@ -1,12 +1,14 @@
 # external_dns.tf defines a Route53 zone and an IAM role and policy to allow
-# the k8s external-dns addon to manage the zone.
+# the k8s external-dns addon to manage the zone, plus a wildcard TLS cert which
+# covers the external-dns domain and the user-facing
+# *.(env.)?publishing.service.gov.uk domain.
 #
 # The k8s side of the external-dns addon is in
 # ../cluster-services/external_dns.tf.
 
 locals {
   external_dns_service_account_name = "external-dns"
-  external_dns_domain_name          = "${var.external_dns_subdomain}.${data.terraform_remote_state.infra_root_dns_zones.outputs.external_root_domain_name}"
+  external_dns_zone_name            = trimsuffix("${var.external_dns_subdomain}.${data.terraform_remote_state.infra_root_dns_zones.outputs.external_root_domain_name}", ".")
 }
 
 module "external_dns_iam_role" {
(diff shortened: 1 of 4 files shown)


terraform/deployments/cluster-infrastructure/external_dns.tf
```diff
@@ -44,14 +46,63 @@ data "aws_iam_policy_document" "external_dns" {
...
+    # the validation records for *.(env.)?publishing.service.gov.uk can easily
+    # be added here as part of turnup/DR automation. Those records currently
+    # exist for the test/integration/staging/prod environments in a single zone
+    # managed in alphagov/govuk-dns-config.
+    for dvo in aws_acm_certificate.cluster_public.domain_validation_options : dvo.domain_name => {
+      name   = dvo.resource_record_name
+      record = dvo.resource_record_value
+      type   = dvo.resource_record_type
+    }
+    if length(regexall("${local.external_dns_zone_name}\\.$", dvo.resource_record_name)) > 0
```
sengi, line 95: I hate this btw, but I couldn't think of a less horrible way to avoid creating nonsense records like `_abadcafedeadbeef29ece89c0e163b94.test.publishing.service.gov.uk.eks.test.govuk.digital.` without loss of generality. Suggestions welcome.

kerin (approved):
LGTM - the ACM DNS validation bit isn't hugely clear to me at the moment, because I'm not familiar with the wider DNS topography yet, so I've got no suggestions on that front just now.

sengi:
> LGTM - the ACM DNS validation bit isn't hugely clear to me at the moment, because I'm not familiar with the wider DNS topography yet, so I've got no suggestions on that front just now.

No worries - basically it's one big zone not split by environment/account and managed via a bunch of whacky home-grown stuff in https://github.com/alphagov/govuk-dns-config/blob/master/service.gov.uk.yaml (private repo but you should have access). We wanna replace that stuff with plain TF at some point but it's not directly related to Replatforming.

PR #445, opened 2021-09-14
Build image pipeline fixes

karlbaker02:
This PR address two fixes from the recent work to use `govuk-ruby` as the base image for building our apps - setting Frontend to reference the correct Ruby version, and building `infra-concourse-tasks` also using the `govuk-ruby` base image.

Trello: https://trello.com/c/dZQwFSKK


fredericfran-gds (approved):
ok

PR #446, opened 2021-09-14
Add pipeline and registries to build govuk-ruby base images

rtrinque:
We are building most of our govuk apps docker images on the official ruby images, as we are using older versions of ruby however, those images are not updated for security fixes.
The way we solve this problem is by building our own ruby images (govuk-ruby) that are based on the official ruby images but with an upgraded debian distro.
For this purpose we use a pipeline that build each version of govuk-ruby that we need (for now 2.7.2 for most apps, 2.7.3 for frontend and 2.6.6 for whitehall)
base_images_pipeline
This pipeline is triggered daily, after the build the list of packages of the newly built image is compared to the list of the current latest fetched from ECR, the newly built image is pushed to ECR only if the packages list has changed to avoid spamming ECR with identical images every day.
The first job in the pipeline will built the new image and compare it to latest, the job will then fail if packages are identical or proceed to the next job if they are not. The next job then tag and pushes the new image to ECR.

Trello card : https://trello.com/c/gOqiwHcZ/604-build-our-own-base-image
List of conainers per app : https://docs.google.com/spreadsheets/d/1v469JXRtZGCsWqiJdryPzXM-ckrVHw7DuwLb3oTpeHg/edit#gid=0

### Possible improvements and gotchas:
This pipeline is triggered on a time basis which is not optimal as ideally we would like to trigger it only when there are new packages/security fixes, there is no quick and easy way to monitor that however, an improvement would be to trigger only when needed. This would have the added benefit of not requiring any check to make sure the image is worth pushing to ECR
The comparison between the built image and latest is necessary to avoid spamming ECR with identical images, there should be a better way of doing this than checking the list of installed packages, using a digest of the image should be the logical way, it's not easy however as the task we use to build images provides us with a digest that changes even when the content of the image doesn't.

### DIFF
--- a/concourse/pipelines/build-base-images.yml
+++ b/concourse/pipelines/build-base-images.yml
@@ -1,0 +1,443 @@
+---
+definitions:
+
+resources:
+  - name: daily
+    type: time
+    icon: clock-outline
+
+  - &git-repo
+    name: govuk-infrastructure
+    icon: github
+    type: git
+    source: &git-repo-source
+      uri: git@github.com:alphagov/govuk-infrastructure.git
+      branch: base_image
+      private_key: | # pragma: allowlist secret
+        ((govukci_private_key))
+
+  - &docker-image
+    name: govuk-ruby-2.7.2-image
+    type: registry-image
+    icon: docker
+    source: &docker-image-source
+      repository: govuk-ruby-2.7.2
+      tag: latest
+      aws_region: ((concourse-ecr-user_aws-region))
+      aws_access_key_id: ((concourse-ecr-user_aws-access-key))
+      aws_secret_access_key: ((concourse-ecr-user_aws-secret-access-key))
+      aws_role_arn:
+        "arn:aws:iam::((aws_production_account_id)):role\
+        /push_image_to_ecr_role"
+
+  - <<: *docker-image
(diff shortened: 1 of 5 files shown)


smford:
Looking good

The container build process  means there is three stages: build, check-to-upload and upload.  The check-to-upload step is conditional, meaning only upon a condition being met should the upload occur.

Given concourse only handles success or fail, there is no clear way for the conditional to occur, nor for it to report the success or failure correctly within the concourse UI, which inturn flows on to our ability to monitor and alert when problems with the process occurs.  A possible solution would be to write a concourse resource to solve this problem. 

This PR will enable us to build the containers and progress with the important tasks, but is acknowledged that further work needs to be done to flesh out a complete solution as detailed in the PR gotchas.

Use cases:
1. If build job fails, turn red/alert
2. If check-to-upload code has a error, turn red/alert
3. If check-to-upload code run is successful, but decides not to upload, do not turn red/alert (be green) and not attempt an upload
4. If upload fails, turn red/alert

Use case 3 is the problem use case.

smford (approved):
Looks good to me.

Thanks Roch for testing anchors and imports, which interestingly passes yamllint'ing but causes concourse to barf.

PR #447, opened 2021-09-14
Minor updates to helm charts

fredericfran-gds:
Minor updates to the other Helm chart to reflect what was done in #442, i.e:
1. removed default service protocol TCP and name
2. changed the Nginx config of the sidecar to forward IP and host to the app

In addition for Router, we:
1. removed some extra comments in the values.yaml to bring it in line with the other values.yaml files.
2. added the host for the ingress now that the DNS controller is active in the cluster.

Tested end-to-end connectivity: http://www-origin.fred.eks.test.govuk.digital/

### DIFF
--- a/helm/content-store/templates/configmap.yaml
+++ b/helm/content-store/templates/configmap.yaml
@@ -5,7 +5,7 @@ metadata:
 data:
   nginx.conf: |-
     user  nginx;
-    
+
     error_log  /var/log/nginx/error.log warn;
     pid        /var/run/nginx.pid;
 
@@ -30,6 +30,8 @@ data:
         listen {{ .Values.nginxPort }};
 
         location / {
+          proxy_set_header   Host $host;
+          proxy_set_header   X-Real-IP $remote_addr;
           proxy_pass         http://{{ .Release.Name }};
           proxy_redirect     off;
         }
--- a/helm/router/templates/NOTES.txt
+++ b/helm/router/templates/NOTES.txt
@@ -1,6 +1,6 @@
 1. Get the application URL by running these commands:
 {{- if .Values.ingress.enabled }}
-  http{{ if $.Values.ingress.tls }}s{{ end }}://{{ .Values.ingress.host | default (printf "www-origin.%s.%s.%s" .Release.Namespace .Values.govukEnvironment .Values.govukDomainExternal ) }}{{ .path }}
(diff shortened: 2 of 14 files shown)


helm/router/templates/NOTES.txt
```diff
@@ -1,6 +1,6 @@
 1. Get the application URL by running these commands:
 {{- if .Values.ingress.enabled }}
-  http{{ if $.Values.ingress.tls }}s{{ end }}://{{ .Values.ingress.host | default (printf "www-origin.%s.%s.%s" .Release.Namespace .Values.govukEnvironment .Values.govukDomainExternal ) }}{{ .path }}
+  http{{ if $.Values.ingress.tls }}s{{ end }}://{{ .Values.ingress.host | default (printf "www-origin.%s.eks.%s.%s" .Release.Namespace .Values.govukEnvironment .Values.govukDomainExternal ) }}{{ .path }}
```
sengi, line 3: Just a heads-up that this naming scheme won't work with TLS (unless/until we set up a k8s operator for issuing certs), because a single wildcard cert can only match one level unfortunately. I'll change it in my PR shortly.

helm/content-store/templates/configmap.yaml
```diff
@@ -30,6 +30,8 @@ data:
         listen {{ .Values.nginxPort }};
 
         location / {
+          proxy_set_header   Host $host;
```
sengi, line 33: #sadpanda that we have to do this - I'm guessing it's a Rails thing?

sengi (approved):
Thanks for the cleanup.

helm/content-store/templates/configmap.yaml
fredericfran-gds: I got caught with that in Publisher. Basically, some rails app uses the host header to create a redirect link. e.g. publisher redirects to authentication URL.

helm/router/templates/NOTES.txt
fredericfran-gds: ok cool, we should decide how we are going to manage the domains, will we be able to deploy stuff in namespaces and map it accordingly in domains?

helm/content-store/templates/configmap.yaml
sengi: Aha, thanks that's good to know.

helm/router/templates/NOTES.txt
sengi: I don't think there's anything much to decide at this point. As long as we're able to test stuff for the purposes of getting this project done, we're fine. We don't want to take on the huge task of replacing govuk-docker when we need to focus on the April 2022 deadline.

sengi: Later on, we could potentially have an operator for issuing certs, which would enable deploying a complete Publishing stack into a separate namespace and still have TLS work.

PR #448, opened 2021-09-14
Annotate Ingresses for HTTPS redirect and TLS options.

sengi:
- Configure the http->https redirect (301 Moved Permanently).
- Disable TLS 1.0 and 1.1. This won't affect compatibility for the general public as all connections to the ALBs will be either from the CDN or from publisher (CMS) users.
- Set default human-readable names for the ALBs.
- Get rid of PathType as a Helm var. It has to be Prefix; ExactMatch wouldn't make any sense here.
- Remove nasty template logic for obsolete Ingress versions.
- Tweak the default value for `host` so that it's covered by the wildcard cert. We can't use an extra level for the namespace until we have a controller/CRD for issuing TLS certs.
- Remove `appVersion`; it's optional and it's not feasible for us to have it reflect the actual app version at the moment.
- Fix wrong `PullPolicy` for Router (kinda unrelated but 🤷).

Tested: Deployed all the apps to a new namespace, checked that TLS works as expected: https://www-origin-chris.eks.test.govuk.digital/

I'll fix up the Ingresses in `cluster-services/` in the next PR.

### DIFF
--- a/helm/publisher/Chart.yaml
+++ b/helm/publisher/Chart.yaml
@@ -2,5 +2,4 @@ apiVersion: v2
 name: publisher
 description: A Helm chart for GOV.UK Publisher
 type: application
-version: 0.1.0
-appVersion: "1.16.0"
+version: 0.2.0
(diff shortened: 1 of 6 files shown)


kerin (approved):
LGTM

PR #451, opened 2021-09-15
Specify default TLS policy in the Ingress controller config.

sengi:
- Disable legacy TLS (1.0 and 1.1) by specifying the default TLS policy on the controller command line. This removes the need for the `alb.ingress.kubernetes.io/ssl-policy` annotation on every Ingress and reduces the chance of misconfiguration.
- Update to the latest stable aws-load-balancer-controller Helm chart.

Tested: applied in the test account (`tf apply -var-file ../variables/test/common.tfvars`); controller seems happy (apart from some duplicate LB names which it was already complaining about before) (`k logs -n cluster-services -l app.kubernetes.io/name=aws-load-balancer-controller -f`)


nsabri1 (approved):
LGTM

PR #455, opened 2021-09-30
Set Prometheus to auto-discover all `PrometheusRule`s

karlbaker02:
This PR ensures that Prometheus auto-discovers all `PrometheusRule`s (such as those included as part of application-specific Helm charts) by setting `ruleSelectorNilUsesHelmValues: false` in Prometheus' config.

Currently the Prometheus as created by Prometheus Operator is only checking for `PrometheusRule`s that have the labels of `app: kube-prometheus-stack` and `release: kube-prometheus-stack`, meaning that those labels would have to be applied to any new rules in order for them to be picked up, and it doesn't make sense to apply these labels for anything other than rules for Prometheus itself; [this config is defined in the template file of the Prometheus Operator](https://github.com/prometheus-community/helm-charts/blob/2cacc16807caedc6cabf1606db27e0d78c844564/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml#L202).

To ensure that we instead auto-dicsover all rules, we set `ruleSelectorNilUsesHelmValues: false` which forces the conditional logic to end up outputting `ruleSelector: {}` in Prometheus' config. Whilst we theoretically _could_ just specify `ruleSelector: {}` in this commit (rather than `ruleSelectorNilUsesHelmValues: false`), the problem is that when Helm applies this config, `ruleSelector` would be ignored because it is empty and therefore won't have any actual affect on Prometheus, hence the reason that we need to explicitly set the boolean `ruleSelectorNilUsesHelmValues`.

### DIFF
--- a/terraform/deployments/cluster-services/kube_prometheus_stack.tf
+++ b/terraform/deployments/cluster-services/kube_prometheus_stack.tf
@@ -37,6 +37,10 @@ resource "helm_release" "kube_prometheus_stack" {
           "alb.ingress.kubernetes.io/load-balancer-name" = "prometheus"
         })
       }
+      prometheusSpec = {
+        # Allow empty ruleSelector (https://github.com/prometheus-community/helm-charts/blob/2cacc16807caedc6cabf1606db27e0d78c844564/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml#L202)
+        ruleSelectorNilUsesHelmValues = false
+      }
     }
   })]
 }


sengi (approved):
LGTM.

The only downside of this that I can think of is that we'd need to change it if we want to run more than one Prometheus per cluster, but I think that's highly unlikely and even it it happens, it'd be a very straightforward change. The upside of doing it this way is that we/users don't have to remember to put a magic label on every PrometheusRule.

sengi (commented):
Sorry, I cocked up in my review. You'll need to set `ruleSelectorNilUsesHelmValues: false` in order for this to work, because of [this conditional](https://github.com/prometheus-community/helm-charts/blob/3863f999e583b5dcff008d871a9ff7b4ab559f26/charts/kube-prometheus-stack/templates/prometheus/prometheus.yaml#L199).

sengi:
I'd definitely recommend either keeping the `ruleSelector: {}` to make it clearer what's going on (even though it doesn't functionally change anything for now) and/or comment the `ruleSelectorNilUsesHelmValues: false` to say something like `# Allow empty ruleSelector.` It's an exceptionally poorly-named boolean, that 😅

sengi (approved):
Cool. Thanks for documenting the madness!

PR #456, opened 2021-10-01
Enable lifecycle management of core cluster addons.

sengi:
Enable EKS management of the three core cluster addons which are installed by default: coredns, kube-proxy, vpc-cni.

Without this, EKS still installs the default versions of these addons but doesn't manage their lifecycle, leaving it to the user to manually make any necessary upgrades or changes during a cluster upgrade.

I added a module variable for overriding the default addon versions just in case we ever need to do so (i.e. in the unlikely event that Amazon were to make a bad version the default), but I don't expect we'll actually need it. We don't want to pin these versions because it would defeat the benefit of Amazon managing the core addons for us and make things less reliable and secure overall.

Also stop hardcoding the cluster version, as we'll need it to vary between environments when rolling out cluster upgrades.

Tested: applied in test account, addon versions remain as expected. Checked that overriding a version to the default version produces no diff. Checked that overriding a version to a non-default version produces the expected diff.

[Trello card](https://trello.com/c/r3DYG1Fy/664)

### DIFF
--- a/terraform/deployments/cluster-infrastructure/main.tf
+++ b/terraform/deployments/cluster-infrastructure/main.tf
@@ -47,7 +47,7 @@ module "eks" {
   version = "17.8.0"
 
   cluster_name     = var.cluster_name
-  cluster_version  = "1.21"
+  cluster_version  = var.cluster_version
   subnets          = [for s in aws_subnet.eks_control_plane : s.id]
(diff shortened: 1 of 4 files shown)


kerin (approved):
LGTM 👍

PR #457, opened 2021-10-05
Use managed node group and fix some cluster turnup issues.

sengi:
Turns out we were accidentally using self-managed nodes before (my mistake, in 1f30519). In fairness to myself, the terraform-aws-eks module is quite confusing in this area and the docs are in need of some love.

Also some fixes which I had to make in order to test this change:

* Update to [latest terraform-aws-eks module](https://github.com/terraform-aws-modules/terraform-aws-eks/blob/master/CHANGELOG.md#v17200---2021-09-17).
* Fix a Terraform dependency issue that prevented creation of the TLS cert for the external_dns domain when turning up a cluster from scratch. (I think I broke this in efbe2cf.)
* Set a missing create_namespace to fix a race between helm_release resources where external_dns would fail to install if Terraform happened to run its `helm install` before one of the other helm_release resources had created the namespace. (`--create_namespace` itself is safe to run in parallel.)

Tested: brought up a test cluster using `terraform workspace`: https://eu-west-1.console.aws.amazon.com/eks/home?region=eu-west-1#/clusters/chris

### DIFF
--- a/terraform/deployments/cluster-infrastructure/external_dns.tf
+++ b/terraform/deployments/cluster-infrastructure/external_dns.tf
@@ -79,28 +79,33 @@ resource "aws_acm_certificate" "cluster_public" {
   }
 }
 
-# See https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/acm_certificate_validation#example-usage
+# See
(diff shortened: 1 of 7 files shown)


sengi:
Force-pushed for `terraform fmt` fixes.

nsabri1 (approved):
LGTM

PR #458, opened 2021-10-05
Add build-base-image-definition task

rtrinque:
Following 3fbbc90b045402eb7444252160cf3f2af0b031d7 we can't use the build-image-defition task anymore to build base images. build-base-image-definition replace it


sengi (approved):
I see - the two tasks are basically identical except that build-image-definition has base-image as an input. Seems like there ought to be a nice way to just pass that in, but maybe there isn't 🤷

PR #475, opened 2021-11-08
Update kube-prometheus-stack to 19.0.3

sengi:
No config changes required; see [changelog](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#from-18x-to-19x).

This doesn't actually upgrade Prometheus to the latest version, just the Helm chart. The upgrade story is actually kinda interesting, but that's for another day.

Tested: deployed in the test cluster; Prometheus still works fine and there are no errors in the operator logs.


karlbaker02 (approved):
LGTM 👍

PR #476, opened 2021-11-08
Add Argo Workflows and Events

bilbof:
This commit installs Argo Workflows and Events also replaces `argocd-config` with a helm chart `argo-services`, since this now configures the config for all Argo services (cd, cd-notifications, events). It also removes the ArgoCD Notifications configuration from Terraform, as this will be configured in `argo-services`.

Co-Authored-By: @fredericfran-gds 

Trello: https://trello.com/c/h8F7nBbu/676-smoke-tests-after-each-deploy-by-app

### DIFF
--- a/terraform/deployments/cluster-services/argo.tf
+++ b/terraform/deployments/cluster-services/argo.tf
@@ -55,17 +55,19 @@ resource "helm_release" "argo_cd" {
   })]
 }
 
-resource "helm_release" "argo_config" {
-  depends_on = [helm_release.argo_cd]
-  chart      = "argocd-config"
-  name       = "argocd-config"
+resource "helm_release" "argo_services" {
+  # Relies on CRDs
+  depends_on = [helm_release.argo_cd, helm_release.argo_events]
+  chart      = "argo-services"
+  name       = "argo-services"
   namespace  = local.services_ns
   repository = "https://alphagov.github.io/govuk-helm-charts/"
-  version    = "0.1.2" # TODO: Dependabot or equivalent so this doesn't get neglected.
+  version    = "0.1.1" # TODO: Dependabot or equivalent so this doesn't get neglected.
   values = [yamlencode({
     # TODO: This TF module should not need to know the govuk_environment, since
     # there is only one per AWS account.
     govukEnvironment = var.govuk_environment
+    argocdUrl        = "https://${local.argo_host}"
   })]
 }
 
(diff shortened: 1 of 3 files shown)


terraform/deployments/cluster-services/argo.tf
```diff
@@ -76,76 +78,45 @@ resource "helm_release" "argo_notifications" {
...
-        },
-        {
-          when    = "app.status.operationState.phase in ['Error', 'Failed']"
-          oncePer = "app.status.sync.revision"
-          send    = ["send-slack"]
-        },
-      ])
+    workflow = {
+      serviceAccount = {
+        create = true # Specifies whether a service account should be created
```
sengi, line 104: Please remove redundant comment
```suggestion
        create = true
```

terraform/deployments/cluster-services/argo.tf
```diff
@@ -76,76 +78,45 @@ resource "helm_release" "argo_notifications" {
...
+  chart      = "argo-events"
+  name       = "argo-events"
+  namespace  = local.services_ns
+  repository = "https://argoproj.github.io/argo-helm"
+  version    = "1.7.0" # TODO: Dependabot or equivalent so this doesn't get neglected.
+  values = [yamlencode({
+    namespace                        = local.services_ns
+    "sensorController.extraEnv"      = [{ "name" : "DEBUG_LOG", "value" : "true" }]
+    "eventsourceController.extraEnv" = [{ "name" : "DEBUG_LOG", "value" : "true" }]
+    "eventbusController.extraEnv"    = [{ "name" : "DEBUG_LOG", "value" : "true" }]
```
sengi, line 120: Don't forget to remove these - you mentioned you didn't intend to leave them in.

terraform/deployments/cluster-services/main.tf
```diff
@@ -56,6 +56,7 @@ provider "helm" {
 }
 
 locals {
+  apps_ns                = "apps"
```
sengi, line 59: It doesn't smell right to me that the cluster Terraform stuff should be aware of which namespace the apps are running in - I'd say that's a fairly major layering violation between base cluster services and userland. I'd suggest making this a var and renaming it to something like `argo_workflow_namespaces` to reflect what it's actually doing (and make it a list and merge it on argo.tf line 99. It'd be very unfortunate if someone were to use `apps_ns` for other things in this TF module.

sengi (approved):

terraform/deployments/cluster-services/argo.tf
```diff
@@ -76,76 +78,45 @@ resource "helm_release" "argo_notifications" {
   repository = "https://argoproj.github.io/argo-helm"
   version    = "1.5.1" # TODO: Dependabot or equivalent so this doesn't get neglected.
   values = [yamlencode({
-    argocdUrl = "https://${local.argo_host}"
+    # Configured in argo-services Helm chart
+    "cm.create" = false
```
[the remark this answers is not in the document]
bilbof: Thanks, I've changed this. Interestingly this _seemed_ to work when using `helm_release`, though I'm probably misremembering, since it would be strange if this resource used `set` (e.g. `--set=cm.create=false`) behind the scenes (especially given there's a `set {}` block on the resource!).

bilbof:
I'll wait for https://github.com/alphagov/govuk-helm-charts/pull/20 to be merged before merging this one.

PR #478, opened 2021-11-10
Fix terraform.sh to use common.tfvars

fredericfran-gds:
The terraform.sh that is used by the concourse pipelines does not use the common.tfvars that is common across all gov.uk environment, it only uses the common one inside an environment. This commit fixes that.


bilbof (approved):

fredericfran-gds:
close in favour of: https://github.com/alphagov/govuk-infrastructure/pull/478. It is the same

PR #479, opened 2021-11-15
Rename integration cluster turnup pipeline to satisfy Concourse.

sengi:
Concourse appears to require that the pipeline YAML config filename ends in `.yml` and not `.yaml`.

Was failing with `file 'concourse/parameters/integration/eks.yml' not found within artifact 'govuk-infrastructure'` despite the correct filename being specified when `fly sp` was run.

[Trello card](https://trello.com/c/Rp2xjAYJ/674)


nsabri1 (commented):
LGTM

nsabri1 (approved):

PR #480, opened 2021-11-15
Add Terraform backend configs for the integration environment.

sengi:
Also fix an inconsistency in which S3 bucket we're using for integration Terraform state.

[Trello card](https://trello.com/c/Rp2xjAYJ/674)


rtrinque (approved):
lgtm

PR #482, opened 2021-11-16
cluster-services shouldn't depend on GOV.UK-specific stuff.

sengi:
The cluster-services Terraform module isn't supposed to depend on any application-specific stuff, so it shouldn't need to wait for the govuk-publishing-infrastructure module.

This is more of a documentation fix than anything functional; it fixes the graph on the Concourse dashboard so that it's no longer misleading.


nsabri1 (approved):
LGTM

PR #483, opened 2021-11-17
Include "redis" in the DNS name of the Redis cluster.

sengi:
Looks like nothing much is using this yet so it's a good time to rename.

This changes the names from e.g. `shared-eks.integration.govuk-internal.digital` (which conveys no clue that it's a Redis cluster 😅) to `shared-redis.eks.integration.govuk-internal.digital`.

We can't just set `shared_redis_cluster_name` to "shared-redis.eks" because Redis doesn't allow dots in replica names, hence we're appending it at the point where we specify the Route53 CNAME record.

[Trello](https://trello.com/c/Rp2xjAYJ/674)

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/shared_redis.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/shared_redis.tf
@@ -41,7 +41,8 @@ resource "aws_elasticache_replication_group" "shared_redis_cluster" {
 
 resource "aws_route53_record" "shared_redis_cluster_internal_service_record" {
   zone_id = local.internal_dns_zone_id
-  name    = var.shared_redis_cluster_name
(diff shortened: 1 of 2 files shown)


karlbaker02 (approved):
After initially being worried about whether or not this would have implications for any certificate we're using (presumably issued under `integration.govuk-internal.digital` and how the wildcard domain would extend two levels down, I've come to realise that this should be a none-issue due to the fact that Redis isn't and shouldn't be publicly accessible and as such we're probably not going to be using TLS when sending stuff to Redis.

sengi:
> we're probably not going to be using TLS when sending stuff to Redis.

That's a really good point - I'd kinda just assumed that we weren't using TLS for this. Turns out we're not, so all good (for now at least - and we can always drop the `.eks` if we need to).

PR #484, opened 2021-11-17
Bump `argo-services` Helm chart to `0.1.2`

karlbaker02:
This PR bumps the `argo-services` Helm chart to version `0.1.2`, fixing various errors with the `argocd-notifications-cm` when running a Terraform apply of the form:

```
Error: error validating "": error validating data: [ValidationError(ConfigMap.data.context): invalid type for io.k8s.api.core.v1.ConfigMap.data: got "map", expected "string"
```
(see the commit https://github.com/alphagov/govuk-helm-charts/commit/72911bce9a96f3d9b52c86566ee00249297d6cd2#diff-5ed1a66d94c1c1e2851d74e0df3ebe7195f2732641d67e36a73ef197b9afd2bb for details)

As described in this ConfigMap, at some stage in the future we should have Dependabot or something which operates in a similar fashion keep our charts up-to-date, at least with patch versions and probably too minor versions.

### DIFF
--- a/terraform/deployments/cluster-services/argo.tf
+++ b/terraform/deployments/cluster-services/argo.tf
@@ -62,7 +62,7 @@ resource "helm_release" "argo_services" {
   name       = "argo-services"
   namespace  = local.services_ns
   repository = "https://alphagov.github.io/govuk-helm-charts/"
-  version    = "0.1.1" # TODO: Dependabot or equivalent so this doesn't get neglected.
+  version    = "0.1.2" # TODO: Dependabot or equivalent so this doesn't get neglected.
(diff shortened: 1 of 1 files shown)


sengi (approved):
Thanks for spotting this!

sengi:
I wonder if on balance it'd be the lesser of evils to use latest for Helm charts that we ourselves own, until we have the Dependabot plugin or whatever.

sengi:
Probably not quite that simple though and needs a bit of thought even for an interim solution.

karlbaker02:
One to raise as a team discussion perhaps.

PR #485, opened 2021-11-17
Update GitHub.com SSH known hosts for ArgoCD

bilbof:
On November 16 2021 (yesterday) GitHub [expired an SSH key](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/githubs-ssh-key-fingerprints) `SHA256:br9IjFspm1vxR3iA35FWE+4VTyz1hYVLIE2t1/CeyWQ` (DSA).

ArgoCD maintains a ConfigMap containing SSH known_hosts, and by default their Helm chart ships with various fingerprints for github, gitlab, etc.

Unfortunately, when GitHub removed their DSA key this caused an SSH error when Argo attempted to sync with our GitHub repos: "key mismatch error".

The fix is to remove the expired key from the ConfigMap, as done here. Now the ConfigMap matches the result of `ssh-keyscan github.com`, and there's no longer a mismatch.


sengi (approved):
Thanks for finding and fixing 👍

PR #487, opened 2021-11-17
Argo has updated its GitHub SSH known_hosts; remove our workaround.

sengi:
As of [argo-helm](https://github.com/argoproj/argo-helm) PR 1018 (intentionally not linking to avoid linkspamming that PR), the official Helm chart no longer contains the offending expired DSA key.

Reverts #485 and updates `argo-cd` to the latest chart version.

[Trello](https://trello.com/c/Rp2xjAYJ/674)


bilbof (approved):
Awesome thanks, wow they did that quickly :)

PR #488, opened 2021-11-17
Reduce DNS negative cache time from 15m to 5m.

sengi:
This makes testing less frustrating. Also matches the default TTL.

Also remove an embarrassingly wrong comment. I must have been tired when I wrote 5c2839c 😅


rtrinque (approved):
Looks good

PR #493, opened 2021-11-29
Provide env-specific Logit port to Filebeat

bilbof:
The logit endpoint and port are unique to an environment, so I've set both the hostname and the port in SecretsManager.


sengi (approved):
Makes sense (given than the hostname - bizarrely - is already kind of a secret 🙃) 👍

PR #496, opened 2021-12-02
Various fixes to external-dns configuration.

sengi:
See commit messages for more details.

* Update the external-dns chart to the latest version.
* Use cluster ID for external-dns ownership record, not zone ID.
* Reduce external-dns's polling interval to 5m and enable watches, so that we don't blow through the 5 qps hard limit on Route53.

Rollout: requires deleting the A/TXT records in the external-dns-managed zone after deploying (so that external-dns recreates them with the new ownership details in the TXT records).

Tested: works in the test cluster; controller logs look good; records were recreated after deletion. Editing an Ingress triggered the reconciliation loop without waiting 5 minutes.


kerin (approved):
LGTM 👍

PR #499, opened 2021-12-06
Rewrite Signon Bootstrap task for Kubernetes

bilbof:
The signon:bootstrap rake task was originally written
to be run in Concourse as part of the deploy pipeline
when bringing up applications in ECS, with all secrets
stored and managed in AWS SecretsManager.

We have pivoted to using Kubernetes (EKS) rather than ECS
and so we have updated the code here in two ways:

1. Removed the AWS dependency
   Rather than write to AWS SecretsManager we'll create Secrets
   directly in Kubernetes. This enables local development.
2. Removed the Lambda/Concourse requirement. This code will run
   in a container in-cluster, as a Job, with arguments provided
   via env vars in Helm. We won't need external depedencies such
   as Concourse or Lambda to bring up Signon and its resources.

Associated work:
1. Modified endpoints in Signon's API to support the new usage: https://github.com/alphagov/signon/pull/1768
2. Added a Helm chart which runs this rake task: https://github.com/alphagov/govuk-helm-charts/pull/60

Trello: https://trello.com/c/fUXJ6X1U/725-create-signon-resources-when-signon-is-deployed


sengi (commented):
This is looking great — thanks for porting this to k8s Secrets.

lib/kubernetes/client_factory.rb
```diff
@@ -0,0 +1,29 @@
...
+          kubernetes_uri,
+          version,
+          {
+            auth_options: {
+              bearer_token_file: bearer_token_file,
+            },
+            ssl_options: ssl_options,
+          },
+        ),
+        namespace: File.read(namespace_file),
```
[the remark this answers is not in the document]
bilbof: Discussed offline. We've decided to use Kubeclient despite this unfortunate bit of code, as it is appears to be a more mature gem (in terms of adoption, version, and maintenance).

docker/images/infra-task/Dockerfile
```diff
@@ -0,0 +1,15 @@
+ARG base_image=ruby:2.7.3
```
sengi, line 1: Can we use the latest patch version of the base image? (I'm assuming we have control of this because we're just running our code here as opposed to one of the GOV.UK apps that's stuck on an old patch release?)

Also worth trying the `-slim` version as it saves quite a lot on the image size.
```suggestion
ARG base_image=ruby:2.7.5-slim
```

docker/images/infra-task/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+FROM ${base_image}
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential && apt-get clean
```
sengi, line 3:
```suggestion
# TODO: use the govuk-ruby base/builder images once they're ready, so that we
# can avoid bloating the image with package updates.
RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y build-essential && apt-get clean
```

lib/signon/client.rb
```diff
@@ -4,74 +4,90 @@
...
+      res = do_request(req)
       raise ApplicationNotFound, "Status: #{res.code}; #{res.message}; #{res.body}" unless res.code == "200"
 
       JSON.parse(res.body)
     end
 
+    def create_api_user(name:, email:)
+      req = build_post_request("/api-users", { name: name, email: email })
+      res = do_request(req)
+      raise ApiUserAlreadyCreated if res.code_type == Net::HTTPBadRequest
```
sengi, line 74: Was it intentional that the API returns 400 when trying to create an already-existing ApiUser but 409 on trying to create an already-existing Application? Is it a legacy thing that's a pain to fix?

lib/signon/client.rb
```diff
@@ -102,19 +130,18 @@ def attempt(&request)
       end
     end
 
-    def do_request(request, uri)
+    def do_request(request)
       request["Authorization"] = "Bearer #{@auth_token}"
       request["Content-Type"] = "application/json"
-      Net::HTTP.start(uri.hostname, uri.port, use_ssl: true) do |http|
+      # NOTE: SSL is not enabled since requests stay in-cluster and we have not
```
sengi, line 136: Thanks for documenting this!

lib/tasks/bootstrap.rake
```diff
@@ -1,72 +1,66 @@
...
+  }.flatten
+
+  Signon::Bootstrap.create_bearer_tokens(
+    bearer_tokens: bearer_tokens,
+    signon: signon,
+    kubernetes: kubernetes,
+  )
+
+  puts <<-HEREDOC
+    Generated Signon resources.
```
sengi, line 58: optional: I think you could easily make the output machine-readable by making this line and the headings into JSON comments (i.e. just prefixing with `# ` or however you'd escape that in Ruby) and putting a `---` line at the start of each section.

spec/lib/kubernetes/client_factory_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+        bearer_token_file: "/tmp/bearer_token",
+        crt_file: "/tmp/crt_file",
+        namespace_file: "/tmp/namespace",
+      }
+    end
+
+    let(:kubeclient) { double(Kubeclient::Client) }
+    let(:secrets_client) { double(Kubernetes::Client) }
+
+    it "inits the kubeclient" do
```
sengi, line 24: or it gets the hose again? 😝

sengi (approved):
Just a minor thing about the base image version but otherwise LGTM 👍 

Thanks for revamping this so quickly!

sengi:
Oh I forgot to mention I'm not really able to review this for Ruby style (so if you really wanted a Ruby-focused review you could add another reviewer, but that's entirely up to you).

lib/signon/client.rb
bilbof: Yes, thanks for flagging that. I've added a commit to the Signon API PR to fix this annoying inconsistency and added some detail in the commit message (https://github.com/alphagov/signon/pull/1768/commits/eb71ff82b173f615eee06e6028916e0be9ef3879).

Essentially the ApiUser and the Application models are intertwined with different gems (Devise and Doorkeeper respectively) which validate models inconsistently, which has leaked into this status code. The commit I've put up should resolve that inconsistency and I can now change the client to check for a 409.

sengi: Cool, thanks for fixing that! 💛

lib/tasks/bootstrap.rake
bilbof: Sure thing, I've changed the output to a more useful structure: `{"applications":["content-store","frontend","publishing-api"],"api_users":["content-store","frontend"],"bearer_tokens":["signon-token-content-store-publishing-api","signon-token-frontend-content-store","signon-token-frontend-publishing-api"]}`

docs/signon-secrets.md
```diff
@@ -12,69 +12,41 @@ about what these credentials are for.
...
+Signon credentials are generated automatically when bringing up a new cluster.
 
-These credentials are generated automatically when bringing up a new workspace.
+Signon credentials are pulled from a Secret by Kubernetes when starting a
+container. If the Secret does not exist, then the container won't start and
+Kubernetes or Argo may re-attempt the operation.
 
-The `govuk-publishing-platform` terraform deployment creates the secretsmanager
-secret for each required credential that Signon will later generate.
+The `signon-resources` Helm chart contains a Job that creates a corresponding
```
hannako, line 21: Would be nice to link to the code defining the job here.

PR #500, opened 2021-12-08
Add security group rules to allow connections to legacy databases.

sengi:
Allow connections from nodes (including all pods) to:

- Router MongoDB (running on EC2 router-backend VMs)
- Shared DocumentDB (which content-store uses)
- Frontend memcache

Intentionally not adding any others preemptively, because I think it's better to add them at the same time as adding the clients, so we can test that the rules work.

Tested: removed the manually-added allow-all-from-10/8 rules (on govuk_router-backend_access, govuk_frontend_cache_access, govuk_mongo_access) in the test environment, checked that frontend, router and content-store break and complain about database connectivity; ran tf apply and checked that all three apps work again and logs look good.

### DIFF
--- a/terraform/deployments/cluster-infrastructure/outputs.tf
+++ b/terraform/deployments/cluster-infrastructure/outputs.tf
@@ -8,6 +8,11 @@ output "worker_iam_role_arn" {
   value       = module.eks.worker_iam_role_arn
 }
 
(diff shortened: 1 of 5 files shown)


nsabri1 (approved):
LGTM

PR #501, opened 2021-12-08
Split Terraform and Ruby CI actions into separate jobs.

sengi:
This speeds up the CI runs significantly by:

- Allowing the Ruby lint and tests to run in parallel with the Terraform lint/validation.
- Skipping the Ruby checks if no Ruby-related stuff has changed; similarly for Terraform.
- Only running `bundle install` once per run instead of twice.
- Only running `terraform init` once per root module per run instead of twice.
- Only fetching the latest revision from Git instead of all the history.

Also:

- Enables linting and validation for some Terraform modules which were missing checks before.
- Checks all Terraform root modules in `/terraform/deployments` without requiring us to add a magic `.tflint.hcl` config file when we create a new one.
- Fixes the problem where the Terraform checks abort at the module which has an error. Now continues to check the rest of the modules so you don't have to keep re-running the job.
- Uses terraform-linters/setup-tflint instead of rolling our own. This gives us [Problem Matchers] so that lint issues show up inline on PRs.
- Fixes the problem where a failed tflint install produces no useful error message.

Tested: Pushed a Ruby commit; Ruby checks ran. Pushed a Terraform commit; Terraform checks ran. Terraform checks continue to check all root modules rather than stopping at the first module which has an error.

[Problem Matchers]: https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md


.github/workflows/ci-ruby.yml
```diff
@@ -0,0 +1,13 @@
+name: Ruby linting and tests
+on:
+  push:
+    paths: ['**/*.rb', '**/*.rake', '.github/workflows/ci-ruby.yml']
```
bilbof, line 4: Looks good. There are a few more Ruby files to add: `.ruby-version`, `.rspec`, `.rubocop.yml`, `Rakefile`, `Gemfile`.

sengi: Aha, thanks

.github/workflows/ci-terraform.yml
```diff
@@ -0,0 +1,97 @@
...
+    paths: ['**/*.tf', '**/*.hcl', '.github/workflows/ci-terraform.yml']
+env:
+  TF_IN_AUTOMATION: true
+  TF_PLUGIN_CACHE_DIR: ${{ github.workspace }}/.terraform.d/plugin-cache
+jobs:
+  tflint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      name: Checkout source code
```
fredericfran-gds, line 13: maybe checkout the latest commit only, i.e.:
with:
          fetch-depth: 0

.github/workflows/ci-ruby.yml
bilbof: I'm happy if we want to push this stuff down into a subdirectory (hard to name though, `scripts` or `ruby` both feel too generic, though we have directories for `docker` and `terraform`)

.github/workflows/ci-terraform.yml
sengi: That actually causes it to fetch all revisions. The default is `fetch-depth: 1`, which fetches only the latest revision. See https://github.com/actions/checkout#checkout-v2

fredericfran-gds: thanks, I got confused on that one.

.github/workflows/ci-ruby.yml
sengi: Yeah, I wondered about moving stuff around but thought it's probably more robust to have the job not care where the source files are.

bilbof (approved):
Thanks for doing this! 🚀 

Looks like a reasonable change. It's nice having these as multiple actions, it's much easier to understand the various jobs and steps.

According to the GitHub docs if we break up the steps into separate `jobs` they will [run in parallel](https://docs.github.com/en/actions/learn-github-actions/understanding-github-actions) (e.g. you can run `lint` and `test` concurrently). Steps seem to run serially. Not necessary for this PR though!

sengi:
> if we break up the steps into separate jobs they will run in parallel (e.g. you can run lint and test concurrently).

That's a good idea - I'll add that to my rainy-day TODO list. I think Jobs run in separate containers though so to share stuff you have to use caches. I'm pretty new to GH Actions though so I'm probably not making the best use of it yet.

PR #502, opened 2021-12-08
Give the EKS environment its own frontend memcache.

sengi:
We can't share the Frontend memcache with the EC2 environments, because Frontend uses it to cache rendered HTML which contains links to Rails assets. My mistake 🤦

Also:

- Tweak the config for the shared Redis while we're there; it's very similar to the memcache one so let's make the two consistent.
- Remove the defaults for instance types and added those vars to the per-environment `.tfvars`, to make it less likely that we'll forget to set them appropriately when adding staging and prod.
- Hardcode the standard ports for Redis and memcached; we'd never want to run them on nonstandard ports so on balance it's cleaner to get rid of the module variable and save on interface clutter.
- Use a slightly smaller and cheaper instance type for Redis.
- Use provider default tags to simplify tagging the Redis cluster and associated resources. Saves the repeated boilerplate merge on every tagged resource.

Tested: applied successfully in the test account. Clusters/instances show up healthy in the AWS web console.

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/frontend_memcache.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/frontend_memcache.tf
@@ -1,0 +1,43 @@
+locals {
+  frontend_memcached_name = "frontend-memcached-${local.cluster_name}"
+}
+
+resource "aws_elasticache_subnet_group" "frontend_memcached" {
+  name       = "frontend-memcached"
+  subnet_ids = local.elasticache_subnets
+}
+
+resource "aws_security_group" "frontend_memcached" {
+  name        = local.frontend_memcached_name
+  vpc_id      = local.vpc_id
+  description = "${local.frontend_memcached_name} memcached instance"
+  tags = {
+    Name = local.frontend_memcached_name
+  }
+}
+
+resource "aws_elasticache_cluster" "frontend_memcached" {
+  cluster_id = "frontend-memcached-${local.cluster_name}"
+
+  engine          = "memcached"
+  engine_version  = "1.6.6"
+  node_type       = var.frontend_memcached_node_type
+  num_cache_nodes = 1 # TODO: consider whether failover is needed
+
+  apply_immediately    = var.govuk_environment != "production"
+  parameter_group_name = "default.memcached1.6"
+  subnet_group_name    = aws_elasticache_subnet_group.frontend_memcached.name
+  security_group_ids   = [aws_security_group.frontend_memcached.id]
+  tags = {
+    Name = local.frontend_memcached_name
+  }
+}
+
+resource "aws_route53_record" "frontend_memcached" {
+  zone_id = local.internal_dns_zone_id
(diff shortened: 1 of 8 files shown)


terraform/deployments/govuk-publishing-infrastructure/frontend_memcache.tf
```diff
@@ -0,0 +1,43 @@
...
+  engine          = "memcached"
+  engine_version  = "1.6.6"
+  node_type       = var.frontend_memcached_node_type
+  num_cache_nodes = 1 # TODO: consider whether failover is needed
+
+  apply_immediately    = var.govuk_environment != "production"
+  parameter_group_name = "default.memcached1.6"
+  subnet_group_name    = aws_elasticache_subnet_group.frontend_memcached.name
+  security_group_ids   = [aws_security_group.frontend_memcached.id]
+  tags = {
```
fredericfran-gds, line 31: do we want to add the default tags too? I saw that you removed it in the Redis ones too so might be expected here.

terraform/deployments/govuk-publishing-infrastructure/outputs.tf
```diff
@@ -1,7 +1,3 @@
 output "shared_redis_cluster_host" {
-  value = aws_route53_record.shared_redis_cluster_internal_service_record.fqdn
-}
-
-output "shared_redis_cluster_port" {
-  value = var.shared_redis_cluster_port
```
fredericfran-gds, line 6: Why I added port before was if this output was used to construct the Redis URL for the apps, the URL has a port in it. I see that now this is hardcoded to its default/conventional port.

terraform/deployments/govuk-publishing-infrastructure/frontend_memcache.tf
sengi: Ah so there's a pretty recent [new feature](https://www.hashicorp.com/blog/default-tags-in-the-terraform-aws-provider) where it now does that for us (see [main.tf line 28](https://github.com/alphagov/govuk-infrastructure/blob/176ae02/terraform/deployments/govuk-publishing-infrastructure/main.tf#L28))

terraform/deployments/govuk-publishing-infrastructure/outputs.tf
sengi: Ah yeah thanks I forgot to mention in the description that I'd changed that. Simplifies things a bit (by eliminating a var that adds no real benefit) since we would never have any reason to run it on a non-standard port these days.

sengi:
Added those few extra bits to the description for posterity 👍

terraform/deployments/govuk-publishing-infrastructure/frontend_memcache.tf
sengi: I could have inlined `local.default_tags` into `provider.aws.default_tags` but I left it because there are still some resource types where the provider defaults don't get applied (because ugly Amazon API reasons) 🤷

PR #503, opened 2021-12-14
Fix ability to apply `ecr` module in the test account.

sengi:
Fix the ability to apply the `ecr` Terraform module in the test account by adding missing backend files. We had added a test ECR config but no backend file for it. This allows us to deploy both of the environments again.

Also a bunch of other cleanup while we're there:

* Use the new S3 bucket for the prod tfstate rather than the legacy one (i.e. govuk-terraform-production, not govuk-terraform-steppingstone-production).
* Instead of declaring and passing a variable for every AWS account that's authorised to pull from the registry, just pass a list of principals. It's simpler, more flexible and easier to understand and maintain.
* Move the vars (actually now just one var) for the ECR module into separate files to make it clear that they're specific to that module. Helps to avoid warnings about unused vars when applying other modules.
* Add the staging account to the puller ACLs for the prod registry, since it was missing.
* Remove the prod account from the puller ACLs for the test registry, since we don't ever want prod or staging to pull from the test registry.
* Add missing backend files for the module. We had added a test ECR config but no backend file for it. This allows us to deploy both of the environments again.
* Add provider default AWS tags for resources created by the module.
* Some style cleanup in ecr/main.tf and add a couple of TODOs.

I wanted to add CNAMEs for the registries to make them more developer-friendly, but unfortunately this can't work because Amazon still doesn't support custom domains on ECR. So we're stuck with ugly, long, opaque URLs in all the image names for the time being.

Tested: applied in the test account, checked that the integration cluster can still pull cross-account from the test ECR.

Rollout: for prod, will need to copy the tfstate to the new location (probably something like: check versioning enabled, apply with old backend config, then apply with new backend config and let TF do the copying, then delete the old file from the old S3 bucket)

### DIFF
--- a/terraform/deployments/ecr/README.md
+++ b/terraform/deployments/ecr/README.md
@@ -1,11 +1,24 @@
-# GOV.UK AWS ECR
+# Elastic Container Registry for GOV.UK in Kubernetes
 
-In the new platform, container images are stored in AWS ECR. The AWS account
-that has been chosen to host the ECR is the GOV.UK production one
+Container images for GOV.UK on Kubernetes are stored in AWS ECR. The registry is hosted in the GOV.UK production AWS account. There is also a test registry (for example for testing IAM changes to the registry itself) in the test AWS account.
 
-## Applying
+## Applying Terraform
 
 ```shell
-gds aws govuk-production-admin -- terraform apply \
- -var-file ../variables/common.tfvars
+# Test registry (for testing changes to the registry config)
+
+gds aws govuk-test-admin -- \
+  terraform init -backend-config test.backend -upgrade
+
+gds aws govuk-test-admin -- \
+  terraform apply -var-file ../variables/test/ecr.tfvars
+
+
+# Production registry (for use by all environments)
+
+gds aws govuk-production-admin -- \
+  terraform init -backend-config production.backend -upgrade
+
+gds aws govuk-production-admin -- \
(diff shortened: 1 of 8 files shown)


nsabri1 (approved):
LGTM

sengi:
State move is done. All working with the new state bucket now.

```
$ rm -r .terraform
$ tf init -backend-config production.backend 

Initializing the backend...

Successfully configured the backend "s3"! Terraform will automatically
use this backend unless the backend configuration changes.

Initializing provider plugins...
- Reusing previous version of hashicorp/aws from the dependency lock file
- Installing hashicorp/aws v3.69.0...
- Installed hashicorp/aws v3.69.0 (signed by HashiCorp)

Terraform has been successfully initialized!

You may now begin working with Terraform. Try running "terraform plan" to see
any changes that are required for your infrastructure. All Terraform commands
should now work.

If you ever set or change modules or backend configuration for Terraform,
rerun this command to reinitialize your working directory. If you forget, other
commands will detect it and remind you to do so if necessary.

GDS10889:ecr chrisbanks$ tf apply -var-file ../variables/production/ecr.tfvars 
aws_iam_policy.pull_from_ecr: Refreshing state... 

...

No changes. Your infrastructure matches the configuration.

...

Apply complete! Resources: 0 added, 0 changed, 0 destroyed.
```

PR #504, opened 2021-12-17
Add production backend file for `terraform-lock` module.

sengi:
We need this because we're already using the `ecr` module in production but it didn't have locking enabled.

Also fix up the docs for this module.

Tested: `ecr` module is able to use state locking.
Rollout: already done.


nsabri1 (approved):
LGTM

PR #505, opened 2021-12-17
Remove the now-defunct concourse-iam Terraform module.

sengi:
Big Concourse is gone so we don't need this any more.

Tested: ran terraform destroy in the test account.
Rollout: needs to be rolled out in the integration account.


nsabri1 (approved):
LGTM

sengi:
Rollout done for integration. That's the last of the cleanup for this one.

PR #506, opened 2021-12-17
Remove disused Concourse pipelines and update docs.

sengi:
Since the ceasing of support for Big Concourse (GDS's shared Concourse service) on 15 Dec, we're no longer using Concourse. We plan to replace our Concourse pipelines with other alternatives (for example GitHub Actions and perhaps a managed CI/CD service for Terraform).

I've made a `concourse` branch from just before this goes in, for convenient reference when we come to port some of the Concourse pipeline to GitHub Actions or whatever.


nsabri1 (approved):
LGTM

PR #508, opened 2021-12-21
Fix security group rules for nodes talking to databases.

sengi:
EKS creates *managed* nodes in the *cluster* SG, not the worker node SG 🙃

Amazon's docs leave much to be desired on this topic, but there are some explanations buried in the UI:

> The Cluster Security Group is a unified security group that is used to
> control communications between the Kubernetes control plane and compute
> resources on the cluster. The cluster security group is applied by
> default to the Kubernetes control plane managed by Amazon EKS **as well as
> any managed compute resources created through the Amazon EKS API**.
>
> Additional cluster security groups control communications from the
> Kubernetes control plane to compute resources in your account. Worker
> node security groups are security groups applied to unmanaged worker
> nodes that control communications from worker nodes to the Kubernetes
> control plane.

Kudos to @bilbof for spotting my mistake and finding the elusive documentation!

### DIFF
--- a/terraform/deployments/cluster-infrastructure/outputs.tf
+++ b/terraform/deployments/cluster-infrastructure/outputs.tf
@@ -8,9 +8,9 @@ output "worker_iam_role_arn" {
   value       = module.eks.worker_iam_role_arn
 }
 
-output "worker_security_group_id" {
-  description = "ID of the security group which contains the worker nodes."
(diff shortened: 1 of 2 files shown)


bilbof (approved):
Great one, thanks for fixing it - love the comment.

bilbof (approved):

PR #517, opened 2022-01-21
Change Argo Workflow default config

fredericfran-gds:
We need to change the default Argo Workflow TTL so
that resources are not wasted storing Argo workflow
artifacts "indefinitely".

Ref:
1. [trello card](https://trello.com/c/N9ER3ukt/755-cleanup-completed-jobs)
2. [argo workflow docs](https://argoproj.github.io/argo-workflows/default-workflow-specs/#setting-default-workflow-values)

### DIFF
--- a/terraform/deployments/cluster-services/argo.tf
+++ b/terraform/deployments/cluster-services/argo.tf
@@ -101,6 +101,17 @@ resource "helm_release" "argo_workflows" {
   values = [yamlencode({
     controller = {
       workflowNamespaces = concat([local.services_ns], var.argo_workflow_namespaces)
+      workflowDefaults = {
+        spec = {
+          activeDeadlineSeconds = 7200
+          ttlStrategy = {
+            secondsAfterSuccess = 432000
+          }
+          podGC = {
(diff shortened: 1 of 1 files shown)


terraform/deployments/cluster-services/argo.tf
```diff
@@ -101,6 +101,17 @@ resource "helm_release" "argo_workflows" {
   values = [yamlencode({
     controller = {
       workflowNamespaces = concat([local.services_ns], var.argo_workflow_namespaces)
+      workflowDefaults = {
+        spec = {
+          activeDeadlineSeconds = 7200
```
bilbof, line 106: This is a tricky one. This timeout will be unexpected. 2 hours seems fine, easy to change later.

bilbof (approved):
Nice one!

terraform/deployments/cluster-services/argo.tf
fredericfran-gds: yeah, I was thinking that most jobs run at most for 1 hour or less afaik. Maybe we can investigate further or add the warning in the docs.

PR #520, opened 2022-01-24
Add Powerusers as new category of k8s users

fredericfran-gds:
A new category of K8s users `powerusers` are added to the GOV.UK
cluster. These users will have admin access to a restricted
number of namespaces, `apps` namespace initially where GOV.UK
apps are deployed. In addition, the users will have read-access
at a cluster level.

A cluster role is used as it is possible to bind it to a set of roles
in different namespaces rather than a single namespace.

Tested by running:
1. `gds aws govuk-integration-poweruser -- kubectl -n apps delete deployment signon` (works - pass)
2. `gds aws govuk-integration-poweruser -- kubectl get all -n apps` (works - pass)
3. `gds aws govuk-integration-poweruser -- kubectl -n cluster-services get all` (works - pass)
4. `gds aws govuk-integration-poweruser -- kubectl -n cluster-services delete deployment.apps/sync-notification-sensor-lvmm5` (unathorised - pass)

Ref:
1. [trello card](https://trello.com/c/IM6wt4Vj/718-create-basic-rbac-structure)
2. [k8s docs](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-example)

### DIFF
--- a/terraform/deployments/cluster-services/aws_auth_configmap.tf
+++ b/terraform/deployments/cluster-services/aws_auth_configmap.tf
@@ -40,6 +40,15 @@ locals {
       groups   = ["readonly"]
     }
   ]
+
+  poweruser_roles_and_arns = data.terraform_remote_state.infra_security.outputs.poweruser_roles_and_arns
+  poweruser_configmap_roles = [
+    for user, arn in local.poweruser_roles_and_arns : {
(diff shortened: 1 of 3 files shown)


sengi (approved):
LGTM, thanks!

PR #522, opened 2022-01-27
Use Workflows rather than workflow for Argo Workflows

fredericfran-gds:
The official name is Argo Workflows (plural) and to keep
consistency, we put the plural name in the url of Argo
Workflows

1. [original pr](govuk-infrastructue pr)


fredericfran-gds:
superseded: https://github.com/alphagov/govuk-infrastructure/pull/523

PR #523, opened 2022-01-27
Use Workflows rather than workflow for Argo Workflows

fredericfran-gds:
The official name is Argo Workflows (plural) and to keep
consistency, we put the plural name in the url of Argo
Workflows and other relevant places.

1. [original pr](https://github.com/alphagov/govuk-infrastructure/pull/521)


sengi (approved):
Cool, thanks for keeping this tidy!

PR #529, opened 2022-02-02
Manage GitHub resources with Terraform

bilbof:
This adds a new GitHub deployment which enables us to manage GitHub resources with Terraform.

Currently it simply permits alphagov repositories with the tag 'govuk' to push images to ECR, by providing access to credentials.

There are various additional features we can add to simplify and automate the configuration of new repositories and deploying new applications to the Kubernetes cluster. This should tie in nicely with the ECR module, though that module could also use the github provider to fetch the repositories to create ECR registries, which would mean we don't need to manually add to a list of applications in the module when we add a new app.

One downside is that folks that use this module need to have org-level permissions. Ideally this deployment would be run by an automated process that sits in-cluster and watches GitHub for tag changes.

### DIFF
--- a/terraform/deployments/github/README.md
+++ b/terraform/deployments/github/README.md
@@ -1,0 +1,44 @@
+# GOV.UK GitHub Infrastructure configuration
+
+> **Note**: Currently this module can only be applied by an alphagov GitHub
+Organisation Admin.
+
+This module configures GitHub resources (currently just GitHub Action Organisation
+Secret Repositories) so that the platform can automatically give permissions to
+repositories with specific tags.
+
+We anticipate that teams will add GitHub tags to their repositories (similar to
+annotations on Kubernetes resources) to enable platform functionality.
+For instance creating an ECR registry and giving permissions to push to the
+registry might be enabled with the repo tag `dockerised`.
+
+This module provides similar functionality to the [govuk-saas-config]
+repository, which configures repositories with sensible defaults such as
+protecting the `main` branch and ensuring the repository is compatible with
+our old EC2-based platform (such as checking Jenkins config).
+This is noted as [tech debt] and we should migrate govuk-saas-config to this
+repository.
+
+Our intent is to replace govuk-saas-config with Terraform configuration.
+
+[govuk-saas-config]: https://github.com/alphagov/govuk-saas-config/tree/master/github/lib
+[tech debt]: https://trello.com/c/mojlsebq/226-we-have-two-tools-for-managing-github-resources
+
+## Applying Terraform
+
+1. Generate access token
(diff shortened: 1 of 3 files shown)


terraform/deployments/github/README.md
```diff
@@ -0,0 +1,23 @@
...
+2. Configure Terraform
+  ```shell
+  gds aws govuk-production-poweruser -- \
+    terraform init -backend-config production.backend -upgrade
+  ```
+3. Plan Terraform
+  ```shell
+  GITHUB_TOKEN=<token> gds aws govuk-production-poweruser -- terraform plan
+  ```
+4. Plan Terraform
```
sengi, line 20:
```suggestion
4. Apply Terraform
```

terraform/deployments/github/README.md
```diff
@@ -0,0 +1,23 @@
+# GOV.UK GitHub
```
sengi, line 1: I think this needs a more specific name and a description of what specifically it's for, especially as we already have [another config-as-code tool that's configuring GitHub](https://github.com/alphagov/govuk-saas-config/tree/master/github). We'll also need to signpost the two tools from each other - and from devdocs - otherwise it's gonna be painful for anyone trying to figure out how to change GitHub settings in future.

terraform/deployments/github/main.tf
```diff
@@ -0,0 +1,41 @@
...
+}
+
+#
+# Gives repositories access to push OCI images to GOV.UK Production AWS ECR
+# NOTE: AWS_GOVUK_ECR_ACCESS_KEY_ID and AWS_GOVUK_ECR_SECRET_ACCESS_KEY are
+# manually created.
+#
+
+data "github_repositories" "govuk" {
+  # TODO: Add more granular tags to repositories (i.e. `dockerised`, `replatformed`)
```
sengi, line 24: It's not clear what this TODO is intended to accomplish - is it really worth requiring people to set another tag on (almost) every GOV.UK repo (including new ones and the time taken troubleshooting it when someone forgets, etc.)?

sengi (approved):
I think this is a good solution, though it introduces a potentially confusing piece of tech debt: splitting our GitHub config across two very different places. I think if we signpost it nicely (here, any relevant places in govuk-saas-config and wherever saas-config is mentioned in devdocs) then it'll be fine though.

sengi:
FWIW I think this is a cleaner approach than govuk-saas-config - TF is a better fit than some Ruby scripts for this kind of thing.

bilbof:
Thanks @sengi. I've updated the commit, created a tech debt card, and added [a PR to flag this on govuk-saas-config](https://github.com/alphagov/govuk-saas-config/pull/101).

PR #530, opened 2022-02-02
Add reusable GitHub workflow for ECR build and push

bilbof:
We had duplicated this config to every repository, making it hard to update the workflow quickly.

This moves the config to a single reusable workflow as in
https://docs.github.com/en/actions/using-workflows/reusing-workflows.

This is not a starter workflow, as you might find in alphagov/.github/.github/workflows.


sengi (approved):
Awesome! Major win for maintainability here and will speed things up for Replatforming.

PR #532, opened 2022-02-03
Update External Secrets Operator to latest version.

sengi:
There are important bug fixes, specifically https://github.com/external-secrets/external-secrets/ 510 and 545 (intentionally not linking directly to avoid GitHub reference spam).


bilbof (approved):
Thank you 👍

PR #533, opened 2022-02-03
Fix broken commands in TF docs

sengi:
Also add some missing instructions for applying some of the modules, some general cleanup and add a bit more background as to why we're using ECR the way we are.


terraform/deployments/terraform-lock/README.md
```diff
@@ -18,11 +18,11 @@ Terraform docs for more detail.
 ## Applying
 
 ```sh
-terraform init -backend-config=<environment>.backend -upgrade
-terraform apply
+ENV=test
```
bilbof, line 21: Does this advise folks to use the test env dynamodb for locking in all envs?

bilbof (approved):
Great job 👍 Thanks so much for this. I have a small question but nothing blocking

terraform/deployments/terraform-lock/README.md
sengi: Fixed

terraform/docs/applying-terraform.md
```diff
@@ -7,36 +7,41 @@ The EKS cluster is deployed via Terraform in two stages. See [adr-3] for backgro
...
+* [`ecr`](../deployments/ecr/README.md) (test and production accounts only)
+* [`github`](../deployments/github/README.md)
+* [`terraform-lock`](../deployments/terraform-lock/README.md)
 
-### Running kubectl
+## Running kubectl
 
 ```sh
 AWS_DEFAULT_REGION=eu-west-1
-eval $(gds aws govuk-test-admin -e)
```
sengi, line 39: For some reason the eval approach stopped working on my setup 🤷 Hence switching to this, which should be more robust.

PR #536, opened 2022-02-04
Use OnWorkflowSuccess PodGC strategy

bilbof:
We want pods to stick around until the ttlStrategy.secondsAfterSuccess.
OnPodCompletion deletes the pod immediately which isn't desirable as
it makes it harder to debug failing workflows.


fredericfran-gds (approved):
Thanks. Maybe we can test the archive functionality later and ensure it still works with the argo cli command

PR #538, opened 2022-02-04
Create AWS ECR repositories for all GOV.UK git repositories

bilbof:
This uses the GitHub API to create an ECR repository for each GitHub repository with the corresponding tags.

### DIFF
--- a/terraform/deployments/ecr/main.tf
+++ b/terraform/deployments/ecr/main.tf
@@ -7,6 +7,10 @@ terraform {
       source  = "hashicorp/aws"
       version = "~> 3.0"
     }
+    github = {
+      source  = "integrations/github"
+      version = "~> 4.0"
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


terraform/deployments/ecr/main.tf
```diff
@@ -21,32 +25,28 @@ provider "aws" {
   }
 }
 
-# TODO: Get rid of this list and just give CI permission to create repos, e.g.
-# with https://github.com/byu-oit/github-action-create-ecr-repo-if-missing
```
richardTowers, line 25: Did you consider giving CI permission to create ECR repos as needed? It'd be nice not to have to apply this terraform whenever we start a new project.

bilbof: Allowing the GHA to create an ECR repo would be nice. We'd still need a method to update and delete ECR repositories (e.g. we want to modify the [`image_scanning_configuration`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ecr_repository#image_scanning_configuration), which would be better achieved with Terraform. Also note that further down this file the ARNs of repos are used to add policies to the IAM role, which would need to be broader (i.e. `arn::.../*` rather than `arn::.../content-store`). Finally, in the `tf/deploys/github` module, we already need to apply TF after a new project is added, in order to provide credentials to the GH Action.

I think a nice solution would be to have an observer component (e.g. a Concourse input) trigger Terraform applies whenever the set of repos with the tags "govuk,docker,archived:false,..." changes.

terraform/deployments/ecr/main.tf
```diff
@@ -21,32 +25,28 @@ provider "aws" {
...
+# NOTE: Uses GITHUB_TOKEN env var, an OAuth / Personal Access Token, for auth
+provider "github" {
+  owner = "alphagov"
+}
+
+data "github_repositories" "govuk" {
+  query = "org:alphagov topic:container topic:govuk fork:false archived:false"
```
sengi, line 34: Thanks for tagging these up with `topic:container`. I think that'll avoid creating a bunch of unused/empty container repos in the registry (which would degrade the UI/UX) so this is a win for usability.

sengi (approved):
SGTM. I guess once we (eventually) have a CD solution for Terraform this will be as nice (or a bit nicer, because continuous reconciling / declarative vs. fire-and-forget) than having GH Actions create the repo configs.

fredericfran-gds (approved):
Good work. Thanks for the manual tagging of the repos, it is excellent work.

PR #540, opened 2022-02-07
Argo fix

fredericfran-gds:
ArgoCD has a bug when ArgoCD does not reload the OIDC config if its
underlying secrets are changed. This was fixed upstream in this
[PR](https://github.com/argoproj/argo-cd/pull/8350) and is now
included in the release candidate v2.3.0-rc5.

The reason why we want this fix is that when a new cluster is
created, the OIDC secrets are only pulled from AWS secret manager
after ArgoCD installs the `cluster-secrets` Helm chart. So ArgoCD
needs to reload its config to take in the new interpolated OIDC
config.

We need to wait for the fix to reach a mainstream release, i.e.
v2.3.0 and the helm chart of Argo to default use this image or
higher in order to remove this chart here.


bilbof (approved):
SGTM

PR #544, opened 2022-02-08
Limit number of repos with ECR push authorisation

bilbof:
We tagged the app GitHub repos which we'll deploy to k8s with
'container'.


fredericfran-gds (approved):
excellent

PR #547, opened 2022-02-10
Create RabbitMQ broker for Kubernetes platform

bilbof:
This is a work-in-progress. We're not yet sure if we want to use an AWS MQ broker before we switchover to Kubernetes. We can use this in place of the unmaintained RabbitMQ cluster running on EC2 in the AWS test env.

Merging this is **not** a decision to migrate RabbitMQ as part of the first cutover.

https://trello.com/c/ZEn9p3Hk/810-migrate-publishing-api-to-eks

### DIFF
--- a/terraform/deployments/variables/integration/common.tfvars
+++ b/terraform/deployments/variables/integration/common.tfvars
@@ -22,6 +22,12 @@ eks_private_subnets = {
   c = { az = "eu-west-1c", cidr = "10.1.32.0/22" }
 }
 
+rabbitmq_subnets = {
+  a = { az = "eu-west-1a", cidr = "10.1.36.0/28" }
+  b = { az = "eu-west-1b", cidr = "10.1.36.16/28" }
+  c = { az = "eu-west-1c", cidr = "10.1.36.32/28" }
+}
+
 govuk_environment = "integration"
 
 publishing_service_domain = "integration.publishing.service.gov.uk"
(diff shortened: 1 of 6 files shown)


terraform/deployments/variables/integration/common.tfvars
```diff
@@ -22,6 +22,12 @@ eks_private_subnets = {
   c = { az = "eu-west-1c", cidr = "10.1.32.0/22" }
 }
 
+rabbitmq_subnets = {
+  a = { az = "eu-west-1a", cidr = "10.200.36.0/28" }
```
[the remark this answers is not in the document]
bilbof: Fixed - sorry, I assumed we would have used the same CIDRs in each environment.

fredericfran-gds (approved):
looks good, thanks

terraform/deployments/variables/integration/common.tfvars
```diff
@@ -22,6 +22,12 @@ eks_private_subnets = {
   c = { az = "eu-west-1c", cidr = "10.1.32.0/22" }
 }
 
+rabbitmq_subnets = {
```
sengi, line 25: Do we actually want dedicated subnets for RabbitMQ? Is there some advantage to doing that that outweighs the admin hassle? Bear in mind at some point we'll have RDS databases and such as well.

bilbof: @sengi thanks for your thoughts, I thought this would be consistent with how we have configured other managed services (elasticache, rds). I saw we have [private_subnet_cidrs](https://github.com/alphagov/govuk-aws-data/blob/master/data/infra-networking/production/govuk.tfvars#L19) of size `/24`, should we reuse those? Maybe you could talk me through the hassle we've had with this in the past, would love to learn 😃

PR #548, opened 2022-02-11
Add Postgres and Mysql security rule for EKS nodes

fredericfran-gds:
This change applies to GOV.UK environment other than `test` since
`test` uses a single Postgres instance for multiple app still.

Note that the `content-data-api` and `transition` postgres are not
covered by this change here as they are in different terraform
projects.


bilbof (approved):
Thanks for fixing 👍 

Would be nice to have separate outputs from the `app_govuk_rds` module for each DB engine, to prevent on redundant SG rules, but we can do it later.

PR #549, opened 2022-02-11
fix EKS output security group

fredericfran-gds:
The wrong security group was outputted for the EKS cluster.

See [doc](https://registry.terraform.io/modules/terraform-aws-modules/eks/aws/latest?tab=outputs)


sengi (approved):
Nice find!

PR #554, opened 2022-02-23
Permit ingress to EC2-hosted Content Store from EKS

bilbof:
This will enable EKS services such as the Frontend app
to send requests to the EC2 hosted Content Store application.


fredericfran-gds (approved):
ok

PR #555, opened 2022-02-25
Raise helpful error when bearer_tokens are misconfigured

bilbof:
We will immediately raise an error when invalid bearer token config is discovered and tell the user which bearer token is invalid.

Currently it takes a while for the error to occur, since invalid requests are made to the Signon API (and these back off exponentially).


fredericfran-gds (approved):
thanks!!!!

PR #556, opened 2022-02-25
Allow EKS nodes to access EC2 Account-API and Email-Alert-API

fredericfran-gds:
This is needed because of the spin-up of the email-alert-frontend
in EKS and it is dependent on the EC2 Account-API and
Email-Alert-API since these are not being spunned in EKS for
milestore one.

Ref:
1. [trello card](https://trello.com/c/0gdHdyUe/833-package-email-alert-frontend-for-migration)


sengi (approved):

sengi:
Merging to resolve hanging diff in Terraform, since the change is already applied in the integration account. Hope that's ok!

PR #558, opened 2022-03-01
Add Terraform for `app-assets-$env` S3 buckets.

sengi:
Define the S3 buckets for app assets (CSS, JS etc.) in Terraform.

These buckets (one per AWS account / environment) are for storing and serving app assets such as CSS, JavaScript, logo images and so on. Asset files are uploaded when an app is deployed.

See https://github.com/alphagov/govuk-helm-charts/pull/178 for more background.

Also [update `govuk-publishing-infrastructure` to v4 of the AWS provider](https://github.com/alphagov/govuk-infrastructure/commit/932ae88cb5a97712a47c7eb9a2282bcc91e1f80f) because v4 includes significant changes to the S3 resources so it makes sense to upgrade now rather than define a bunch of S3 stuff using the old provider and have to port it to v4 later. Since we don't yet have any S3 stuff defined, the upgrade requires only [trivial fixes](https://github.com/alphagov/govuk-infrastructure/commit/da2a2a100b907071948b3baa1abd323cb506f2e2).

And a trivial (but annoying and unrelated) [fix](https://github.com/alphagov/govuk-infrastructure/commit/dbe66684a05469b8195101f26f41ab61a70f1661) to the MQ config to work around a TF provider bug in order to apply the `govuk-publishing-infrastructure` module. This will be a game of whack-a-mole until the bug gets fixed upstream or we remove the Amazon MQ instance (which I think we should - see commit message).

Trello: https://trello.com/c/ZR6s2YhF/839

#### Rollout

- [x] The buckets already exist in the integration account (created manually and imported into TF state before applying) and the test account (created by applying this TF).
- [ ] Need to remove the manually-added IAM policy attached to the role associated with the managed nodes' EC2 instance profile.

#### Testing

Applied TF in the test account, verified that upload-app-assets jobs run successfully and that assets can be downloaded via app nginx proxy sidecars.

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/app_assets_s3.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/app_assets_s3.tf
@@ -1,0 +1,52 @@
+resource "aws_s3_bucket" "app_assets" {
+  bucket = "govuk-app-assets-${var.govuk_environment}"
+  tags = {
+    Name        = "App static assets for ${var.govuk_environment}"
+    Environment = var.govuk_environment
+  }
+}
+
+resource "aws_s3_bucket_versioning" "app_assets" {
+  bucket = aws_s3_bucket.app_assets.id
+  versioning_configuration { status = "Suspended" }
+}
+
+resource "aws_s3_bucket_policy" "app_assets" {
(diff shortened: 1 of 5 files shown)


fredericfran-gds (approved):
looks good.

PR #559, opened 2022-03-03
Update Signon bootstrap process to update application configuration

theseanything:
This adds functionality to the signon bootstrap process to update existing application configuration in Signon if diverges from the bootstrapping configuration.

This allows the configuration to be written declaratively.


lib/signon/bootstrap.rb
```diff
@@ -5,22 +5,49 @@ module Signon
...
-        rescue Signon::Client::ApplicationAlreadyCreated
-          signon.get_application(name: app_data["name"])
         end
-        kubernetes.put_secret_value(
-          secret_name: "signon-app-#{app_slug}",
-          secret_data: application,
+      end
+
+      kubernetes.put_secret_value(
+        secret_name: "signon-app-#{app_slug}",
```
bilbof, line 37: Here could we do `secret_name: app_data.fetch("secret_name")`? The existing code seems to ignore the [`secret_name`](https://github.com/alphagov/govuk-helm-charts/blob/main/charts/signon-resources/templates/bootstrap-job.yaml#L28) field we've been diligently adding to apps.

lib/signon/client.rb
```diff
@@ -59,6 +59,18 @@ def create_application(name:, description:, home_uri:, permissions:, redirect_ur
...
     end
 
+    def update_application(id:, name:, description:, permissions:)
+      req = build_patch_request("/applications/#{id}", {
+        name: name,
+        description: description,
+        permissions: permissions,
+      })
+      res = do_request(req)
+      raise ApplicationNotFound, "Status: #{res.code}; #{res.message}; #{res.body}" unless res.code == "200"
```
bilbof, line 69: Raising `ApplicationNotFound` if `res.code != "200"` may not be appropriate here - perhaps we can raise this if it's a 404, otherwise raise a `ApplicationNotUpdated` or something 🤔

lib/signon/bootstrap.rb
theseanything: Good shout, updated!

lib/signon/client.rb
theseanything: Yup - makes sense! I've added error handling for ApplicationNotFound for 404s and ApplicationNotUpdated for everything else.

.byebug_history
```diff
@@ -0,0 +1,6 @@
+q
+Net::HTTPNotFound == 404
+Net::HTTPNotFound
+res.code
+res.code == Net::HTTPNotFound
+res
```
bilbof, line 6: Oops left this debugging file in.

bilbof (approved):
Looks good, just need to remove the byebug file

.byebug_history
theseanything: 😬 ah sorry I should have caught this

PR #560, opened 2022-03-04
Add EFS security rule for EKS nodes

kentsanggds:
## What

Add EFS security rule for EKS nodes, this is needed by asset-manager as a temporary store for uploaded files before they are virus scanned and moved to s3.

## Reference 

https://trello.com/c/pRBFWfGx/688-asset-manager-replatforming


sengi (approved):
LGTM

PR #561, opened 2022-03-04
Set up CI for Signon Resources

bilbof:
Co-opting the old `infra-task` image for `signon-resources`.


fredericfran-gds (approved):
looks good, maybe change the workflow file name too. I would have liked to see the signon-resources app be in its own folder rather than top level.

PR #564, opened 2022-03-07
Enable manually triggering GHA with workflow_dispatch

bilbof:
This also fixes up the glob for the lib directory.


fredericfran-gds (approved):
cool, thanks

PR #566, opened 2022-03-09
Attach search-api s3 bucket policies to EKS nodes

fredericfran-gds:
For search-api to run in EKS, it needs to access 2 S3 buckets.
We export the arn of the iam_policies of these 2 buckets in
[govuk-aws](https://github.com/alphagov/govuk-aws/pull/1558) to
the terraform remote state.

We then attach these iam_policies arns to attach the EKS nodes for
node. Later, we need to attach them to a custom  service account
used to run the search-api pod/container.

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/app_search_s3.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/app_search_s3.tf
@@ -1,0 +1,11 @@
+# TODO: instead of granting write access to nodes, use IRSA (IAM Roles for
+# Service Accounts aka pod identity) so that only search-api containers can write.
+resource "aws_iam_role_policy_attachment" "search_relevancy_s3_eks_policy_attachment" {
+  role       = data.terraform_remote_state.cluster_infrastructure.outputs.worker_iam_role_name
+  policy_arn = data.terraform_remote_state.app_search.outputs.search_relevancy_s3_policy_arn
+}
+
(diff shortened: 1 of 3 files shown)


terraform/deployments/govuk-publishing-infrastructure/app_search_s3.tf
```diff
@@ -0,0 +1,13 @@
+# TODO: instead of granting write access to nodes, use IRSA (IAM Roles for
+# Service Accounts aka pod identity) so that only search-api containers can write.
+resource "aws_iam_role_policy_attachment" "search_relevancy_s3_eks_policy_attachment" {
+  role       = data.terraform_remote_state.cluster_infrastructure.outputs.worker_iam_role_name
+  policy_arn = data.terraform_remote_state.app_search.outputs.search_relevancy_s3_policy_arn
+}
+
+# TODO: instead of granting write access to nodes, use IRSA (IAM Roles for
+# Service Accounts aka pod identity) so that only search-api containers can write.
```
sengi, line 9: trivial suggestion: no need to duplicate this TODO; it's clear that it applies to both.
```suggestion
```

sengi (approved):
👍

terraform/deployments/govuk-publishing-infrastructure/app_search_s3.tf
fredericfran-gds: thanks, I will fix that.

PR #572, opened 2022-03-15
Add asset_manager_s3 policy

kentsanggds:
This policy will enable listing of bucket objects, putting, deleting and getting for govuk assets uploaded to s3


nsabri1 (approved):
LGTM

PR #574, opened 2022-03-17
Change update image tag job to merge branch

theseanything:
This changes the behaviour from pushing commits directly to main. The update image tag reusable workflow now creates a branch and merges the branch into main on GitHub using the REST API. This helps prevents
contention between multiple processes trying to push new commits to main. E.g. multiple GitHub Actions running this workflow or when a PR is merged whilst this workflow is run. If one process beats another in pushing new commits to main, having started with same git history for main, the other process will have its push rejected.


sengi (approved):
LGTM. Seems obvious to me now that I think about it that the rollout commits would race with each other and mostly end up failing when several apps get deployed at the same time 😅

PR #575, opened 2022-03-18
Bump argo-services helm chart

fredericfran-gds:
Needed since new version has been released
(https://github.com/alphagov/govuk-helm-charts/pull/256)
which will allow better post-sync workflow naming.


marcpomfret (approved):
LGTM

PR #579, opened 2022-03-25
Add security rule for EKS workers to access mapit

fredericfran-gds:
This is needed since mapit is still being run in EC2.


nsabri1 (approved):
LGTM

PR #583, opened 2022-03-30
Add kubescape scanner via Helm

lwilts:


sengi (approved):

terraform/deployments/cluster-services/kubescape.tf
```diff
@@ -0,0 +1,24 @@
...
+
+resource "helm_release" "kubescape" {
+  chart      = "armo-cluster-components"
+  name       = "armo"
+  namespace  = "armo-system"
+  repository = "https://armosec.github.io/armo-helm/"
+  version    = "1.7.0"
+  set {
+    name  = "clusterName"
+    value = "govuk-${var.govuk_environment}"
```
sengi, line 17: Can avoid hardcoding the `govuk` like this (if you're not keen on the long name, could add a `cluster_name` in locals in main.tf and use `local.cluster_name`). Helps avoid name conflicts in the case where we want to bring up a temporary cluster for testing turnup-related stuff.
```suggestion
    value = "${data.terraform_remote_state.cluster_infrastructure.outputs.cluster_id}-${var.govuk_environment}"
```

sengi:
Cheers for getting this up and running - sorry it's been a long journey!

PR #586, opened 2022-04-04
Add Grafana persistence

fredericfran-gds:
Add AWS Serverless Mysql Aurora to the infra to act as a backing database for Grafana to persist in its state.


fredericfran-gds:
> LGTM - I do see there is a empty output.tf file - was that intended?

yeah, I copied the structure but not currently used so I will delete that redundant file. thanks.

PR #588, opened 2022-04-04
Update AWS LB controller to latest version.

sengi:
Temporarily preserve the [aws-lb-controller](https://kubernetes-sigs.github.io/aws-load-balancer-controller/)'s permissions to read all secrets, so as not to break things while we upgrade. I'll remove this once alphagov/govuk-helm-charts#348 is merged and confirmed working. Once this is done, the upgrade will be a win for security because the LB controller will no longer have excessive permissions.

There's also a security improvement in the IAM policy: the controller is now only permitted to create its own service-linked role (`elasticloadbalancing.amazonaws.com`) rather than any service-linked role.


sengi:
Superseded by #594. We're not in prod yet so let's just go for it.

PR #590, opened 2022-04-04
Unpin chart versions for charts that we own.

sengi:
During cluster turnup, we always want to use the latest chart versions for charts that we release ourselves. Unlike for third-party charts, long-term pinning of our own chart versions introduces extra toil and risk for no benefit.

Unfortunately we still need to bump the version numbers of these charts when making changes to the charts themselves, otherwise chart-releaser wouldn't build a new chart tarball and the turnup automation (Terraform) would break because it'd be stuck with the last-released version rather than the latest chart from the source repo.


nsabri1 (approved):
LGTM

PR #592, opened 2022-04-05
Add github-cli image

theseanything:
This adds the Dockerfile and a GitHub Action to build the an image for using the github-cli. This need by the update-image-tag Argo workflow. This will improve the security of the workflow by not running a container as root and improve the speed by pre-installing required dependencies.

### DIFF
--- a/.github/workflows/build-github-cli-image.yaml
+++ b/.github/workflows/build-github-cli-image.yaml
@@ -11,34 +11,11 @@ on:
       - "docker/images/github-cli/Dockerfile"
 
 jobs:
-  publish:
-    name: Build image
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-
-      - name: Configure AWS credentials
-        uses: aws-actions/configure-aws-credentials@v1
-        with:
-          # TODO: Remove long-lived keys and switch to OIDC once https://github.com/github/roadmap/issues/249 lands.
-          aws-access-key-id: ${{ secrets.AWS_GOVUK_ECR_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.AWS_GOVUK_ECR_SECRET_ACCESS_KEY }}
-          aws-region: eu-west-1
-
-      - name: Login to Amazon ECR
-        id: login-ecr
-        uses: aws-actions/amazon-ecr-login@v1
-
-      - name: Build, tag, and push image to Amazon ECR
-        id: build-image
-        env:
-          ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
-          ECR_REPOSITORY: github-cli
-          IMAGE_TAG: ${{ github.sha }}
-        run: |
-          # Build a docker container and push it to ECR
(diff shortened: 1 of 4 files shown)


docker/images/github-cli/Dockerfile
```diff
@@ -0,0 +1,21 @@
+ARG base_image=bitnami/minideb:latest
+
+FROM ${base_image}
+
+RUN apt-get update -qq && apt-get upgrade -y && apt-get install -y ca-certificates git
```
sengi, line 5: With minideb we no longer need to run `apt-get upgrade`, because Bitnami keeps the base image up to date for us. This way we stay secure and avoid bloating the image with old versions of packages.

```suggestion
RUN apt-get update -qq && apt-get install -y ca-certificates git
```

.github/workflows/build-github-cli-image.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+      - name: Build, tag, and push image to Amazon ECR
+        id: build-image
+        env:
+          ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
+          ECR_REPOSITORY: github-cli
+          IMAGE_TAG: ${{ github.sha }}
+        run: |
+          # Build a docker container and push it to ECR
+          docker build -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG -f docker/images/github-cli/Dockerfile .
```
sengi, line 41: I think you'll also want to update the `latest` tag so as not to create a confusing situation when people run one-off jobs or want to test stuff manually.
```suggestion
          docker build \
            -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG \
            -t $ECR_REGISTRY/$ECR_REPOSITORY:latest \
            -f docker/images/github-cli/Dockerfile .
```

.github/workflows/build-github-cli-image.yaml
```diff
@@ -0,0 +1,44 @@
...
+          echo "Pushing image to ECR..."
+          docker push $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG
```
sengi, line 43:
```suggestion
          docker push $ECR_REGISTRY/$ECR_REPOSITORY -a
```

.github/workflows/build-github-cli-image.yaml
```diff
@@ -0,0 +1,44 @@
...
+    branches:
+      - main
+  push:
+    branches:
+      - main
+    paths:
+      - "docker/images/github-cli/Dockerfile"
+
+jobs:
+  publish:
```
sengi, line 14: Is there a reason you can't just call https://github.com/alphagov/govuk-infrastructure/blob/main/.github/workflows/ci-ecr.yaml here?

sengi (approved):

.github/workflows/build-github-cli-image.yaml
theseanything: Good call - I've updated the ci-ecr workflow to accept inputs to make it also usable in this context (e.g. ecr repo name was hardcode to the git repo name etc). I've also applied your suggestions above to the ci-ecr workflow.

sengi: Awesome, thanks!

PR #594, opened 2022-04-06
Update EKS module to v18 and fix cluster turnup blockers.

sengi:
* Update EKS module to v18 and fix up security group rules. The nodes now belong to a different security group than the cluster control plane, per [Amazon's guidance](https://docs.aws.amazon.com/eks/latest/userguide/sec-group-reqs.html) and various things needed fixing up to reflect that.
* Fix some dependency issues that were breaking cluster turnup. — mostly around namespace creation.
* Update AWS LB controller from v2.2 to v2.4. This provides some security improvement as the controller now only has access to the secrets that it needs, rather than all secrets the cluster like in the previous version.
* Update Terraform AWS provider to v4.
* When turning up test clusters, use spot instances and only run 3 nodes initially.

See individual commits for more details.

Tested: cluster turnup works in the test account and I was able to create a Deploy, Service and Ingress and serve traffic from it.

### DIFF
--- a/terraform/deployments/cluster-services/variables.tf
+++ b/terraform/deployments/cluster-services/variables.tf
@@ -1,6 +1,13 @@
+variable "apps_namespace" {
+  type        = string
+  description = "Name of the namespace to create for ArgoCD to deploy apps into by default."
+  default     = "apps"
+}
+
 variable "argo_workflows_namespaces" {
   type        = list(string)
   description = "Namespaces in which Argo will run workflows."
+  default     = ["apps"]
 }
 
 variable "govuk_aws_state_bucket" {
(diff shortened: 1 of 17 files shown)


terraform/deployments/cluster-services/variables.tf
```diff
@@ -21,9 +28,11 @@ variable "govuk_environment" {
 variable "dex_github_orgs_teams" {
   type        = list(object({ name = string, teams = list(string) }))
   description = "List of GitHub orgs and associated teams that Dex authorises. Format [{name='github_org', teams=['github_team_name']}] "
+  default     = [{ name = "alphagov", teams = ["gov-uk-production"] }]
```
nsabri1, line 31: is it missing teams = gov-uk?

nsabri1 (approved):
Other then one comment - LGTM

terraform/deployments/cluster-services/variables.tf
sengi: Oh, good catch! So this particular line is right, but I'd deleted the overrides in test and integration by mistake. Fixed. Thanks!

PR #596, opened 2022-04-07
Factor out monitoring module, update Aurora module, switch to Postgres.

sengi:
- Move the Terraform resources into the cluster-infrastructure module, because it fits with the purpose of that module (AWS resources that are required by the cluster and its base services, like monitoring) and it simplifies the cluster turnup process (which is currently manual, so extra manual steps are unwelcome unless we're really getting something worthwhile in return).
- Use Postgres instead of MySQL. (The general direction of travel in GOV.UK is towards Postgres.) 
- Disable scale-to-zero by default (because it takes 30s to start up again and that's an irritation we don't need when troubleshooting production outages) and enable it in the non-prod environments.
- Update to the current version of the Terraform module that creates the RDS instance.
- Allow Terraform to delete/re-create the Secrets Manager secret (for the database password etc.) in the non-production environments so that it doesn't break annoyingly and require manual pokery to get it going again while testing cluster turnup/turndown.
- Update the (manual) cluster turnup instructions.

Tested: works in the test account.


fredericfran-gds (approved):
looks good thanks

PR #599, opened 2022-04-07
Fastly exporter

fredericfran-gds:
fastly-exporter allows Fastly metrics to be exposed in Prometheus
format. It also configures Prometheus to scrape it.


nsabri1 (approved):
LGTM

PR #600, opened 2022-04-08
Add missing security group rule for shared-documentdb.

sengi:
Asset Manager needs this. There was previously a leftover manually-added rule which was allowing this to work by accident.

Tested: applied in integration


nsabri1 (approved):
LGTM

PR #602, opened 2022-04-08
Update Argo helm charts

fredericfran-gds:
The main change is that ArgoCD helm chart now includes
ArgoCD Notifications so we remove the latter Helm chart.

ArgoCD now uses an official release rather than a candidate
release because the bug fix is now in the mainline

Other minor changes are updating Argo Events and Argo
Workflows Helm charts to the latest version.

### DIFF
--- a/terraform/deployments/cluster-services/argo.tf
+++ b/terraform/deployments/cluster-services/argo.tf
@@ -18,15 +18,8 @@ resource "helm_release" "argo_cd" {
   namespace        = local.services_ns
   create_namespace = true
   repository       = "https://argoproj.github.io/argo-helm"
-  version          = "3.32.1" # TODO: Dependabot or equivalent so this doesn't get neglected.
+  version          = "4.5.0" # TODO: Dependabot or equivalent so this doesn't get neglected.
   values = [yamlencode({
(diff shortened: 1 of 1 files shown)


terraform/deployments/cluster-services/argo.tf
```diff
@@ -85,6 +78,18 @@ resource "helm_release" "argo_cd" {
...
+    notifications = {
+      cm = {
+        create = false
+      }
+
+      "argocdUrl" = "https://${local.argo_host}"
+
+      secret = {
+        create = false
+      }
```
sengi, line 91: nit: no need for the quotes around the key there. Also can make this a bit more concise like this: (also in alphabetical order cos why not) 
```suggestion
    argocdUrl = "https://${local.argo_host}"
    cm        = { create = false }
    secret    = { create = false }
```

sengi (approved):
Awesome - thanks for updating these!

terraform/deployments/cluster-services/argo.tf
fredericfran-gds: cool thanks, I thought about the first one and then I forgot :-(

PR #605, opened 2022-04-12
Improve Argo Workflows security

fredericfran-gds:
1. first commit (7f2ea00e2588e69621aab25f982d8dd40d0ab141): use emissary rather than docker as workflow executor
2. second commit (0d55d47ded0342534a2b120e3e9c88f92a9a46a5): do run root as user when running containers


nsabri1 (approved):
LGTM

PR #606, opened 2022-04-12
Enable CD for signon-resources

theseanything:
This updates the GitHub Action (GHA) workflow to have an additional step to update the image tag reference in the argocd-apps chart, so its automatically deployed to integration.

This also update GHA to build the signon-resources image using the existing `ci-ecr` reusable workflow instead having duplicated logic.


fredericfran-gds (approved):
looks good, thanks for cleaning up.

PR #607, opened 2022-04-12
Fix github-cli image

fredericfran-gds:
The user and group use must have id 1001 because the argo-workflows
are defaulting to these when running.


docker/images/github-cli/Dockerfile
```diff
@@ -9,7 +9,7 @@ RUN chmod 644 /usr/share/keyrings/githubcli-archive-keyring.gpg
 RUN echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" > /etc/apt/sources.list.d/github-cli.list
 RUN apt-get update -qq && apt-get install -y gh
 
-RUN groupadd -r user && useradd -m -r -g user user
+RUN groupadd -g 1001 -r user && useradd -u 1001 -m -r -g user user
```
nsabri1, line 12: Seems you just need to `RUN useradd -m -u 1001 -g 1001 user` 
as userad creates a group by default same as the username specified

nsabri1 (approved):
LGTM

docker/images/github-cli/Dockerfile
fredericfran-gds: unfortunately it fails, the group must be present.

PR #612, opened 2022-04-20
Fixes to Argo

fredericfran-gds:
Fixes to Argo:
1. add more memory resources to Argo-workflow as it was OOM
2. add pod security policy to Argo-workflow as it was only set at container level
   by default
3. heredoc update argocd


nsabri1 (approved):
LGTM

PR #624, opened 2022-04-28
Update cluster sizing, volumes; include kubescape deployment

lwilts:


terraform/deployments/cluster-infrastructure/variables.tf
```diff
@@ -59,7 +59,7 @@ variable "force_destroy" {
 variable "workers_instance_types" {
   type        = list(string)
   description = "List of instance types for the managed node group, in order of preference. The second and subsequent preferences are only relevant when using spot instances."
-  default     = ["m5.xlarge"]
+  default     = ["m5.2xlarge"]
```
sengi, line 62: Ah is m6i no good?

If the issue was stockouts when using spot instances, we could still set m6i.2xlarge as the default (for non-spot clusters) and specify a priority list of instance types for the test account (so it'll use whatever's available from the list when trying to get spot instances).

terraform/deployments/variables/test/common.tfvars
```diff
@@ -3,7 +3,7 @@ cluster_infrastructure_state_bucket = "govuk-terraform-test"
 
 cluster_version               = 1.21
 cluster_log_retention_in_days = 7
-workers_default_capacity_type = "SPOT"
+# workers_default_capacity_type = "SPOT"
```
sengi, line 6: Accidental commit? We probably want to continue to use spot instances by default for test clusters.

sengi (approved):

terraform/deployments/cluster-infrastructure/variables.tf
lwilts: cc @marcpomfret

lwilts: Yeah that was the issue - leaving m5 for now to get this change in (too many things in flight at once), but have added the instance type change to the Trello card.

PR #626, opened 2022-04-28
Increase argo workflow pod memory limits

theseanything:
Previous limits may have been too low as containers were being terminated. This doubles the limits of the workflows controller and server pods.


fredericfran-gds (approved):
looks good

PR #627, opened 2022-04-28
Skip dex approval screen to share data with application

theseanything:
This is to save some clicking when logging into an application. At the moment every new session shows a screen that ask for your approval for Dex to share data with the application (argo, grafana etc.). Approval for sharing data from connected IdP to Dex is separate process on IdP.

image


sengi (approved):
Ooh excellent! I was looking for that but didn't manage to find it.

PR #628, opened 2022-04-28
Skip the unnecessary prompt screen when logging into Prometheus etc.

sengi:
See `--skip-provider-prompt` in https://oauth2-proxy.github.io/oauth2-proxy/docs/configuration/overview/


nsabri1 (approved):
LGTM

PR #630, opened 2022-04-29
Improve kube-prometheus-stack config.

sengi:
- Fix a typo (`efaultRegion`) in one of the Helm values.
- Use a more conventional indentation style, consistent with the rest of the repo.
- Factor out some repeated config for the OAuth2 proxies.
- Consistently refer to Alertmanager by its official name (without space or `_`).

Tested: applied in integration and can still log into Prometheus/Grafana/Alertmanager. Deployments are healthy. Also checked the diffs.


fredericfran-gds (approved):
Yay, thanks for that. I was going to planning to do that at some point.

PR #635, opened 2022-05-05
Added worker_instance_types list

marcpomfret:
Added multiple instance types to allow for different sized spot instances to be used in test env.

https://trello.com/c/iKwiAmxK/906-fix-instance-sizing-and-cluster-autoscaling


desenhar (approved):
lgtm

PR #637, opened 2022-05-09
Fix the output for the node security group ID again.

sengi:
Setting `create_launch_template = false` (#624) changes the meanings of the [outputs of the EKS module](https://registry.terraform.io/modules/terraform-aws-modules/eks/aws/latest?tab=outputs). In particular, `module.eks.node_security_group_id` is no longer the SG that contains the worker nodes. The nodes now belong to the cluster primary SG, like they used to before we upgraded the EKS module.

So this should fix all the security group / firewall issues that we've started seeing lately.

Tested: already ran `terraform refresh` to update the outputs in integration and then applied govuk-publishing-infrastructure to update all the SG rules; traffic passes again.


nsabri1 (approved):
LGTM

PR #640, opened 2022-05-11
Enable ArgoCD prometheus metrics

theseanything:
This enables the metrics server for ArgoCD components and adds prometheus serviceMonitors to scrape those endpoints.


marcpomfret (approved):
LGTM

fredericfran-gds (commented):
Great work!

I don't know if we could create a local value (https://www.terraform.io/language/values/locals#when-to-use-local-values) and just use it in every place needed.

For Dex, it is disabled, no need to add metrics.

PR #644, opened 2022-05-17
Add DNS records for www Fastly service to Terraform.

sengi:
I'd created these manually when setting up www.eks on Fastly.

The process of adding a Fastly service is still somewhat manual anyway, but at least this way we won't have to recreate the records manually if we tear down integration/staging and turn it back up again.

https://trello.com/c/gipscM7t

Tested: already applied successfully in the integration and staging accounts (imported the existing records).


nsabri1 (approved):
LGTM

PR #650, opened 2022-05-23
Update cluster-autoscaler from 1.21 to 1.23.

sengi:
Pretty much just bug fixes as far as we're concerned.

cluster-autoscaler changelog: https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler#notable-changes

https://trello.com/c/smNta30K


fredericfran-gds (approved):
thanks

PR #654, opened 2022-05-23
Update kube-prometheus-stack from 0.55.0 to 0.56.2.

sengi:
Also update the OAuth2 proxy from 7.2.0 to 7.2.1.

https://trello.com/c/smNta30K

Rollout: CRDs need updating (because Helm doesn't do it) - see https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#from-34x-to-35x


nsabri1 (approved):
LGTM

PR #658, opened 2022-05-26
Update GHA workflow to allow for update inputs

alexbakervr:
https://trello.com/c/tUyJa4cb/923-extend-gha-to-deploy-to-specific-environment

Adding in additional inputs to allow for the passing of an environment name when running the workflow manually in the source app repo. This will allow us to more easily roll back deployments to environments other than integration.

See Trello card for more detail.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -42,6 +42,11 @@ on:
         required: false
         default: ${{ github.event.repository.name }}
         type: string
+      environment:
+        description: 'Environment to deploy to'
+        required: false
+        default: 'integration'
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yaml
```diff
@@ -37,6 +37,15 @@ on:
...
         default: "false"
         type: string
+      environment:
+        description: 'Environment to deploy to'
+        required: true
+        type: choice
+        options:
+        - integration
+        - staging
+        - production
```
theseanything, line 47: As this is just a workflow input, I don't think we need type and options key?

PR #659, opened 2022-05-27
Create GP3 StorageClass

desenhar:
For https://trello.com/c/Wz9dRNWA/931-create-gp3-storageclass.
Added the [EBS CSI driver Helm chart](https://docs.aws.amazon.com/eks/latest/userguide/ebs-csi.html), this is to create a GP3 StorageClass for the cluster, using the EBS CSI driver as the provisioner.

update, this has been tested and is working correctly.
A pod, persistent volume claim and storage class has been deployed into the default namespace, using gp3 volume type.

### DIFF
--- a/terraform/deployments/cluster-infrastructure/aws_ebs_csi_iam.tf
+++ b/terraform/deployments/cluster-infrastructure/aws_ebs_csi_iam.tf
@@ -1,0 +1,169 @@
+locals {
+  csi_driver_controller_service_account_name = "ebs-csi-controller-sa"
+}
+
+module "aws_ebs_csi_driver_iam_role" {
+  source                        = "terraform-aws-modules/iam/aws//modules/iam-assumable-role-with-oidc"
+  version                       = "~> 4.0"
+  create_role                   = true
+  role_name                     = "${local.csi_driver_controller_service_account_name}-${var.cluster_name}"
+  role_description              = "Role for the AWS EBS CSI driver controller. Corresponds to ${local.csi_driver_controller_service_account_name} k8s ServiceAccount."
+  provider_url                  = module.eks.oidc_provider
+  role_policy_arns              = [aws_iam_policy.aws_ebs_csi_driver.arn]
+  oidc_fully_qualified_subjects = ["system:serviceaccount:kube-system:${local.csi_driver_controller_service_account_name}"]
+}
+
+resource "aws_iam_policy" "aws_ebs_csi_driver" {
+  name        = "AWSEbsCsiController-${var.cluster_name}"
+  description = "Allow the driver to manage AWS EBS"
+
+  # The argument to jsonencode() is the verbatim contents of
+  # https://raw.githubusercontent.com/kubernetes-sigs/aws-ebs-csi-driver/master/docs/example-iam-policy.json
+  # (except for whitespace changes from terraform fmt).
+  policy = jsonencode({
+
+    "Version" : "2012-10-17",
+    "Statement" : [
+      {
+        "Effect" : "Allow",
+        "Action" : [
+          "ec2:CreateSnapshot",
+          "ec2:AttachVolume",
+          "ec2:DetachVolume",
+          "ec2:ModifyVolume",
+          "ec2:DescribeAvailabilityZones",
+          "ec2:DescribeInstances",
+          "ec2:DescribeSnapshots",
+          "ec2:DescribeTags",
+          "ec2:DescribeVolumes",
+          "ec2:DescribeVolumesModifications"
+        ],
+        "Resource" : "*"
+      },
+      {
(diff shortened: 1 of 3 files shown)


terraform/deployments/cluster-services/aws_ebs_csi_driver.tf
```diff
@@ -0,0 +1,33 @@
+resource "helm_release" "csi_driver" {
+  chart            = "aws-ebs-csi-driver"
+  name             = "aws-ebs-csi-driver"
+  namespace        = "kube-system"
+  repository       = "https://kubernetes-sigs.github.io/aws-ebs-csi-driver"
+  version          = "1.6.2" # TODO: Dependabot or equivalent so this doesn't get neglected.
```
sengi, line 6: The [current version of this chart](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/Chart.yaml#L5) appears to be `2.6.8`. (1.6.2 is the app version.)

terraform/deployments/cluster-services/aws_ebs_csi_driver.tf
```diff
@@ -0,0 +1,33 @@
...
+    name  = "controller.serviceAccount.name"
+    value = data.terraform_remote_state.cluster_infrastructure.outputs.aws_ebs_csi_driver_controller_service_account_name
+  }
+  set {
+    name  = "enableVolumeResizing"
+    value = true
+  }
+  set {
+    name  = "node.tolerateAllTaints"
+    value = true
```
sengi, line 22: This seems surprising. Why do we want the ebs-plugin pods to tolerate **all** taints? The defaults are `CriticalAddonsOnly` and (for 5m) `NoExecute`, which seem pretty sensible.

terraform/deployments/cluster-infrastructure/aws_ebs_csi_iam.tf
```diff
@@ -0,0 +1,169 @@
...
+  role_policy_arns              = [aws_iam_policy.aws_ebs_csi_driver.arn]
+  oidc_fully_qualified_subjects = ["system:serviceaccount:kube-system:${local.csi_driver_controller_service_account_name}"]
+}
+
+resource "aws_iam_policy" "aws_ebs_csi_driver" {
+  name        = "AWSEbsCsiController-${var.cluster_name}"
+  description = "Allow the driver to manage AWS EBS"
+
+  # The argument to jsonencode() is the verbatim contents of
+  # https://raw.githubusercontent.com/kubernetes-sigs/aws-load-balancer-controller/v2.4.0/docs/install/iam_policy.json
```
sengi, line 21: Forgot to update the link? :)

sengi (commented):
Once you've tested this, please mention how/what you tested in the PR description. Code looks good, just a few minor things. Looks like you have some lint errors from `terraform fmt` as well. (I recommend configuring your editor to run `terraform fmt` automatically.)

There is a managed [EKS Addon for the EBS driver](https://docs.aws.amazon.com/eks/latest/userguide/managing-ebs-csi.html) now btw, but I agree installing it ourselves via the Helm chart is the right call for now. Maybe something to try out once we've got the big launch out of the way!

sengi:
Oh btw if you stick a link to the Trello card in the description (you only need the first two path components of the URL, like this: https://trello.com/c/Wz9dRNWA), it'll get automatically added to the card.

terraform/deployments/cluster-services/aws_ebs_csi_driver.tf
```diff
@@ -0,0 +1,36 @@
...
+
+  set {
+    name  = "controller.serviceAccount.annotations.eks\\.amazonaws\\.com/role-arn"
+    value = data.terraform_remote_state.cluster_infrastructure.outputs.aws_ebs_csi_driver_iam_role_arn
+  }
+  set {
+    name  = "enableVolumeResizing"
+    value = true
+  }
+  values = [yamlencode({
```
sengi, line 25: I'd expect this would look a lot more readable if you put it in the `values` YAML below - should be able to avoid all the nasty double escaping and whatnot. Something like:

```suggestion
  values = [yamlencode({
    enableVolumeResizing = true
    controller = {
      serviceAccount = {
        create =  true
        name   = data.terraform_remote_state.cluster_infrastructure.outputs.aws_ebs_csi_driver_controller_service_account_name
        annotations = {
      "eks.amazonaws.com/role-arn" = data.terraform_remote_state.cluster_infrastructure.outputs.aws_ebs_csi_driver_iam_role_arn
        }
      }
    }
```

sengi (approved):
Thanks - looks good, just a suggestion to clean up those Helm values that you added in the meantime.

PR #662, opened 2022-05-30
Fix signon-resources build, publish and deploy workflow

fredericfran-gds:
Update the worflow to match other apps such as [frontend](https://github.com/alphagov/frontend/blob/main/.github/workflows/deploy.yaml).

Mistake in the workflow file for some reason where the 2nd job
names a non-existing dependent job.


kentsanggds (approved):
LGTM

PR #666, opened 2022-06-01
Add backing services MySQL and Redis to CI workflow

theseanything:
This adds the ability to enable a MySQL or/and Redis backing service when running CI test using GitHub Actions. This is required by some apps such as Whitehall and Signon.


kentsanggds (approved):
LGTM

PR #669, opened 2022-06-14
Check for deploy permissions in deploy workflow

theseanything:
This adds a check in the GitHub Actions workflow for triggering a deploy in Argo Workflows. The script now checks to see if the actor (who called triggered the GitHub Action) is a member of the relevant GitHub team that allows permission to deploy (GOV.UK Production Deploy). This check will prevent those without permission for manually triggering a deploy using a workflow dispatch event.


nsabri1 (approved):
LGTM

PR #670, opened 2022-06-22
Add production terraform config

theseanything:
This adds the common tfvars for the production environment and the production backend files for each of the terraform modules.

This also fixes an issue where cluster services resources failed to come up in the correct order by adding explicit dependency information.


fredericfran-gds (approved):
thanks, lgtm.

PR #676, opened 2022-07-04
Update the Fastly DNS validation URL for Production

theseanything:
This updates the URL used by Fastly to verify domain ownership as part of setting up CDN capable of serving HTTPS traffic.


theseanything:
> How did this work before? 😅

So I think Fastly can also verify domain ownership by creating a CNAME record for www.eks.production.govuk.digital that points at "www-gov-uk.map.fastly.net". Which we automatically [do here](https://github.com/alphagov/govuk-infrastructure/blob/32687846fc2f4809f2c7a57e9541d8433a617a79/terraform/deployments/govuk-publishing-infrastructure/cdn.tf#L1).

image

So not sure if we even need DNS challenge anymore - but just updating it for correctness anyways.

PR #678, opened 2022-07-06
Fix cloud-autoscaler

fredericfran-gds:
Hypothesis is the cloud-autoscaler has been upgraded but its
IAM permissions have not and they have changed causing the
cloud-autoscaler to crash loop.

This PR:
1. fixes the IAM permissions
2. updates the cloud-autoscaler to latest version


sengi (approved):
LGTM, thanks for fixing this!

PR #679, opened 2022-07-06
fix argo workflows controller crash looping

fredericfran-gds:
Fixes include:
1. increase pod memory due to OOM (out of memory) issue
2. increase no. of workers


sengi (approved):
LGTM. I wonder how much the lack of GC on finished workflows is contributing to the RAM usage. Anyway not to worry for now - thanks for fixing this!

PR #682, opened 2022-07-07
Fix Cluster worker instance type list

fredericfran-gds:
There was a bug where an element was repeated twice in the list,
this caused constant diffs in the terraform and terraform
destroying and creating node groups on each apply.

Trello card: https://trello.com/c/dZgRACDP/956-investigate-why-terraform-always-wants-to-destroy-and-create-a-new-node-group

Co-authored-by: nadeem.sabri@digital.cabinet-office.gov.uk


sengi:
Oh did you mean to add that second commit just now?

fredericfran-gds:
Yeah, rushing too much...

PR #683, opened 2022-07-07
Add HA to Argo components by adding more replicas

fredericfran-gds:
Default replicas is 3 but set to:
1. 2 in integration
2. 1 in test

Co-authored-by: nadeem.sabri@digital.cabinet-office.gov.uk

PR #685, opened 2022-07-08
Add HA via replicas to cluster autoscaler

fredericfran-gds:
Similar to #683, we add HA by adding replicas.

Note that leader election is already active here:
https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-parameters-to-ca


nsabri1 (approved):
LGTM

PR #687, opened 2022-07-12
Add ability to lock version if manual deploy

fredericfran-gds:
This feature augments the Deploy GitHub Actions

Ref:
1. [trello card](https://trello.com/c/5uj2U5rS/900-prevent-manual-deploys-being-overridden-by-cd)

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -38,6 +38,11 @@ on:
         required: false
         default: "false"
         type: string
+      lockIfManualDeploy:
+        description: 'Whether to lock version of app if manual deployment'
+        required: true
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yaml
```diff
@@ -38,6 +38,11 @@ on:
         required: false
         default: "false"
         type: string
+      lockIfManualDeploy:
```
theseanything, line 41: Nit: As per this [comment](https://github.com/alphagov/frontend/pull/3289#discussion_r919913419) - maybe this should be called "Disable continuous deployment"

fredericfran-gds:
superseded

PR #688, opened 2022-07-12
Gem publish workflow be resilient to out of order executions

kevindew:
This is to resolve a situation where this publish workflow can produce errors when GitHub actions don't run in the order of merges. To handle this scenario (and likely some similar ones) it performs a check that the commit which is being built is the HEAD commit of the default branch of the repo.

It also does minor hygiene of this code. Adding additional information that is printed out and making variable names clearer.

There is an example run of this workflow on the govspeak gem: https://github.com/alphagov/govspeak/runs/7308887077?check_suite_focus=true

There's a bunch more info in the first commit about approaches consider.

### DIFF
--- a/.github/workflows/publish-rubygem.yaml
+++ b/.github/workflows/publish-rubygem.yaml
@@ -63,10 +63,15 @@ jobs:
         if [ "$GEM_VERSION" != "$GEM_NAME (${VERSION})" ]; then
           gem build ${GEM_NAME}.gemspec
           gem push "${GEM_NAME}-${VERSION}.gem"
+        else
+          echo "Skipping gem publish because version (${VERSION}) is already published"
         fi
+
         if ! git ls-remote --tags --exit-code origin v${VERSION}; then
           git tag v${VERSION}
           git push --tags
+        else
+          echo "Skipping git tagging because tag (v${VERSION}) already exists"
         fi
     - if: ${{ github.sha != steps.fetch_default_branch_head.outputs.sha }}
       name: "Skip publish: commit is not HEAD of default branch"
--- a/.github/workflows/publish-rubygem.yaml
+++ b/.github/workflows/publish-rubygem.yaml
@@ -49,7 +49,12 @@ jobs:
       with:
         rubygems: latest
         bundler-cache: true
-    - env:
+    - name: "Determine HEAD of default branch"
+      id: fetch_default_branch_head
+      run: echo "::set-output name=sha::$(git ls-remote origin HEAD | cut -f 1)"
+    - if: ${{ github.sha == steps.fetch_default_branch_head.outputs.sha }}
+      name: "Check if gem needs publishing"
+      env:
         GEM_NAME: ${{ inputs.gem_name }}
         GEM_HOST_API_KEY: ${{ secrets.GEM_HOST_API_KEY }}
       run: |
@@ -63,3 +68,7 @@ jobs:
           git tag v${VERSION}
           git push --tags
         fi
+    - if: ${{ github.sha != steps.fetch_default_branch_head.outputs.sha }}
+      name: "Skip publish: commit is not HEAD of default branch"
+      run: |
+        echo "Skipping publish because commit (${GITHUB_SHA}) isn't the HEAD of the default branch (${{ steps.fetch_default_branch_head.outputs.sha }})"
--- a/.github/workflows/publish-rubygem.yaml
+++ b/.github/workflows/publish-rubygem.yaml
@@ -58,20 +58,20 @@ jobs:
         GEM_NAME: ${{ inputs.gem_name }}
         GEM_HOST_API_KEY: ${{ secrets.GEM_HOST_API_KEY }}
       run: |
-        VERSION=$(ruby -e "puts eval(File.read('$GEM_NAME.gemspec')).version")
-        GEM_VERSION=$(gem list --exact --remote $GEM_NAME)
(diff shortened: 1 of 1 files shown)


sengi:
Thanks, Kevin! The context in the commit messages was super helpful ❤️

+ @theseanything since he's leading our CI/CD work.

sengi (approved):

theseanything:
@kevindew thanks for raising this - Just to check my understanding - out of order runs causes an error when the release job tried to [push a gem version that already has been published](https://github.com/alphagov/govspeak/runs/6982747734?check_suite_focus=true). This happens because a release job triggered by a later commit can publish a newer version of the Gem before a release job for an earlier commit. This causes the earlier release job attempt to re-publish the Gem because the latest published version is now different (i.e. new version), however this command will fail (`Repushing of gem versions is not allowed.`) as that version of the Gem has already been published when the Gem version was originally bumped. 

[This check](https://github.com/alphagov/govuk-infrastructure/blob/c34054dd9ca45aff71fa19d5fb75f4fed345a4a7/.github/workflows/publish-rubygem.yaml#L58) assumes that local version of the gem will always be equal or greater than the latest published version. This assumption breaks when due to the ability for release jobs to run out of order.

Is there any other issues that are caused by out of order running of the release job? My only concern is this particular solution could prevent older version of Gems being pushed i.e. if you do two version bumps consecutively, then older version could possibly never get published. Might be more explicit and simpler to fix the check around trying to publish the Gem - e.g. [don't fail or error when tried to re-push a Gem](https://github.com/fac/ruby-gem-push-action/blob/1a5013c6be8e98db0ffdb0c1f95d23ad1b9016ba/gem-push-action.sh#L78).

kevindew:
> @kevindew thanks for raising this

No problem, thanks for your review and thinking on this 👍 

> Just to check my understanding - out of order runs causes an error when the release job tried to [push a gem version that already has been published](https://github.com/alphagov/govspeak/runs/6982747734?check_suite_focus=true). This happens because a release job triggered by a later commit can publish a newer version of the Gem before a release job for an earlier commit. This causes the earlier release job attempt to re-publish the Gem because the latest published version is now different (i.e. new version), however this command will fail (`Repushing of gem versions is not allowed.`) as that version of the Gem has already been published when the Gem version was originally bumped.

Yup, that's correct

> [This check](https://github.com/alphagov/govuk-infrastructure/blob/c34054dd9ca45aff71fa19d5fb75f4fed345a4a7/.github/workflows/publish-rubygem.yaml#L58) assumes that local version of the gem will always be equal or greater than the latest published version. This assumption breaks when due to the ability for release jobs to run out of order.

Yup also correct. 

As an aside, in an earlier version of this I worked out how we could do a more elaborate version check https://github.com/alphagov/govuk-infrastructure/commit/08f40051469737e915ba932295cfdf48747d46d3 but decided that checking the HEAD was a better approach.

> Is there any other issues that are caused by out of order running of the release job? My only concern is this particular solution could prevent older version of Gems being pushed i.e. if you do two version bumps consecutively, then older version could possibly never get published. Might be more explicit and simpler to fix the check around trying to publish the Gem - e.g. [don't fail or error when trying to re-push a Gem](https://github.com/fac/ruby-gem-push-action/blob/1a5013c6be8e98db0ffdb0c1f95d23ad1b9016ba/gem-push-action.sh#L78).

Sure, I see what you mean it's an issue I thought about. Essentially if you're merging multiple releases at the same time or merging PRs straight after a release you're in a danger zone - well with our implicit release system. Given we don't have guarantees of the ordering of when an action will run the only thing that seems that you can rely on is what seems to actually be the HEAD - and there's essentially always a risk if you merge anything after a release is intended but the release is not confirmed.

theseanything:
> As an aside, in an earlier version of this I worked out how we could do a more elaborate version check [08f4005](https://github.com/alphagov/govuk-infrastructure/commit/08f40051469737e915ba932295cfdf48747d46d3) but decided that checking the HEAD was a better approach.

> Sure, I see what you mean it's an issue I thought about. Essentially if you're merging multiple releases at the same time or merging PRs straight after a release you're in a danger zone - well with our implicit release system. Given we don't have guarantees of the ordering of when an action will run the only thing that seems that you can rely on is what seems to actually be the HEAD - and there's essentially always a risk if you merge anything after a release is intended but the release is not confirmed.

Ideally we'd like to guarantee that every version of Gem released in the code (i.e. version bump) is published to Ruby Gems. It's not a problem if an older version of a Gem is pushed after newer version. Ultimately we just don't consider jobs "failed" if we try re-pushing an existing gem version. I think we're trying to address that in a round about way by adding a constraint that we only publish Gems that are a newer version that the latest release (either by checking HEAD or the actual Gem version). That constraint means we breaks the guarantee, because it depends on the assumption that things run in order. Here we're trying to fix that assumption, rather than the actual issue directly.

I think the logic is much easier to understand if we say "the job passed because it tired to re-push a gem and it doesn't matter". Rather than "the job passed because we didn't bother running the release step, because the commit isn't latest commit on main, because we don't want to try and push a Gem that might be an older version, because we can't try and re-push a gem".

kevindew:
@theseanything thanks for the face-to-face chat on this the other day.

As we discussed some of the information that may have been missing from this discussion was that this brings predictability into a situation that otherwise doesn't have it (merging multiple things with a gem release is dangerous unless you wait until you are sure a gem has been released) and I've amended the commit message to capture some of those concerns.

PR #692, opened 2022-07-15
Create resuable workflow for setting automatic_deploy_enabled

fredericfran-gds:


nsabri1 (approved):
OK

.github/workflows/set-automatic-deploys-enabled.yaml
```diff
@@ -0,0 +1,115 @@
...
+  workflow_call:
+    inputs:
+      resetImageTag:
+        description: 'Whether to reset image tag to main'
+        required: false
+        default: false
+        type: boolean
+      automaticDeploysEnabled:
+        description: 'Set whether automatic deploys is active or not'
+        required: false
```
sengi, line 54: Would it make sense for this to be required? (given that the purpose of the job is to set it one way or the other? Or are they other usability implications of it being a required input?)

Similar question for `environment` below also.

fredericfran-gds: I guess if i put required true, we don't need the default?

sengi: Anyway sorry didn't mean to split hairs - I should have prefixed this with "optional"! By all means try it out either way and as long as the UI isn't super awkward it's all good with me :)

fredericfran-gds: No worries, I was thinking about it before but could not see the required info. :-)

PR #698, opened 2022-07-19
fix if statement in workflow

fredericfran-gds:


.github/workflows/set-automatic-deploys-enabled.yaml
```diff
@@ -86,7 +86,7 @@ jobs:
           RESET_IMAGE_TAG: ${{ inputs.resetImageTag }}
         run: |
           IMAGE_TAG=""
-          if [ "${RESET_IMAGE_TAG}" ]; then
+          if [ "${RESET_IMAGE_TAG}" = "true" ]; then
```
sengi, line 89: nit: for consistency with the rest of the file, might as well use `[[` (i.e. the bash builtin, since we're already using bash-only stuff elsewhere in this script)

```suggestion
          if [ "${RESET_IMAGE_TAG}" = "true" ]; then
```

fredericfran-gds: cool, thanks

PR #699, opened 2022-07-19
added force destroy to allow tf to destroy buckets with versioning

desenhar:


theseanything (approved):

terraform/deployments/govuk-publishing-infrastructure/app_assets_s3.tf
```diff
@@ -1,5 +1,6 @@
 resource "aws_s3_bucket" "app_assets" {
-  bucket = "govuk-app-assets-${var.govuk_environment}"
+  bucket        = "govuk-app-assets-${var.govuk_environment}"
+  force_destroy = true
```
sengi, line 3: Just for future reference, we already have a `var.force_destroy` for exactly this use case (false in production and true everywhere else).

PR #700, opened 2022-07-20
Use per-env force_destroy setting in govuk-infra and enable it in integration.

sengi:
See commit messages for more details.

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/app_assets_s3.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/app_assets_s3.tf
@@ -1,6 +1,6 @@
 resource "aws_s3_bucket" "app_assets" {
   bucket        = "govuk-app-assets-${var.govuk_environment}"
-  force_destroy = true
+  force_destroy = var.force_destroy
   tags = {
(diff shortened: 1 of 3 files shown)


desenhar (approved):
lgtm

theseanything:
Nit: Wondering if we need this safeguard even for production? Assumed that anything in the assets bucket would be ephemeral. But understand if we just want to be careful.

sengi:
> Nit: Wondering if we need this safeguard even for production? Assumed that anything in the assets bucket would be ephemeral. But understand if we just want to be careful.

Great question. I think it's worthwhile leaving all the delete protection stuff in place in production because it guards against accidental / "sorceror's apprentice" scenarios where Terraform deletes and re-creates something and it's not immediately obvious from the diff that it'd involve data loss.

As a (slightly contrived, but probably still representative) example: say we were to rename the assets bucket; TF isn't smart enough to do that. It'd delete and re-create it and that would definitely not be something we'd want to let it do in prod — we'd want to create the new bucket, sync the contents, `tf state mv` and then delete the old bucket.

theseanything (approved):

PR #702, opened 2022-07-21
Reduce logging for filebeat to resolve OOM errors.

marcpomfret:
Added file rotation to see if this stops the OOM restarts on the filebeat pods. 
We've manually applied this to staging for now, will check this in week and if its sorted will apply.

### DIFF
--- a/terraform/deployments/cluster-services/logging.tf
+++ b/terraform/deployments/cluster-services/logging.tf
@@ -28,14 +28,16 @@ resource "helm_release" "filebeat" {
             }]
           }
         ]
-
         "output.logstash" : {
           "hosts" : ["$${LOGSTASH_HOST:? LOGSTASH_HOST variable is not set}:$${LOGSTASH_PORT:? LOGSTASH_PORT variable is not set}"]
           "loadbalance" : true
           "ssl.enabled" : true
         }
-        "logging.files" : {
-          "keepfiles" : 7
+        "logging.metrics.enabled" : false
+        "http.enabled" : false
+        "logging.level" : "warn"
+        "output.file" : {
+          "enabled" : false
         }
       })
     }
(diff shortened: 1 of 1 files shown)


fredericfran-gds (approved):
ok

sengi (commented):
Isn't this just the log output from filebeat itself? https://www.elastic.co/guide/en/beats/filebeat/current/configuration-logging.html

sengi:
Somewhat related question: are filebeat's own logs appearing in LogIt? If not, we might need to set `logging.to_stderr` if it's not already defaulting to true. We probably don't want it to keep logs in the local filesystem for very long, since we'd only ever want those for debugging filebeat itself in the event that it's failing to send logs to logstash.

terraform/deployments/cluster-services/logging.tf
```diff
@@ -34,6 +34,21 @@ resource "helm_release" "filebeat" {
...
+        }
+        "logging.metrics" : {
+          "enabled" : false
+        }
+        "http" : {
+          "enabled" : false
+        }
+        "logging" : {
+          "level" : "warn"
+        }
```
sengi, line 48: If I'm reading the [docs](https://www.elastic.co/guide/en/beats/filebeat/current/configuration-logging.html#configuration-logging) right, this should be a single key called `logging.level` rather than `level` nested inside `logging`.
```suggestion
        "logging.level" : "warn"
```

Similarly for most (but not all) of the others, e.g. `logging.metrics.enabled` should be all one key, but `output.file { enabled }` is already good.

It's super confusing that they use dots inside some of the keys. Where there really is nesting, it seems to be shown as nested in the docs (see logging.files for example) but it's not very clear 🙃

terraform/deployments/cluster-services/logging.tf
```diff
@@ -34,6 +34,21 @@ resource "helm_release" "filebeat" {
           "loadbalance" : true
           "ssl.enabled" : true
         }
+        "logging.files" : {
+          "keepfiles" : 7
+        }
```
sengi, line 39: I think we want it not to log to the filesystem at all, no? Just stderr.
```suggestion
        "logging.to_files" : false
        "logging.to_stderr" true
```

sengi (approved):
👍

PR #703, opened 2022-07-22
fix metrics exporter image

desenhar:


sengi (approved):
LGTM but looks like just style fixes - is that what you intended? (asking because the commit message / PR title sounds like something more, so checking just on the off change you forgot to add something!)

desenhar:
its meant to be just a very small style fix, need to work on my commit messages 😓 .

sengi:
No worries!

PR #711, opened 2022-08-01
Set ArgoCD application controller replicas to 1

fredericfran-gds:
I initially thought it would give HA by setting replicas > 1 but
it does not so reverting.

See also: #710


sengi (approved):
Nice one! 👍

PR #712, opened 2022-08-08
Add permissions check for deployments

theseanything:
This adds a permissions check in the deployment re-usable GitHub Action workflow. The check prevents user who are not members of the GOV.UK Production Deploy team to deploy. This is a simply check to prevent those user from accidentally deploy.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -86,7 +86,26 @@ jobs:
     name: Update image tag
     runs-on: ubuntu-latest
     steps:
-      - name: Send webhook
+      - name: Check deploy permissions
+        id: deploy-permissions
+        env:
+          ENVIRONMENT: ${{ inputs.environment || 'integration' }}
+          GH_TOKEN: ${{ secrets.GOVUK_CI_GITHUB_API_TOKEN }}
+          GITHUB_TEAM: gov-uk-production-deploy
+          GITHUB_USER: ${{ github.triggering_actor }}
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yaml
```diff
@@ -86,26 +86,36 @@ jobs:
...
+          ENVIRONMENT: ${{ inputs.environment || 'integration' }}
+          GH_TOKEN: ${{ secrets.GOVUK_CI_GITHUB_API_TOKEN }}
+          GITHUB_TEAM: gov-uk-production-deploy
+          GITHUB_USER: ${{ github.triggering_actor }}
+        run: |
+          # Check actor has permission to deploy
+          TEAM_MEMBERSHIP=$(gh api orgs/alphagov/teams/${GITHUB_TEAM}/memberships/${GITHUB_USER} -q .state || echo "false")
+
+          if [[ "${TEAM_MEMBERSHIP}" = "active" || "${ENVIRONMENT}" = 'integration' ]]; then
+            echo "::set-output name=authorised::true"
```
fredericfran-gds, line 101: I wonder if we could not simplify this by using:
if: steps.deploy-permissions.outcome == 'success'  on line 108

see docs: https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context

We could remove the output setting.

fredericfran-gds (commented):
See inline, thanks.

fredericfran-gds (approved):
looks good thanks

PR #713, opened 2022-08-08
Remove workflowTrigger parameter from deploy workflow

theseanything:
This parameter is unused and can be removed.


fredericfran-gds (approved):
thanks for finding this out, leftover from some experimentation.

PR #714, opened 2022-08-08
Change Rails CI workflow to not test JavaScript by default

theseanything:
This changes the default value for requiresJavascript parameter in the test-rails workflow. This is to make the parameter usage consistent and predictable.


fredericfran-gds (approved):
ok

PR #718, opened 2022-08-09
Lifecycle rules for ecr to clean up images

desenhar:
https://trello.com/c/MK34KlBn/980-lifecycle-rules-for-ecr-to-clean-up-images

One rule retains 20 tagged images and cleans up any untagged images after 30 days.

### DIFF
--- a/terraform/deployments/ecr/main.tf
+++ b/terraform/deployments/ecr/main.tf
@@ -209,6 +209,55 @@ resource "aws_iam_user_policy" "github_ecr_user_policy" {
   })
 }
 
+resource "aws_ecr_lifecycle_policy" "untaggedImage_expiry_policy" {
+  for_each   = toset([for repo in local.repositories : aws_ecr_repository.repositories[repo].name])
+  repository = each.key
+
+  policy = <<EOF
+{
+    "rules": [
+        {
+            "rulePriority": 1,
+            "description": "Expire images older than 30 days",
+            "selection": {
+                "tagStatus": "untagged",
+                "countType": "sinceImagePushed",
+                "countUnit": "days",
+                "countNumber": 30
+            },
+            "action": {
+                "type": "expire"
+            }
+        }
+    ]
+}
+EOF
+}
+
+resource "aws_ecr_lifecycle_policy" "keep_image_policy" {
+  for_each   = toset([for repo in local.repositories : aws_ecr_repository.repositories[repo].name])
+  repository = each.key
+
+  policy = <<EOF
+{
+    "rules": [
+        {
+            "rulePriority": 1,
+            "description": "Keep last 20 images",
+            "selection": {
+                "tagStatus": "tagged",
+                "tagPrefixList": ["v"],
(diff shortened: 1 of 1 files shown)


terraform/deployments/ecr/main.tf
```diff
@@ -209,6 +209,42 @@ resource "aws_iam_user_policy" "github_ecr_user_policy" {
...
+                "tagStatus": "untagged",
+                "countType": "sinceImagePushed",
+                "countUnit": "days",
+                "countNumber": 30
+            },
+            "action": {
+                "type": "expire"
+            },
+            "rulePriority": 1,
+            "description": "Keep last 20 images",
```
theseanything, line 232: We might need to update the deploy script to not re-build an image if it already exists. Currently, if someone were to manually deploy (i.e. which builds and pushes a new image) 20 times, we could potentially lose the version we have deployed in production.

desenhar: thats a good point, would it be easier to have another rule with higher priority, that always leave one image tagged 'latest' ? 
something like:
`{
            "rulePriority": 1,
            "description": "Rule 1",
            "selection": {
                "tagStatus": "tagged",
                "tagPrefixList": ["latest"],
                "countType": "imageCountMoreThan",
                "countNumber": 1
            },
            "action": {
                "type": "expire"
            }`

theseanything: Hmm, I don't think we can use "latest" tho - as that's just assigned to whatever image was publish last. Maybe we need to switch to better tagging scheme e.g. release_<number> that is more representative version order?

theseanything: tbh, I think this can be a further improvement and shouldn't stop these initially lifecycle rule being enabled.

theseanything (approved):

sengi:
Not to rain on anyone's parade, but as currently implemented this seems to be a timebomb. As at least a quick fix, I'd recommend tagging `deployed_to_production` etc. and excluding those three tags from GC.

sengi:
Or maybe I've misunderstood the semantics of these rules, in which case please excuse the intrusion 😅 

As long as this can't remove images that are in use, I'm happy. (Bonus points if it can't remove the last or or two images that were previously in use as well, i.e. ones that someone might conceivably want to roll back to in an emergency.)

terraform/deployments/ecr/main.tf
```diff
@@ -209,6 +209,42 @@ resource "aws_iam_user_policy" "github_ecr_user_policy" {
   })
 }
 
+resource "aws_ecr_lifecycle_policy" "ecr_lifecycle_policy" {
+  for_each   = toset([for repo in local.repositories : aws_ecr_repository.repositories[repo].name])
+  repository = each.key
+
+  policy = <<EOF
```
sengi, line 216: Tip for next time: use `jsonencode()` instead of string templates, so that the parser can check the syntax ahead of time.

PR #719, opened 2022-08-09
Fix GitHub deployment

fredericfran-gds:
fixes:
1. add to the README.md that the GitHub token creator must be an
   `alphagov` org admin
2. fixes to #704, create only list of repos which has access to secret,
   rather than the secret themselves. This is what prior art is doing.


kentsanggds (approved):
LGTM

PR #720, opened 2022-08-12
Support installing extra system dependencies for CI

theseanything:
This adds the ability to install additional system dependencies for the CI GitHub Action workflow. This is needed by Whitehall as the tests have a dependencies on Ghostscript.

### DIFF
--- a/.github/workflows/test-rails.yaml
+++ b/.github/workflows/test-rails.yaml
@@ -60,6 +60,11 @@ on:
         description: 'Password for the MySQL database'
         required: false
         type: string
+      extraSystemDependencies:
+        description: 'Install additional system dependencies'
+        required: false
+        default: ''
+        type: string
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


.github/workflows/test-rails.yaml
```diff
@@ -60,6 +60,11 @@ on:
         description: 'Password for the MySQL database'
         required: false
         type: string
+      extraSystemDependencies:
+        description: 'Install additional system dependenices'
```
sengi, line 64: typo (similarly below on L87)
```suggestion
        description: 'Install additional system dependencies'
```

.github/workflows/test-rails.yaml
```diff
@@ -79,6 +84,12 @@ jobs:
            ref: deployed-to-production
            path: vendor/govuk-content-schemas
 
+       - name: Install additional system dependenices
+         if: ${{ inputs.extraSystemDependencies }}
+         run: |
+           sudo apt-get update
+           sudo apt-get install -y ${{ inputs.extraSystemDependencies }}
```
sengi, line 91: `--no-install-recommends` is good practice here to avoid installing unnecessary packages.
```suggestion
           sudo apt-get install -y --no-install-recommends ${{ inputs.extraSystemDependencies }}
```

sengi (approved):
It might be cleaner to include these packages in our builder image so as to avoid having to install them on every test run, but this is ok for now.

PR #723, opened 2022-08-12
Build basic Clamav image

fredericfran-gds:
Can be used for scanning assets in Asset-manager for example.


fredericfran-gds:
We will use the upstream clamav one: https://hub.docker.com/r/clamav/clamav

PR #725, opened 2022-08-15
Add clamev efs

fredericfran-gds:
We want to create a EFS volume to store the Clamav virus
database. Before, we attempted to create the EFS in AWS only
and mount it as NFS in the Clamav pod but this was unsuccessful
since the EFS has root ownership and the pod is running as non-root.

We could create the access point of the EFS and sets the same ownership
as the pod in AWS but we would have to use the volume handler option of
k8s PersistentVolume for k8s to mount the access point properly. This
option requires the AWS EFS CSI driver as a prerequisite.

The chosen solution is:
1. install AWS EFS CSI driver and associated permissions
2. The storage class requires a EFS id so we create the EFS and
   pass it to the driver.

### DIFF
--- a/terraform/deployments/cluster-infrastructure/aws_efs_csi_iam.tf
+++ b/terraform/deployments/cluster-infrastructure/aws_efs_csi_iam.tf
@@ -1,0 +1,66 @@
+locals {
+  efs_csi_driver_controller_service_account_name = "efs-csi-controller-sa"
+}
+
+module "aws_efs_csi_driver_iam_role" {
+  source                        = "terraform-aws-modules/iam/aws//modules/iam-assumable-role-with-oidc"
+  version                       = "~> 4.0"
+  create_role                   = true
+  role_name                     = "${local.efs_csi_driver_controller_service_account_name}-${var.cluster_name}"
+  role_description              = "Role for the AWS EFS CSI driver controller. Corresponds to ${local.efs_csi_driver_controller_service_account_name} k8s ServiceAccount."
+  provider_url                  = module.eks.oidc_provider
+  role_policy_arns              = [aws_iam_policy.aws_efs_csi_driver.arn]
+  oidc_fully_qualified_subjects = ["system:serviceaccount:kube-system:${local.efs_csi_driver_controller_service_account_name}"]
+}
+
+resource "aws_iam_role_policy_attachment" "eks_nodes_efs" {
+  role       = module.eks.eks_managed_node_groups["main"].iam_role_name
+  policy_arn = aws_iam_policy.aws_efs_csi_driver.arn
+}
+
+resource "aws_iam_policy" "aws_efs_csi_driver" {
+  name        = "AWSEfsCsiController-${var.cluster_name}"
+  description = "Allow the driver to manage AWS EFS"
+
+  # The argument to jsonencode() is the verbatim contents of
+  # https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/docs/iam-policy-example.json
+  # (except for whitespace changes from terraform fmt).
+  policy = jsonencode({
+
+    "Version" : "2012-10-17",
+    "Statement" : [
+      {
+        "Effect" : "Allow",
+        "Action" : [
+          "elasticfilesystem:DescribeAccessPoints",
+          "elasticfilesystem:DescribeFileSystems",
+          "elasticfilesystem:DescribeMountTargets",
+          "ec2:DescribeAvailabilityZones"
+        ],
+        "Resource" : "*"
+      },
+      {
+        "Effect" : "Allow",
+        "Action" : [
+          "elasticfilesystem:CreateAccessPoint"
+        ],
+        "Resource" : "*",
+        "Condition" : {
+          "StringLike" : {
+            "aws:RequestTag/efs.csi.aws.com/cluster" : "true"
+          }
+        }
+      },
+      {
+        "Effect" : "Allow",
+        "Action" : "elasticfilesystem:DeleteAccessPoint",
+        "Resource" : "*",
+        "Condition" : {
+          "StringEquals" : {
+            "aws:ResourceTag/efs.csi.aws.com/cluster" : "true"
+          }
+        }
+      }
+    ]
+  })
+}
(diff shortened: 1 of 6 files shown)


terraform/deployments/govuk-publishing-infrastructure/clamav_efs.tf
```diff
@@ -0,0 +1,28 @@
+locals {
+  clamav_efs_name = "clamav-efs-${local.cluster_name}"
+}
+
+resource "aws_efs_file_system" "clamav-efs" {
```
sengi, line 5: Consider putting a `Name` and/or `Description` tag on this to make it clear that it's for the ClamAV *signatures* (database). Or maybe just work it into the name somehow? (Could probably lose the `efs` from the name in most contexts, but signposting that it's for the sigs as opposed to, say, the files being scanned or anything like that, is quite helpful.)

terraform/deployments/govuk-publishing-infrastructure/security.tf
```diff
@@ -207,3 +207,23 @@ resource "aws_security_group_rule" "licensify_frontend_from_eks_workers" {
   security_group_id        = data.terraform_remote_state.infra_security_groups.outputs.sg_licensify-frontend_internal_lb_id
   source_security_group_id = data.terraform_remote_state.cluster_infrastructure.outputs.node_security_group_id
 }
+
+resource "aws_security_group_rule" "clamav_efs_to_any_any" {
+  description       = "Clamav sends requests to anywhere over any protocol"
```
sengi, line 212:
```suggestion
  description       = "Clam DB EFS sends requests to anywhere over any protocol"
```

terraform/deployments/govuk-publishing-infrastructure/security.tf
```diff
@@ -207,3 +207,23 @@ resource "aws_security_group_rule" "licensify_frontend_from_eks_workers" {
...
+  type              = "egress"
+  from_port         = 0
+  to_port           = 0
+  protocol          = -1
+  cidr_blocks       = ["0.0.0.0/0"]
+  security_group_id = aws_security_group.clamav-efs.id
+}
```
sengi, line 219: I strongly suspect we don't need this rule at all - doesn't the client always initiate the TCP connection to the NFS server?

sengi (approved):
Cool - took me a while to realise why we need a shared filer here but I think it's the right compromise - thanks! 👍

terraform/deployments/govuk-publishing-infrastructure/clamav_efs.tf
sengi: `clamav-sigs-db`? `clamav-db`? 🤷

fredericfran-gds: I will, thanks

terraform/deployments/govuk-publishing-infrastructure/security.tf
fredericfran-gds: thanks

terraform/deployments/govuk-publishing-infrastructure/security.tf
fredericfran-gds: Yeah, it is stateful, could remove that. Thanks.

terraform/deployments/cluster-infrastructure/aws_efs_csi_iam.tf
```diff
@@ -0,0 +1,66 @@
...
+  version                       = "~> 4.0"
+  create_role                   = true
+  role_name                     = "${local.efs_csi_driver_controller_service_account_name}-${var.cluster_name}"
+  role_description              = "Role for the AWS EFS CSI driver controller. Corresponds to ${local.efs_csi_driver_controller_service_account_name} k8s ServiceAccount."
+  provider_url                  = module.eks.oidc_provider
+  role_policy_arns              = [aws_iam_policy.aws_efs_csi_driver.arn]
+  oidc_fully_qualified_subjects = ["system:serviceaccount:kube-system:${local.efs_csi_driver_controller_service_account_name}"]
+}
+
+resource "aws_iam_role_policy_attachment" "eks_nodes_efs" {
```
sengi, line 16: Looks like you have a service account for the controller, but you're attaching the policy to the nodes. You want to attach the policy to the IAM role that corresponds to the k8s serviceaccount, right?

If you attach the policy to the nodes, you're granting all pods the ability to manage EFS — which is presumably what you're trying to avoid by creating the serviceaccount and corresponding IRSA IAM role etc.?

terraform/deployments/cluster-infrastructure/clamav_db_efs.tf
```diff
@@ -0,0 +1,31 @@
...
+  security_group_id        = aws_security_group.clamav-db.id
+  source_security_group_id = local.node_security_group_id
+}
+
+resource "aws_efs_mount_target" "clamav-db-mount-targets" {
+  for_each        = toset(data.terraform_remote_state.infra_networking.outputs.private_subnet_ids)
+  file_system_id  = aws_efs_file_system.clamav-db.id
+  subnet_id       = each.key
+  security_groups = [aws_security_group.clamav-db.id]
+}
```
sengi, line 31: Do you still need this bit now that you have the CSI provisioner? (I genuinely have no idea!)

terraform/deployments/cluster-infrastructure/aws_efs_csi_iam.tf
fredericfran-gds: I got the info from here, the same specs: https://github.com/kubernetes-sigs/aws-efs-csi-driver#installation

terraform/deployments/cluster-infrastructure/clamav_db_efs.tf
fredericfran-gds: yep, it didn't work without. I added it after.

terraform/deployments/cluster-infrastructure/aws_efs_csi_iam.tf
sengi: Right so where it talks about "several methods to grant IAM permission", you want the first one (IAM Role for Service Account). It looks like you're kinda doing both here at the moment though.

PR #730, opened 2022-08-24
Allow cross-communication between EKS and EC2 www-origins

fredericfran-gds:
For traffic splitting experiment, communication between the
EKS ad EC2 www-origins should be allowed since only one origin
sometimes have the required rails assets so the other origin has
to go and fetch it from there if it gets such a request.

Breaking changes:
1. Before EKS integration www-origin did not have an IP blocklist,
   it now does. It is now similar to the EC2 one. Did not want to
   complicate the code by making an exception.

Caveats:
1. Integration origins is still password protected, maybe we should
   remove this as the IP block list is sufficient. Otherwise, this
   will complicate the code for cross-communication for rails assets.

### DIFF
--- a/terraform/deployments/cluster-infrastructure/outputs.tf
+++ b/terraform/deployments/cluster-infrastructure/outputs.tf
@@ -121,3 +121,7 @@ output "monitoring_namespace" {
 output "clamav_db_efs_id" {
   value = aws_efs_file_system.clamav-db.id
 }
+
+output "public_nat_gateway_ips" {
+  value = [for eip in aws_eip.eks_nat : eip.public_ip]
+}
--- a/terraform/deployments/govuk-publishing-infrastructure/main.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/main.tf
@@ -11,6 +11,10 @@ terraform {
       source  = "hashicorp/random"
       version = "~> 3.1"
     }
+    fastly = {
+      source  = "fastly/fastly"
+      version = "~> 2.1"
+    }
   }
 }
 
@@ -34,3 +38,9 @@ provider "aws" {
 }
 
 provider "random" {}
+
+# used by the fastly ip ranges provider.
+# an API key is needed but 'fake' seems to work.
+provider "fastly" {
+  api_key = "fake"
+}
--- a/terraform/deployments/govuk-publishing-infrastructure/security.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/security.tf
@@ -207,3 +207,75 @@ resource "aws_security_group_rule" "licensify_frontend_from_eks_workers" {
   security_group_id        = data.terraform_remote_state.infra_security_groups.outputs.sg_licensify-frontend_internal_lb_id
   source_security_group_id = data.terraform_remote_state.cluster_infrastructure.outputs.node_security_group_id
 }
+
+# TODO: Remove after EC2 GOV.UK decommissioned
+resource "aws_security_group_rule" "ec2_www_origin_from_eks_workers" {
+  description       = "EC2 www-origin accepts requests from EKS NAT gateways"
+  type              = "ingress"
+  from_port         = 443
+  to_port           = 443
+  protocol          = "tcp"
+  security_group_id = data.terraform_remote_state.infra_security_groups.outputs.sg_cache_external_elb_id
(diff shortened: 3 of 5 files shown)


terraform/deployments/cluster-infrastructure/outputs.tf
```diff
@@ -121,3 +121,7 @@ output "monitoring_namespace" {
 output "clamav_db_efs_id" {
   value = aws_efs_file_system.clamav-db.id
 }
+
+output "public_ips_nat_gateways" {
```
sengi, line 125: optional: `public_nat_gateway_ips` might be slightly clearer name
```suggestion
output "public_nat_gateway_ips" {
```

terraform/deployments/govuk-publishing-infrastructure/main.tf
```diff
@@ -34,3 +38,9 @@ provider "aws" {
 }
 
 provider "random" {}
+
+# used by the fastly ip ranges provider.
+# an API key is needed but 'fake' seems to work.
```
sengi, line 43: cool 😅 (I guess there's a bit of a bug in the provider that it requires it for some calls where Fastly doesn't 🤷)

terraform/deployments/govuk-publishing-infrastructure/security.tf
```diff
@@ -207,3 +207,73 @@ resource "aws_security_group_rule" "licensify_frontend_from_eks_workers" {
...
 }
+
+resource "aws_security_group_rule" "ec2_www_origin_from_eks_workers" {
+  description       = "EC2 www-origin accepts requests from EKS NAT gateways"
+  type              = "ingress"
+  from_port         = 443
+  to_port           = 443
+  protocol          = "tcp"
+  security_group_id = data.terraform_remote_state.infra_security_groups.outputs.sg_cache_external_elb_id
+  cidr_blocks       = formatlist("%s/32", data.terraform_remote_state.cluster_infrastructure.outputs.public_ips_nat_gateways)
```
sengi, line 218: Nice! I hadn't come across `formatlist` before. TIL :D

terraform/deployments/variables/common.tfvars
```diff
@@ -1 +1,13 @@
 dex_github_orgs_teams = [{ name = "alphagov", teams = ["gov-uk", "gov-uk-production-deploy"] }]
+
+gds_office_cidrs = ["213.86.153.211/32",
```
sengi, line 3: optional: is there an existing place we can get this from conveniently (e.g. infra-networking remote state maybe??)

terraform/deployments/govuk-publishing-infrastructure/security.tf
```diff
@@ -207,3 +207,73 @@ resource "aws_security_group_rule" "licensify_frontend_from_eks_workers" {
...
   source_security_group_id = data.terraform_remote_state.cluster_infrastructure.outputs.node_security_group_id
 }
+
+resource "aws_security_group_rule" "ec2_www_origin_from_eks_workers" {
+  description       = "EC2 www-origin accepts requests from EKS NAT gateways"
+  type              = "ingress"
+  from_port         = 443
+  to_port           = 443
+  protocol          = "tcp"
+  security_group_id = data.terraform_remote_state.infra_security_groups.outputs.sg_cache_external_elb_id
```
sengi, line 217: I have a sneaky suspicion that `sg_cache_external_elb_id` might be the confusing disused/left-over one from the early days of the AWS migration. Definitely check this out if you haven't already. My money's on that it turns out we want `cache_public_lb_id` instead :)

sengi (approved):
Cool - just one important comment which is the bit about double-checking we've got the right load balancer (there's a confusing "decoy" one left over from years ago that we never got around to cleaning up 😅)

terraform/deployments/govuk-publishing-infrastructure/security.tf
sengi: Ohh hang on, I think I see what I might have missed here. I bet it's that the security group still bears the name of the old "external" LB but refers to the new "public" one.

If you've tested it and it works, I must be missing something 😅

sengi:
Oh I thought of another thing: I don't think we'll be able to use an IP allowlist for integration because other gov depts need to access it for training content people to use the publishing tools :(

It's fine in the short term though, i.e. before Replatforming MVP launch. It only becomes an issue once we fully switch over.

Also thanks for the lovely clear context in the PR desc — I wouldn't have thought of this issue without it! 🙌

PR #733, opened 2022-09-01
Enable the AWS LB controller pod readiness gate in `ns/apps`.

sengi:
See https://kubernetes-sigs.github.io/aws-load-balancer-controller/latest/deploy/pod_readiness_gate/

This should:
* eliminate the small availability blips that we've seen on rolling update in a few cases
* shake out any remaining misconfigured healthchecks that might be lurking

I had no idea that Amazon doesn't enable this by default, though you could say by now I ought to know better 😅

### DIFF
--- a/terraform/deployments/cluster-services/argo.tf
+++ b/terraform/deployments/cluster-services/argo.tf
@@ -14,8 +14,12 @@ locals {
 
 resource "kubernetes_namespace" "apps" {
   metadata {
-    name   = var.apps_namespace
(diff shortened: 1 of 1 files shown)


sengi:
force-push to fix a `terraform lint` nit 🙄

nsabri1 (approved):
OK

sengi:
Applied the TF change in all three clusters and restarted everything in order to pick up the readiness gate (`k rollout restart deploy; k rollout status deploy -w`).

PR #739, opened 2022-09-08
update lifecycle rule for new deployed-to-production tag

desenhar:
https://trello.com/c/MK34KlBn/980-lifecycle-rules-for-ecr-to-clean-up-images

This is a lifecycle rule that applies to all repositories in production ECR,
Rule priorities are set in this way so any in-use images, which are tagged with 'deployed-to-{{env}}' are retained.
Untagged images are expired and so are images older than 30 days.

### DIFF
--- a/terraform/deployments/ecr/main.tf
+++ b/terraform/deployments/ecr/main.tf
@@ -214,34 +214,51 @@ resource "aws_ecr_lifecycle_policy" "ecr_lifecycle_policy" {
     "rules" : [
       {
         "rulePriority" : 1,
-        "description" : "Rule 1",
+        "description" : "Keep last 100 images with tag prefix deployed-to",
         "selection" : {
           "tagStatus" : "tagged",
           "tagPrefixList" : ["deployed-to"],
           "countType" : "imageCountMoreThan",
-          "countNumber" : 1
+          "countNumber" : 100
         },
         "action" : {
           "type" : "expire"
-        },
-        "rulePriority" : 2,
-        "description" : "Expire images older than 30 days",
+        }
+      },
+      {
+         "rulePriority": 2,
+         "description": "Expire untagged images older than 1 day",
+         "selection": {
+             "tagStatus": "untagged",
+             "countType": "sinceImagePushed",
+             "countUnit": "days",
+             "countNumber": 1
+         },
+         "action": {
+             "type": "expire"
+         }
+      },
+      {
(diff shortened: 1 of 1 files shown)


terraform/deployments/ecr/main.tf
```diff
@@ -211,36 +211,45 @@ resource "aws_ecr_lifecycle_policy" "ecr_lifecycle_policy" {
...
+        "rulePriority" : 1,
+        "description" : "Rule 1",
+        "selection" : {
+          "tagStatus" : "tagged",
+          "tagPrefixList" : ["deployed-to"],
+          "countType" : "imageCountMoreThan",
+          "countNumber" : 1
+        },
+        "action" : {
+          "type" : "expire"
```
theseanything, line 226: Not sure I understand this rule - it reads as if its to expire any image with the tag prefix "deployed-to", rather than retain.

terraform/deployments/ecr/main.tf
```diff
@@ -211,36 +211,45 @@ resource "aws_ecr_lifecycle_policy" "ecr_lifecycle_policy" {
...
-                "countNumber": 20
-            },
-            "action": {
-                "type": "expire"
-            }
+  policy = jsonencode({
+    "rules" : [
+      {
+        "rulePriority" : 1,
+        "description" : "Rule 1",
```
theseanything, line 218: Missing description

terraform/deployments/ecr/main.tf
```diff
@@ -211,36 +211,45 @@ resource "aws_ecr_lifecycle_policy" "ecr_lifecycle_policy" {
...
+          "countNumber" : 30
+        },
+        "action" : {
+          "type" : "expire"
+        },
+        "rulePriority" : 3,
+        "description" : "Keep last 20 images",
+        "selection" : {
+          "tagStatus" : "tagged",
+          "tagPrefixList" : ["v"],
```
theseanything, line 243: Nit: I don't think we prefix any images tags with "v"?

theseanything (changes requested):

terraform/deployments/ecr/main.tf
marcpomfret, line 218: Description should probably be something like 'Keep images tagged as "deployed-to-<env>"'

theseanything:
Closed in favour of https://github.com/alphagov/govuk-infrastructure/pull/765

PR #740, opened 2022-09-09
Switch to static provisioning for ClamAV EFS.

sengi:
Turns out the dynamic provisioning as implemented by Amazon's EFS CSI driver solves a different problem than the one we have: it creates a unique directory an [EFS access point](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html) for each PVC, whereas we just want to have a bunch of pods mount the same NFS volume.

Our use case is ideally suited for static provisioning, given that we're  using NFS. So all we need do is define the PersistentVolume statically in Terraform and refer to it from a PVC defined in Helm.

Our statically-defined PV refers to the EFS volume via an EFS access point, which lets us configure Amazon to set an initial directory with our desired permissions etc.

Bear in mind that the PV is essentially just a representation of a `mount_nfs` command line in our use case. That is, it's just the place where we configure the hostname to connect to and the mount options to use. The EFS access point (i.e. the settings we're telling Amazon to use when we connect to the NFS server) is [specified as just another mount option](https://docs.aws.amazon.com/efs/latest/ug/mounting-access-points.html).

This removes the need for the EFS CSI controller while still solving the problems of uid mapping and creating an initial directory in the remote filesystem with appropriate permissions etc.

The DNS name of the EFS filesystem [resolves automatically to the endpoint in the same AZ as the caller](https://docs.aws.amazon.com/efs/latest/ug/mounting-fs-mount-cmd-dns-name.html) so we should still have regional availability for this storage.

We also move these resources into the `govuk-publishing-infrastructure` module as they're app-specific and GOV.UK-specific, rather than base cluster services.


sengi:
Closing this since it turns out that it's not feasible to mount an EFS volume via an [access point](https://docs.aws.amazon.com/efs/latest/ug/efs-access-points.html) without using the EFS CSI plugin — and that's a disproportionate amount of complexity for the problem we're trying to solve here, i.e. just having a bunch of pods mount the same (long-lived, statically-provisioned) NFS export.

It'll be way simpler not to use access points and just use Volume and VolumeMount within the pod spec. We can solve the bootstrap problem (which we don't even have yet because we can use the existing asset-manager EFS) by running a Job as uid 0 in the `cluster-services` namespace to do a `mkdir -p` and `chown`. That's super simple and low maintenance compared to managing a controller, storageclass, daemonset, pv/pvcs, EFS access point etc. etc.

PR #741, opened 2022-09-12
Fix #720: fix origin security groups

fredericfran-gds:
1. Use the correct terraform state file
2. Convert IPs into CIDR

Tested via terraforming in integration.


nsabri1 (approved):
OK

PR #742, opened 2022-09-12
Only build non-existing ECR images

fredericfran-gds:
If an ECR image exists, we do not build it again.

Future work:
Right now we use the tag to determine whether to rebuild or not and the tag comes from GitHub repo. This logic may need to be changed when images are rebuild due to base image changes due to updates.

Ref:
1. [trello card](https://trello.com/c/nFlcYHdt/1005-dont-re-build-image-if-already-existing)
2. [stack overflow](https://stackoverflow.com/questions/30543409/how-to-check-if-a-docker-image-with-a-specific-tag-exist-locally)

### DIFF
--- a/.github/workflows/ci-ecr.yaml
+++ b/.github/workflows/ci-ecr.yaml
@@ -86,6 +86,15 @@ jobs:
           LATEST_GIT_SHA=$(git ls-remote origin HEAD | cut -f 1)
           IMAGE_TAG_ARGS="-t ${ECR_REGISTRY}/${ECR_REPOSITORY}:${IMAGE_TAG}"
 
+          check_image_in_ecr() {
+            aws ecr describe-images --repository-name="${1}" --image-ids=imageTag="${2}"
+          }
+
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-ecr.yaml
```diff
@@ -86,6 +86,16 @@ jobs:
           LATEST_GIT_SHA=$(git ls-remote origin HEAD | cut -f 1)
           IMAGE_TAG_ARGS="-t ${ECR_REGISTRY}/${ECR_REPOSITORY}:${IMAGE_TAG}"
 
+          check_image_in_ecr() {
+            aws ecr describe-images --repository-name="${1}" --image-ids=imageTag="${2}"
+            return $?
```
sengi, line 91: No need to explicitly return the exit code - the value of a shell function is its exit status already.
```suggestion
```

sengi (approved):
Yay! Thanks 🙌

.github/workflows/ci-ecr.yaml
fredericfran-gds: Will do, thanks

PR #743, opened 2022-09-14
Fix missing image tag output when skip build

theseanything:
Logic was added to skip rebuilding an image that already exists - however it failed to output the image tag which is used by the deployment step.


nsabri1 (approved):
LGTM

fredericfran-gds:
Thanks for fixing

PR #744, opened 2022-09-15
Fix www-origin security groups

fredericfran-gds:
For some reason, the wrong list were concanated for www-origin office and fastly security groups.

This fixes that, tested in integration


sengi (approved):
Thanks!

PR #746, opened 2022-09-26
annotated argo-workflow sa with push-to-ecr role arn

desenhar:
So that the update-image-tag pod can make calls to ECR, it will assume the SA that has the iam-role associated.
This service account seemed convenient to use as it exists in the same namespace the update-image-tag pods are created in.


nsabri1 (approved):
Assuming you have tested this already - OK

PR #758, opened 2022-10-14
Add IAM Role Service Account to retag images

theseanything:
This creates an IAM Role that gives permissions needed to retag existing images in AWS ECR. This is used by a Argo Workflow via Service Account which will add `deployed-to-<env>` tag to images that have been deployed. https://github.com/alphagov/govuk-helm-charts/pull/629

Tested by deploying with Terraform and using Service Account to push tag.

### DIFF
--- a/terraform/deployments/cluster-services/argo_workflows.tf
+++ b/terraform/deployments/cluster-services/argo_workflows.tf
@@ -1,0 +1,44 @@
+locals {
+  tag_image_service_account_name = "add-tag-to-image-workflow"
+}
+
+module "tag_image_iam_role" {
+  source  = "terraform-aws-modules/iam/aws//modules/iam-role-for-service-accounts-eks"
+  version = "~> 5.5"
+
+  role_name        = "${local.tag_image_service_account_name}-${data.terraform_remote_state.cluster_infrastructure.outputs.cluster_id}"
+  role_description = "Role for the add-tag-to-image Argo Workflow. Corresponds to ${local.tag_image_service_account_name} k8s ServiceAccount."
+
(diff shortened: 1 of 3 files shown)


terraform/deployments/cluster-services/argo_workflows.tf
```diff
@@ -0,0 +1,46 @@
...
+      namespace_service_accounts = ["${var.apps_namespace}:${local.tag_image_service_account_name}"]
+    }
+  }
+}
+
+data "aws_iam_policy_document" "tag_image" {
+  statement {
+    actions = [
+      "ecr:GetAuthorizationToken",
+    ]
```
sengi, line 24: optional tiniest of nits: maybe slightly cleaner to write this on a single line like on line 33
```suggestion
    actions = ["ecr:GetAuthorizationToken"]
```

terraform/deployments/cluster-services/argo_workflows.tf
```diff
@@ -0,0 +1,46 @@
...
+    resources = ["*"]
+  }
+
+  statement {
+    actions = [
+      "ecr:BatchGetImage",
+      "ecr:PutImage",
+    ]
+    resources = ["arn:aws:ecr:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:repository/*"]
```
sengi, line 33: This assumes that the registry is in-region, but I think that's a reasonable assumption (and moot for the time being anyway!) 👍

sengi (approved):
Nice job on keeping this clean! The IRSA stuff can be a bit gnarly sometimes.

PR #759, opened 2022-10-14
Rename the GitHub Argo secrets

theseanything:
We are now running the CD process in Production, however we don't need to be concerned about the environment name in the secret.


nsabri1 (approved):
OK

PR #764, opened 2022-10-20
Add "release_" prefix to the image tag value

theseanything:
To implement ECR lifecycle rules, a rule needs to be able to identify image by a tag prefix. Previously image tags were just the commit SHA, which gave no predicable pattern for identifying images associated with a release.


.github/workflows/ci-ecr.yaml
```diff
@@ -82,8 +82,11 @@ jobs:
           ECR_REPOSITORY: ${{ inputs.ecr_repository_name }}
           DOCKER_BUILDKIT: "1"
         run: |
-          IMAGE_TAG=$(git rev-parse HEAD)
-          LATEST_GIT_SHA=$(git ls-remote origin HEAD | cut -f 1)
+          LOCAL_HEAD_SHA="$(git rev-parse HEAD)"
+          REMOTE_HEAD_SHA=$(git ls-remote origin HEAD | cut -f 1)
+
+          IMAGE_TAG="release_${LOCAL_HEAD_SHA}"
```
sengi, line 88: `-` seems to be much more common than `_` in image tags (e.g. looking at the top few images on DockerHub) so I'd recommend we go with `-`. What do you reckon?

```suggestion
          IMAGE_TAG="release-${LOCAL_HEAD_SHA}"
```

sengi (approved):
Good call on adding a prefix 👍

.github/workflows/ci-ecr.yaml
theseanything: nice shout - good to go with conventions!

PR #765, opened 2022-10-20
Update ECR lifecycle rules

theseanything:
This PR updates the ECR lifecycle rules to help manage and clean up old images in ECR. AWS currently doesn't support a rules that explicitly dictate with images to **_NOT_** expire, so we have to leverage the rule priority behaviour. i.e. 

> An image that matches the tagging requirements of a rule cannot be expired by a rule with a lower priority.

Ref: https://docs.aws.amazon.com/AmazonECR/latest/userguide/LifecyclePolicies.html#lp_evaluation_rules

The new rules have the following logic:

1. Find images with the "deployed-to" tag prefix and only delete images if not the 100 most recent. This is an arbitrary high number to mark images with "deployed-to" tag prefix, so they aren't deleted by this rule, but also are protected from being deleted by lower priority rules.
2. Find untagged images older than a day and delete. We want to make sure don't keep untagged images lying around
3. Find images with the "release-" tag prefix and only delete if not the 20 most recent. This is arbitrary number, so we protect the 20 of our most recent "release" images. This is to ensure older images are present in case we need to rollback.
4. Find images older that 30 days and delete. This is to ensure any other old images with tags that don't conform to our tagging conventions are removed. This will not delete images that are marked by higher priority rules and older than 30 days.

### DIFF
--- a/terraform/deployments/ecr/main.tf
+++ b/terraform/deployments/ecr/main.tf
@@ -210,36 +210,62 @@ resource "aws_ecr_lifecycle_policy" "ecr_lifecycle_policy" {
   for_each   = toset([for repo in local.repositories : aws_ecr_repository.repositories[repo].name])
   repository = each.key
 
-  policy = <<EOF
-  {
-    "rules": [
-        {
-            "rulePriority": 1,
-            "description": "Expire images older than 30 days",
-            "selection": {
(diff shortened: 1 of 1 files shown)


sengi (approved):
LGTM

PR #766, opened 2022-10-25
Mapit is now retired.

KludgeKML:
..and mapit ASGs scaled down to 0 (soon to be removed) so doesn't need any access rules.

https://trello.com/c/LMRiT7yz/1581-retire-mapit-and-mapit-scripts


theseanything (approved):

theseanything:
@KludgeKML assume this is the only bit of Mapit infra in this repo?

KludgeKML:
It was the only mention of mapit I could find in this repo, so I hope so!

PR #768, opened 2022-10-27
Improve the set automatic deploy workflow

theseanything:
This PR contains the following improvements:

- Remove functionality to reset the image tag (already removed from the Argo Workflow in https://github.com/alphagov/govuk-helm-charts/pull/662)
- Make the workflow name more human readable and clearer
- Separates the permission check into a separate step, so it consistent with the deploy workflow and easier to see in the GitHub UI
- Renames the workflow filename

Been tested in collections and works


.github/workflows/set-automatic-deploys.yaml
```diff
@@ -0,0 +1,99 @@
...
+# on:
+#   workflow_dispatch:
+#     inputs:
+#       automaticDeploysEnabled:
+#         description: 'Activate automatic deploys'
+#         required: false
+#         default: true
+#         type: boolean
+#       environment:
+#         description: 'Environment to deploy to'
```
sengi, line 15: optional: should this be "Environment to modify" or somesuch?

sengi (approved):

PR #774, opened 2022-11-08
Grant more permissions over the govuk-*-assets S3 buckets.

sengi:
Not all of these are used right now, but they could easily be needed in future with relatively small changes, like enabling or disabling versioning on the assets buckets.

Tested: uploading assets succeeds as far as `state: "uploaded"`.

https://trello.com/c/moEiT5BG

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/asset_manager_s3.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/asset_manager_s3.tf
@@ -6,17 +6,21 @@ resource "aws_iam_policy" "asset_manager_s3" {
     Version = "2012-10-17"
     Statement = [
       {
-        Effect   = "Allow"
-        Action   = "s3:ListBucket"
+        Effect = "Allow"
+        Action = [
+          "s3:GetBucketLocation",
+          "s3:ListBucket",
+        ]
         Resource = data.terraform_remote_state.infra_assets.outputs.asset_manager_bucket_arn
       },
       {
         Effect = "Allow"
         Action = [
-          "s3:PutObjectAcl",
-          "s3:PutObject",
-          "s3:GetObject",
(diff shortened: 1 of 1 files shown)


terraform/deployments/govuk-publishing-infrastructure/asset_manager_s3.tf
```diff
@@ -6,17 +6,21 @@ resource "aws_iam_policy" "asset_manager_s3" {
...
+        ]
         Resource = data.terraform_remote_state.infra_assets.outputs.asset_manager_bucket_arn
       },
       {
         Effect = "Allow"
         Action = [
-          "s3:PutObjectAcl",
-          "s3:PutObject",
-          "s3:GetObject",
-          "s3:DeleteObject",
```
nsabri1, line 19: Delete is not needed?

nsabri1 (approved):
Question on deletion - otherwise looks good to me

terraform/deployments/govuk-publishing-infrastructure/asset_manager_s3.tf
sengi: Seems to be, e.g. [here](https://github.com/alphagov/asset-manager/blob/f4d0b87/lib/tasks/assets.rake#L2) and [here](https://github.com/alphagov/asset-manager/blob/f4d0b874/lib/s3_storage.rb#L37). There's a [DELETE method in the API](https://github.com/alphagov/asset-manager/blob/f4d0b874/lib/s3_storage.rb#L37) for the `assets/:id` resource.

sengi: Oh, just realised I might have misinterpreted your question. Delete is covered by the wildcard in `s3:*Object`.

PR #780, opened 2022-11-14
Remove `--fail-with-body` for set automatic deploys workflow

theseanything:
Fixes 447066c and was missing from dac1da3.


sengi (approved):
Thanks! Sorry about that.

PR #784, opened 2022-11-21
Use build-and-push-image in build-github-cli

richardTowers:
ci-ecr was removed in #769 in favour of build-and-push-image.yaml which was added in #763. This workflow was missed in the update.

It looks like the github-cli image has since disappeared from the registry - not sure if that's a problem with the image retention policy, or whether it was accidentally deleted some other way...

### DIFF
--- a/.github/workflows/build-github-cli-image.yaml
+++ b/.github/workflows/build-github-cli-image.yaml
@@ -1,13 +1,22 @@
 name: Build and publish github-cli image to ECR
 
 on:
+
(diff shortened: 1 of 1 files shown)


samsimpson1 (approved):

richardTowers:
Successful workflow run here: https://github.com/alphagov/govuk-infrastructure/actions/runs/3514007256/jobs/5887501201

samsimpson1:
LGTM

PR #786, opened 2022-11-21
Fix build and push image tagging

richardTowers:
This fixes a bug where only the first image tag would be pushed to ECR, not all of them. The cause was an inaccuracy in bash syntax (that would have worked fine in zsh):
    
In bash `echo "${array}"` only prints the first item in the array, not the whole thing (note, this is different in zshell).
    
This meant that `echo "fullImageTagList=${FULL_IMAGE_TAG_LIST}" >> $GITHUB_OUTPUT` was only collecting the first tag, not the full tag list.

I've also taken the opportunity to clarify the rest of the code handling this array a little bit, as I found it quite difficult to work out what this code was doing at first.

### DIFF
--- a/.github/workflows/build-and-push-image.yaml
+++ b/.github/workflows/build-and-push-image.yaml
@@ -96,7 +96,7 @@ jobs:
           fi
 
           echo "imageTag=${IMAGE_TAG}" >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


nsabri1 (approved):
Good catch

PR #790, opened 2022-11-24
Remove the AWS EFS CSI driver, which we're no longer using.

sengi:
We previously thought we'd need this for the Asset Manager EFS mount, but we don't.

Tested: rolled out to the integration cluster.


nsabri1 (approved):
Thanks for cleaning this up

PR #792, opened 2022-11-25
Add Licence reference in README

catalinailie:
We want to make the reference to the licence in the README consistent with the GDS Way.

Trello card: https://trello.com/c/P0DpFcwW/260-standardise-all-license-files-names-to-licence


sengi (approved):
Thanks!

PR #783, opened 2022-11-16
Update to latest ArgoCD, Argo Workflows and dependencies.

sengi:
Now that we're on Kubernetes 1.23, we can run ArgoCD 2.5 (the latest minor version) without API compatibility issues.

Also update to Workflows 3.43 and the latest NATS releases (dependencies of Argo Events).

Changelogs:
- https://github.com/argoproj/argo-cd/blob/master/CHANGELOG.md
- https://github.com/argoproj/argo-workflows/blob/master/CHANGELOG.md
- https://github.com/nats-io/nats-streaming-server/releases/
- https://github.com/nats-io/prometheus-nats-exporter/releases

Tested: applied in integration.

PR #794, opened 2022-11-30
Use Jetstream instead of Nats streaming server

theseanything:
Nats streaming server has been deprecated in favor of Jetstream. Also this fixes an issue with creating new EventBus resources, as they don't support the new versions of Nat streaming server - which [we recently updated to](https://github.com/alphagov/govuk-infrastructure/commit/fd222318bf1c21aab2f8dc347ebc9575ea87ae79).

Deprecation notice: https://docs.nats.io/legacy/stan/intro#:~:text=NATS%20Streaming%20is%20a%20data,with%20the%20core%20NATS%20platform.

Argo Events docs: https://argoproj.github.io/argo-events/concepts/eventbus/

### DIFF
--- a/terraform/deployments/cluster-services/argo.tf
+++ b/terraform/deployments/cluster-services/argo.tf
@@ -296,13 +296,15 @@ resource "helm_release" "argo_events" {
       replicas = var.default_desired_ha_replicas
     }
     configs = {
-      nats = {
+      jetstream = {
         versions = [
           {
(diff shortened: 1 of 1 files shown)


samsimpson1 (approved):

sengi:
Oof, good find, thanks! Sorry for not testing it properly in #783. Good that we now have one fewer thing that's <1.0.

PR #800, opened 2022-12-14
Add reusable GHA workflows for CI

theseanything:
This adds workflows that can be used by CI workflows in apps to run the different process required. We are using reusable workflow to help centralise and minimise the amount of config we have to manage. This adds the initial set of jobs that are need by Whitehall for CI.

An example of an app using the workflow: https://github.com/alphagov/whitehall/pull/7155/files
An example of the workflow run: https://github.com/alphagov/whitehall/actions/runs/3697377538

The cache dependencies workflow can be run before other jobs to cache shared dependencies.

### DIFF
--- a/.github/workflows/ci-test-ruby.yaml
+++ b/.github/workflows/ci-test-ruby.yaml
@@ -1,0 +1,195 @@
+name: Test Ruby
+
+on:
+  workflow_call:
+    inputs:
+      extraSystemDependencies:
+        description: 'Install additional system dependencies'
+        required: false
+        default: ''
+        type: string
+      enableMongoDB:
+        description: 'Run MongoDB'
+        required: false
+        default: false
+        type: boolean
+      enableElasticsearch:
+        description: 'Run Elasticsearch'
+        required: false
+        default: false
+        type: boolean
+      enableRedis:
+        description: 'Run Redis'
+        required: false
+        default: false
+        type: boolean
+      enableMySQL:
+        description: 'Run MySQL'
+        required: false
+        default: false
+        type: boolean
+      enablePostgres:
+        description: 'Run Postgres'
+        required: false
+        default: false
+        type: boolean
+
+jobs:
+  test-ruby:
+    name: Test Ruby
+    runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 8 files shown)


.github/workflows/ci-test-ruby.yaml
```diff
@@ -0,0 +1,76 @@
...
+    name: Test Ruby
+    runs-on: ubuntu-latest
+    services:
+      mysql:
+        image: mysql:8.0
+        env:
+          MYSQL_ROOT_PASSWORD: root
+        ports:
+          - 3306:3306
+        options: >-
```
sengi, line 23: very optional: might be worth tuning MySQL to use less ram [like we did in the old e2e tests](https://github.com/alphagov/publishing-e2e-tests/blob/main/docker-compose.yml#L67), which should result in slightly faster tests

.github/workflows/ci-test-ruby.yaml
```diff
@@ -0,0 +1,76 @@
...
+          --health-cmd "mysqladmin ping"
+          --health-start-period 10s
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+      redis:
+        image: redis
```
sengi, line 30: optional: consider redis:alpine (should be slightly quicker and use slightly less resources)

sengi (approved):
Awesome!

.github/workflows/ci-test-ruby.yaml
theseanything: Annoyingly there doesn't seem to be a way to pass in command args for service containers. 😞

sengi: Don't worry about it for this PR, but are you sure? It looks like the contents of `options` are [passed as the command line to `docker run`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idservicesservice_idoptions), so I'd expect to just be able to override the image's `CMD` by adding commands at the end of `options` (like `"mysqld", "--foo", "--bar"). I haven't actually tried this though, so apologies if this doesn't work.

PR #802, opened 2022-12-23
Add composite actions for backing services

theseanything:
This adds composite action definitions for GitHub Actions that can be used by other workflows to set up the following backing services:

- Redis
- MySQL
- Postgres
- Elasticsearch
- MongoDB

They are needed for CI workflows to run tests for applications. These backing services rely on creating a docker container. In future we can improve this by adding health checks to the containers and waiting for services until they're functional. We will probably also want to allow the ability to use an input to specify the container version.


marcpomfret (approved):
LGTM

PR #804, opened 2022-12-28
Add precompile Rails assets composite action and workflow 

theseanything:
This PR adds two reusable components for GitHub Actions:

- A composite action that encapsulates the logic to precompile Rails assets and cache them for other workflows
- A workflow that can be used as a job to precompiles assets. Usefully as a caching step and to test if assets precompile successfully.


nsabri1 (approved):
OK

PR #808, opened 2022-12-30
Add minor version to postgres backing service

theseanything:
We are using exactly 13.7 in production for all applications that use postgres. Pinning the minor version to be sure we're testing against the exact version we're using. In future we should add an input to allow workflows to specify which version to use.


nsabri1 (approved):
ok

PR #813, opened 2023-01-04
Add workflows with `.yml` extension

theseanything:
This is part of making the naming conventions consistent.

This is adds copies of workflows that previously had the extension `.yaml`, with the new `.yml` extension. The old workflows have a deprecation noticed and should not be updated. This is intermediate step whilst other repositories that use these workflows are updated, then the old workflows with `.yaml` will be removed.

### DIFF
--- a/.github/workflows/build-and-push-image.yml
+++ b/.github/workflows/build-and-push-image.yml
@@ -20,11 +20,6 @@
 
 # REUSABLE WORKFLOW
 # -----------------
-
-#
-# WORKFLOW DEPRECATED AND SOON WILL BE REMOVED IN FAVOR OF build-and-push-image.yml
-#
-
 name: Build and push image
 
 on:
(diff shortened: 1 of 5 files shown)


.github/workflows/build-and-push-image.yaml
```diff
@@ -20,6 +20,11 @@
 
 # REUSABLE WORKFLOW
 # -----------------
+
+#
+# WORKFLOW DEPRECTATED AND SOON WILL BE REMOVED IN FAVOR OF build-and-push-image.yml
```
sengi, line 25: nit: typo (similarly on other files)
```suggestion
# WORKFLOW DEPRECATED AND SOON WILL BE REMOVED IN FAVOR OF build-and-push-image.yml
```

sengi (approved):

sengi:
(I'm assuming these are just straight copies apart from the deprecation comment, right?)

theseanything:
> (I'm assuming these are just straight copies apart from the deprecation comment, right?)

Yeah exactly the same workflows

PR #815, opened 2023-01-05
Grant nodes access to content-publisher S3 bucket.

sengi:
Content Publisher needs read-write access to an S3 bucket that it uses for file uploads. For now, just grant access to worker nodes as we've done for other apps like asset-manager.

Later (not before launch) we'll consider rolling out IRSA / pod identity.

Tested: applied in integration.

Rollout: depends on https://github.com/alphagov/govuk-aws/pull/1686

Related: https://github.com/alphagov/content-publisher/pull/2705


nsabri1 (approved):
LGTM

PR #818, opened 2023-01-13
Improve cache keys for compiled assets

theseanything:
This moves to cache assets per SHA, rather than dependent a file hash as asset compilation can probably be dependent on more than the files listed in the previous cache key.

We also don't need to delete old compiled assets if we're storing per SHA.


sengi (approved):
Nice!

DilwoarH:
Nice 👌

PR #819, opened 2023-01-18
Remove caching rails assets between commits

theseanything:
This removes extra cache keys that allows GitHub workflows from restoring assets from older commits. This is due to issue when there are old versions of assets for the same source file. The Jasmine test suite will sometimes attempt to test an older asset version.


ollietreend (approved):
Thanks for this @theseanything 👍🏻

PR #821, opened 2023-01-25
Enable automatic OAuth login for Grafana

theseanything:
This saves you having to click the login button if you're already logged into GitHub.


sengi (approved):
Yay, thanks!

PR #823, opened 2023-01-26
Grant nodes access to csvs S3 buckets.

marcpomfret:
Applied in integration, buckets now accessible by apps and files can be downloaded by recipient

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/csvs_s3.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/csvs_s3.tf
@@ -1,0 +1,45 @@
+resource "aws_iam_policy" "write_csvs_buckets" {
+  name        = "csvs_s3"
+  description = "Read and write to this environment's CSVS s3 buckets"
+
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        Effect = "Allow"
+        Action = [
+          "s3:GetBucketLocation",
+          "s3:ListBucket",
+        ]
+        Resource = [
+          "arn:aws:s3:::govuk-${var.govuk_environment}-content-data-csvs",
+          "arn:aws:s3:::govuk-${var.govuk_environment}-specialist-publisher-csvs",
+          "arn:aws:s3:::govuk-${var.govuk_environment}-support-api-csvs",
+          "arn:aws:s3:::govuk-${var.govuk_environment}-whitehall-csvs"
+        ]
+      },
+      {
+        Effect = "Allow"
+        Action = [
+          "s3:*MultipartUpload*",
+          "s3:*Object",
+          "s3:*ObjectAcl",
+          "s3:*ObjectVersion",
+          "s3:GetObject*Attributes",
+        ]
+        Resource = [
+          "arn:aws:s3:::govuk-${var.govuk_environment}-content-data-csvs/*",
+          "arn:aws:s3:::govuk-${var.govuk_environment}-specialist-publisher-csvs/*",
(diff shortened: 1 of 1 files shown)


terraform/deployments/govuk-publishing-infrastructure/csvs_s3.tf
```diff
@@ -0,0 +1,35 @@
...
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        Effect = "Allow"
+        Action = [
+          "s3:GetBucketLocation",
+          "s3:ListBucket",
+        ]
+        Resource = ["arn:aws:s3:::${var.govuk_environment}content-data-csvs", "arn:aws:s3:::${var.govuk_environment}-specialist-publisher-csvs", "arn:aws:s3:::${var.govuk_environment}-support-api-csvs", "arn:aws:s3:::${var.govuk_environment}-whitehall-csvs"]
```
sengi, line 14: Slightly more readable like this?

```suggestion
        Resource = [
          "arn:aws:s3:::${var.govuk_environment}content-data-csvs",
          "arn:aws:s3:::${var.govuk_environment}-specialist-publisher-csvs",
          "arn:aws:s3:::${var.govuk_environment}-support-api-csvs",
          "arn:aws:s3:::${var.govuk_environment}-whitehall-csvs",
        ]
```

(and similarly below)

sengi (approved):
👍

terraform/deployments/govuk-publishing-infrastructure/csvs_s3.tf
marcpomfret: Yeah good shout, will get them updated

terraform/deployments/govuk-publishing-infrastructure/csvs_s3.tf
```diff
@@ -0,0 +1,45 @@
...
+          "s3:*Object",
+          "s3:*ObjectAcl",
+          "s3:*ObjectVersion",
+          "s3:GetObject*Attributes",
+        ]
+        Resource = [
+          "arn:aws:s3:::${var.govuk_environment}content-data-csvs",
+          "arn:aws:s3:::${var.govuk_environment}-specialist-publisher-csvs",
+          "arn:aws:s3:::${var.govuk_environment}-support-api-csvs",
+          "arn:aws:s3:::${var.govuk_environment}-whitehall-csvs"
```
sengi, line 34: I think these need the `/*` suffixes so that they match objects in the buckets (as opposed to just the buckets themselves). IAM is super annoying like that unfortunately.

terraform/deployments/govuk-publishing-infrastructure/csvs_s3.tf
```diff
@@ -0,0 +1,45 @@
+resource "aws_iam_policy" "csvs_bucket_polices" {
```
sengi, line 1: nit: I only just noticed the typo in the name here 😅 Now that I think about the name, how about:
```suggestion
resource "aws_iam_policy" "write_csvs_buckets" {
```

terraform/deployments/govuk-publishing-infrastructure/csvs_s3.tf
marcpomfret: Damn, your right i copy pasted over it when i changed the format.

terraform/deployments/govuk-publishing-infrastructure/csvs_s3.tf
marcpomfret: haha, yeah i'd not even noticed that, going to blame the dyslexia though

PR #824, opened 2023-01-31
Fix Grafana SSO auth with GitHub

theseanything:
Fix formatting of the config by unnesting `generic_oauth` from `auth` as it should be discrete key of `auth.generic_oauth`.


martynipratt (approved):
Fixed Github login for me

PR #825, opened 2023-02-02
Fix TODO: Update curl command, we are on 7.81

dj-maisy:
## What?
As the `ubuntu-latest` GHA runner image is now on Ubuntu 22.04 and [comes pre-packaged with curl 7.81](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md) (correct at time of writing), we can now address the TODO in the Deploy workflow.

* Replaces `-f` with `--fail-with-body`, as suggested by the TODO.
* Swap `-s` for `--silent` for improved legibility.

Hopefully this should result in more useful output in the event of a failure.


sengi (approved):
Thanks!

PR #826, opened 2023-02-03
Add IAM role and policy for CKAN service account

samsimpson1:
This role allows CKAN to access the organogram S3 buckets. The corresponding service account is created in the CKAN helm chart.

### DIFF
--- a/terraform/deployments/cluster-infrastructure/ckan_iam.tf
+++ b/terraform/deployments/cluster-infrastructure/ckan_iam.tf
@@ -1,0 +1,30 @@
+
+module "ckan_iam_role" {
+  source                        = "terraform-aws-modules/iam/aws//modules/iam-assumable-role-with-oidc"
+  version                       = "~> 4.0"
+  create_role                   = true
+  role_name                     = "${var.ckan_service_account_name}-${var.cluster_name}"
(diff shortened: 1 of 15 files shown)


kentsanggds (approved):
Not sure if we need the test bucket, but I guess it could be used to test things when developing CKAN locally.

terraform/deployments/cluster-infrastructure/ckan_iam.tf
```diff
@@ -0,0 +1,30 @@
+
```
sengi, line 1: Seems we don't have any module which is a great fit for this at the moment. It's not cluster infrastructure because it's app-specific, but it's also not really govuk-publishing-infrastructure either.

For now I'd probably go with govuk-publishing-infrastructure (assuming you'd rather avoid the faff of introducing a new root module in this PR) but at some point in the near future we probably want a `data-gov-uk-infrastructure` module. What do you reckon?

sengi (approved):
Cool. Good call on using IRSA / pod identity for this 👍

terraform/deployments/cluster-infrastructure/ckan_iam.tf
samsimpson1: I've added a new root module like you suggested, it does make sense to keep the DGU resources isolated from the main stack.

sengi (approved):
Ooh that's fab, thanks! ❤️

PR #831, opened 2023-02-10
Don't run 3 Redis replicas for ArgoCD in integration.

sengi:
High-availability Redis for ArgoCD on the integration cluster is overkill; we can live with a potential 1-minute blip in availability once a month or so (probably much less in reality) when there's a zone outage.

ArgoCD treats the cache as volatile, so durability of the cache data is not a factor.
https://www.github.com/argoproj/argo-cd/discussions/6618#discussioncomment-955540

Tested: applied in integration.


nsabri1 (approved):
OK

PR #832, opened 2023-02-10
Manage data.gov.uk resources with argo

samsimpson1:


sengi:
There's not a lot of context here. It looks like you're adding a separate installation of Argo - what's the reason for that vs. just using a separate project on the existing Argo installation or managing the chart via `helm upgrade` (which I thought was the original intention)?

sengi:
(ArgoCD is designed to be able to manage multiple *clusters* let alone multiple projects, namespaces etc.)

sengi:
Oh I see, sorry misinterpreted this. I read "argo-bootstrap" as the chart that's installing Argo, but it's just a chart for installing the app-of-apps. That seems reasonable.

sengi (approved):
LGTM (now that I understand what's going on 😅). Please add a description though!

kentsanggds:
Thanks, I'll merge it in as Sam is off next week

sengi:
Bit of background as to why we explain design choices in commit messages and PR descriptions on Replatforming team:

- It's easily searchable and discoverable in the context when someone's trying to figure out the rationale behind the code in months/years to come (and sometimes that someone is yourself 😄).
- It saves reviewers' time (i.e. your time!)
- It tends to make us think in a bit more depth about our design choices and how understandable our solutions are likely to be to others.

If it seems like a hassle, try thinking of it as a small price to pay for not having to write [ADRs](https://adr.github.io/) (yuck!) This way we get the same or better maintainability benefits with less bureaucracy (and less bikeshedding!)

PR #834, opened 2023-03-03
Allow Publishing API to write its event logs to S3.

sengi:
Publishing API has a [Rake task](https://github.com/alphagov/publishing-api/blob/main/lib/tasks/events.rake) that currently runs daily via Jenkins and pushes logs of publishing events to S3.

Tested: applies cleanly in integration.


dj-maisy (approved):
Nice use of wildcard permissions there. 👍

PR #836, opened 2023-03-05
Update cluster-autoscaler chart to 9.25.0.

sengi:
No breaking changes. https://github.com/kubernetes/autoscaler/blob/master/charts/cluster-autoscaler/README.md


marcpomfret (approved):
ok

PR #838, opened 2023-03-06
Attach sagemaker policy to worker IAM role

samsimpson1:
This should fix [these sentry errors](https://govuk.sentry.io/issues/3686364541/?project=1461905)


marcpomfret (approved):
ok

PR #839, opened 2023-03-07
Update container logs path to reflect dockerd->containerd.

sengi:
Because of the removal of Dockershim in k8s 1.24, Amazon has switched EKS from dockerd to containerd in that release. This results in container log files being written to a different directory on the kubelet.


marcpomfret (approved):
ok

PR #843, opened 2023-03-10
Remove fastly-exporter.

sengi:
We're not currently using this, and if/when we start using it then we'd use the [prometheus-community chart](https://www.github.com/prometheus-community/helm-charts/tree/main/charts/prometheus-fastly-exporter) rather than rolling our own.

This should resolve a few Kubesec warnings about missing resource requests etc.


martynipratt (approved):
LGTM

PR #844, opened 2023-03-11
Productionise Prometheus/Alertmanager config.

sengi:
- Enable the kube-prometheus-stack chart's default PodDisruptionBudget for Prometheus and Alertmanager. The default is `minAvailable: 1`, which is fine.
- Set pod anti-affinity for Prometheus and Alertmanager so they can't end up with multiple replicas on the same node.
- Increase Prometheus metrics rentention from 10 days to 90 days and set the sampling (and eval) interval to 1 minute.
- Enable the use of ServiceMonitors and Probes in any namespace, like we already do for PodMonitors.
- Fix a weird variable name.

We only want PDB and anti-affinity where we're running more than one replica, i.e. in staging and production.

### DIFF
--- a/terraform/deployments/cluster-services/argo.tf
+++ b/terraform/deployments/cluster-services/argo.tf
@@ -37,7 +37,7 @@ resource "helm_release" "argo_cd" {
       # server from upgrading the request after TLS termination.
       extraArgs = ["--insecure"]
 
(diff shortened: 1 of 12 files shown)


nsabri1 (approved):
OK

PR #847, opened 2023-03-15
Update Argo chart versions.

sengi:
This updates Argo CD and Argo Workflows to the latest versions. No incompatible changes.

Also stop pinning the NATS Jetstream image versions. We don't have a mechanism to keep them up to date and the risk of sticking with old patch versions is greater than the risk of a bad release.


nsabri1 (approved):
LGTM

sengi:
Partially reverted in #854; turns out there's something buggy about Argo Workflows (or our setup of it, perhaps) that leads to a delayed / action-at-a-distance failure when using the default NATS config rather than passing in specific version like we were going.

PR #848, opened 2023-03-16
Add metrics relabel for CKAN

kentsanggds:
## What

Relabel the CKAN metrics so that existing dashboards can be used in grafana to pick them up.

## Reference 

https://trello.com/c/3ARO34au/1149-relabel-ckan-metrics-to-match-ruby-app-metrics


sengi:
The relabel itself is nice and straightforward though, good call!

kentsanggds:
I'll need to look into whether the relabelling can be done somewhere else

kentsanggds:
Looks like it can be added to the pod monitor so I'll try that out.

kentsanggds:
Managed to add it to podmonitor.yaml on the ckan helm chart instead

PR #853, opened 2023-03-29
Send PROMOTE_DEPLOYMENT variable to argo workflows

theseanything:
This adds an extra attribute sent to argo workflow for the deployment pipeline. PROMOTE_DEPLOYMENT is essentially the inverse to "MANUAL_DEPLOY". This allows us to be explicit when we want to promote deployment and make the logic confusing by not having to use the inverse of MANUAL_DEPLOY.


nsabri1 (approved):
OK

PR #854, opened 2023-03-30
Set allowed version of NATS Jetstream image

theseanything:
When starting the eventbus, its tries to use a specific version of the NATS Jetstream image. The default config only allows use of the image tagged latest, and hence the eventbus fails to start. Explicitly allowing this version for now as a quick fix, until we figure out how to not specify a specific allowed version.


sengi (approved):
Thanks for catching this!

PR #859, opened 2023-04-20
Move GitHub project to Terraform Cloud

theseanything:
This allows us to use Terraform Cloud to plan and apply the Terraform to conifgure GitHub instead of doing it locally.

### DIFF
--- a/terraform/deployments/github/main.tf
+++ b/terraform/deployments/github/main.tf
@@ -1,5 +1,10 @@
 terraform {
-  backend "s3" {}
+  cloud {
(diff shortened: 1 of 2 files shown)


sengi (approved):
At some point we should move those CKAN vars into their own file. (don't worry about it for this PR though!)

sengi:
Oops sorry both, didn't see you'd already assigned a reviewer!

theseanything:
@sengi no problem! Which CKAN vars did you mean?

PR #861, opened 2023-04-20
Use GitHub App credentials for GitHub Provider

theseanything:
This allows terraform to use a GitHub App creds to access GitHub instead of a Personal Token, which needs to be assigned to a specific user.


sengi (approved):
I'm assuming the failed runs are a chicken/egg thing!

PR #876, opened 2023-05-02
Bump Redis cluster to large in integration

beccapearce:
The old redis instance before migration was a [cache.r4.large](https://github.com/alphagov/govuk-aws/blob/main/terraform/projects/app-backend-redis/main.tf#L31). We are bumping the size of this instance to be at least a large so we can run tasks on integration without maxing out the CPU.

### DIFF
--- a/terraform/deployments/variables/integration/common.tfvars
+++ b/terraform/deployments/variables/integration/common.tfvars
@@ -30,7 +30,7 @@ external_dns_subdomain    = "eks"
 www_dns_validation_rdata  = "8xpwlbcbmg9qjx9d2v.fastly-validations.com"
 
 frontend_memcached_node_type   = "cache.t4g.micro"
-shared_redis_cluster_node_type = "cache.t4g.medium"
+shared_redis_cluster_node_type = "cache.m6g.large"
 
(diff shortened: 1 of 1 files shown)


sengi (commented):
Maxing out the CPU isn't a bad thing in itself. Is there a particular thing that you're trying to make faster with this change?

sengi (dismissed):
Approving anyway cos I don't want to hold you up if you really need this, but please do check! :)

sengi (approved):
Ugh let's do this.

sengi:
We can figure out whatever nonsense is going on in integration later and just live with even more hilarious utilization stats in the meantime 🙃

sengi:
Rolling this out now (it'll drop everything in that Redis)

sengi:
Yeah I see where you're coming from and I agree with this change, but I think we need to be careful about drawing general conclusions on these things like we did with the RDS stuff (which turned out very poorly indeed).

To my mind, the convincing argument for chucking resources at this is that we're choosing to live with the funky dual-use of integration for user training. Otherwise there'd be a ton of benefit in catching these spiky usage issues in integration so that prod becomes more robust.

PR #879, opened 2023-05-10
Fix GitHub Actions OIDC trust policy

theseanything:
The trust policy incorrectly only applied the audience condition and not the subject condition. This was because there was duplicate JSON keys for "StringEquals", when actually the audience and subject conditions should be under the same "StringEquals" object.

Also removes the "GitPull", as not needed to mirror the repositories.

### DIFF
--- a/terraform/deployments/github/mirror.tf
+++ b/terraform/deployments/github/mirror.tf
@@ -26,11 +26,9 @@ resource "aws_iam_role" "github_action_mirror_repos_role" {
           "StringEquals" : {
             "token.actions.githubusercontent.com:sub" : [
               "repo:alphagov/govuk-infrastructure:ref:refs/heads/main"
-            ]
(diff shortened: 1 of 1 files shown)


sengi (approved):

terraform/deployments/github/mirror.tf
```diff
@@ -44,7 +44,6 @@ resource "aws_iam_role_policy" "github_action_mirror_repos_policy" {
     Statement = [
       {
         Action = [
-          "codecommit:GitPull",
           "codecommit:GitPush"
```
gclssvglx, line 47: My lack of understanding, but will removing GitPull have any adverse effects. Like it can't then pull (assuming it needs to)?

gclssvglx (approved):
Just an question - more for my education TBH.

PR #881, opened 2023-05-15
Remove sg rules for access to EC2 from EKS

theseanything:
The security groups have been removed in govuk-aws terraform and no longer exist.


sengi (approved):
Oh good catch, thanks!

sengi:
Sorry for the breakage!

PR #892, opened 2023-06-21
Update Prometheus stack, auth-proxy and pushgateway.

sengi:
Rollout: [update CRDs](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#from-46x-to-47x) first; should be no other forward-compat issues. Best to update using `helm upgrade` and then just use Terraform to check there are no diffs (because the Terraform Helm provider sucks at upgrading charts).

Tested: upgraded in staging

(ignore the failing TF Cloud runs; our TF Cloud setup isn't quite ready yet)


nsabri1 (commented):
ok

sengi:
Rolled out everywhere now.

nsabri1 (approved):
ok

PR #896, opened 2023-07-04
Drop a bunch of redundant/not-very-useful log fields.

sengi:
Some of these are plain duplicates; others are just not very useful or can trivially be synthesized from other fields.

Certain fields such as `agent.*` are "global" to Filebeat and have to be dropped in the top-level `processors` section, hence the two different sections in the config.

This should help reduce the pressure on LogIt quota.

Tested: rolled out in integration and staging, logs still being ingested and there's much less noise in the list of fields. Filebeat daemons look happy (`k -n cluster-services log ds/filebeat`).


theseanything (approved):
Fab!

PR #898, opened 2023-07-05
Avoid logging healthcheck requests to Logit.

sengi:
These are a non-negligible fraction of log traffic and they just clutter up query results (or complicate queries by trying to exclude them), so let's not pay for the inconvenience.

This won't stop us debugging healthchecks if we ever need to; it's easier to look directly at `kubectl log` etc. for that because it's real-time.

Also switch from the nasty HCL-encoded YAML to plain YAML for representing the Filebeat config. (Separate commit for ease of reviewing.)

Tested: applied in staging.

### DIFF
--- a/terraform/deployments/cluster-services/filebeat.yml
+++ b/terraform/deployments/cluster-services/filebeat.yml
@@ -1,0 +1,51 @@
+"filebeat.inputs":
+- "paths":
+  - "/var/log/containers/*.log"
+  "processors":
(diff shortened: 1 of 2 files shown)


theseanything (approved):

sengi:
Hm I think I may have fluffed my test queries on this one, it doesn't look like it's actually dropping any records after all. I'll see if I can fix it.

PR #904, opened 2023-07-11
Add Grafana Tempo for distributed tracing

theseanything:
This adds Terraform to deploy the Grafana Tempo helm chart (a distributed tracing tool) in to our EKS clusters. It has Terraform to create the relevant S3 bucket and IAM Role (used as the backing store).

Grafana Tempo integrates with our existing Grafana UI and is relative cheap/simple to run compared to other tracing tools (e.g. Jaeger/Zipkin/Sentry). Running it will allow us to better evaluate the value we get from tracing tools, and specifically if Tempo is the correct tool for us.

To send traces OpenTelemetry instrumentation need to be added/enabled to applications. We've shipped support for [OpenTelemetry instrumentation in govuk_app_config](https://github.com/alphagov/govuk_app_config/pull/303). This has been enabled and tested in integration.

### DIFF
--- a/terraform/deployments/cluster-services/tempo.tf
+++ b/terraform/deployments/cluster-services/tempo.tf
@@ -49,3 +49,81 @@ resource "aws_iam_policy" "tempo" {
   })
 }
 
+resource "helm_release" "tempo" {
+  depends_on = [module.tempo_iam_role, aws_s3_bucket.tempo]
+  chart      = "tempo-distributed"
+  name       = "tempo"
+  namespace  = local.monitoring_ns
+  repository = "https://grafana.github.io/helm-charts"
+  version    = "1.4.8" # TODO: Dependabot or equivalent so this doesn't get neglected.
+  values = [yamlencode({
+    reportingEnabled = false
+
+    ingester = {
+      persistence = {
+        enabled      = true
+        size         = "30Gi"
+        storageClass = "ebs-gp3"
+      }
+    }
+
+    metricsGenerator = {
+      enabled = true
+      config = {
+        storage = {
+          remote_write = [
+            {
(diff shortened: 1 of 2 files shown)


terraform/deployments/cluster-services/tempo.tf
```diff
@@ -0,0 +1,129 @@
+locals {
+  tempo_service_account = "tempo"
+  cluster_name          = data.terraform_remote_state.cluster_infrastructure.outputs.cluster_id
+}
+
+resource "aws_s3_bucket" "tempo" {
+  bucket = "govuk-${var.govuk_environment}-tempo"
+}
+
+module "tempo_iam_role" {
```
sengi, line 10: Cool, thanks for dealing with all the IRSA faff!

terraform/deployments/cluster-services/tempo.tf
```diff
@@ -0,0 +1,129 @@
...
+        grpc = {
+          enabled = true
+        }
+        http = {
+          enabled = true
+        }
+      }
+    }
+
+    overrides = <<EOT
```
sengi, line 120: optional: usually I'd suggest using `yamlencode()`, but tbh keeping it like this as relatively readable YAML is probably worth the loss of syntax checking. Could maybe wrap it in `yamlencode(yamldecode())` to get the syntax checking I guess. (I'm not sure even that's worth it though.)

sengi (approved):
Cool!

terraform/deployments/cluster-services/tempo.tf
theseanything: Good shout - just added yamlencode, small enough to be to difficult to read.

PR #906, opened 2023-07-17
Allow CKAN on EKS to access postgres RDS

kentsanggds:
Enable access to the postgres RDS from CKAN on EKS

### DIFF
--- a/terraform/deployments/datagovuk-infrastructure/remote.tf
+++ b/terraform/deployments/datagovuk-infrastructure/remote.tf
@@ -9,3 +9,22 @@ data "terraform_remote_state" "cluster_infrastructure" {
     region = data.aws_region.current.name
   }
 }
+
+data "terraform_remote_state" "app_govuk_rds" {
+  backend = "s3"
+
+  config = {
+    bucket = var.govuk_aws_state_bucket
+    key    = "blue/app-govuk-rds.tfstate"
+    region = data.aws_region.current.name
+  }
+}
+
+data "terraform_remote_state" "infra_security_groups" {
+  backend = "s3"
+  config = {
+    bucket = var.govuk_aws_state_bucket
+    key    = "govuk/infra-security-groups.tfstate"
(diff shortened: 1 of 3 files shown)


terraform/deployments/datagovuk-infrastructure/remote.tf
```diff
@@ -9,3 +9,28 @@ data "terraform_remote_state" "cluster_infrastructure" {
...
+data "terraform_remote_state" "app_govuk_rds" {
+  backend = "s3"
+
+  config = {
+    bucket = var.govuk_aws_state_bucket
+    key    = "blue/app-govuk-rds.tfstate"
+    region = data.aws_region.current.name
+  }
+
+  # TODO: hack because govuk-aws/app-govuk-rds is not terraformed in `test`.
```
sengi, line 22: Feel free to drop the config for the test account if that's easier; it's unlikely we'll ever deploy this stuff there again. (Fine to keep this if you prefer though, whichever's more convenient for now.)

terraform/deployments/datagovuk-infrastructure/security.tf
```diff
@@ -0,0 +1,12 @@
+resource "aws_security_group_rule" "postgres_from_eks_workers" {
+  for_each = merge(data.terraform_remote_state.app_govuk_rds.outputs.sg_rds, {
```
sengi, line 2: I'm confused as to why the for_each here. It looks like we're adding a rule for every database plus the CKAN one, rather than just for CKAN. Is that right?

Did you mean to remove the for_each and write `data.terraform_remote_state.infra_security_groups.outputs.sg_ckan_id` instead of `each.value` on line 10?

terraform/deployments/datagovuk-infrastructure/remote.tf
kentsanggds: ok, I'll drop it

sengi (approved):

PR #914, opened 2023-07-20
Avoid buggy Terraform/AWS error handling on SG rule changes.

sengi:
See commits for details. What a palaver.

Note to self: don't try to use Terraform to change the description field on a security group. In fact, best avoid using the description field altogether because it's inexplicably immutable. (Why, Amazon, whyyyy?)

Tested: un-f***s staging. Happily we never tried to apply the previous version in prod :)


sengi:
(force-pushed to fix some minor inaccuracies in the commit messages)

samsimpson1 (approved):

PR #915, opened 2023-07-20
Fix S3 bucket names in search IAM policies.

sengi:
This makes the stored config match what's currently in prod.


theseanything:
Ah sorry, good catch!

sengi:
No worries!

PR #920, opened 2023-08-07
Add workflow for creating a release

theseanything:
This GitHub Action workflow can be called from our application repositories to create a GitHub Release and add a git tag with the version. The versioning pattern tags each merge commit that has passed its status checks in sequential order with the format `v<number>`. SemVer compatible format was not used as brings no practical benefit yet, however versioning format can always be updated in future.

The workflow assumes a pre-existing version tag in which it calculates the next version number from. The workflow calculates the version number as last version tag number + the number of merge commit. This allows the worklfow to be idepotent and independent of workflow run order (which can be an issue for GitHub Actions).

### DIFF
--- a/.github/workflows/add-version-tag.yml
+++ b/.github/workflows/add-version-tag.yml
@@ -1,0 +1,84 @@
+name: Add version tag
+
+on:
+  workflow_call:
+
+concurrency: ${{ github.workflow }}
+
+jobs:
+  add-version-tag:
+    name: Add version tag
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Find latest version tag
+        id: find-last-tag
+        run: |
+          git fetch --tags --deepen=100
+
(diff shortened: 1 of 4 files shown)


sengi (approved):
Nice!

I haven't 100% convinced myself that two concurrent release actions can't still race for a given version tag, but I'm sure you've thought this through in more detail than I have. Either way I don't think it'll a problem in practice (and we can always add a merge queue if need be).

theseanything:
> I haven't 100% convinced myself that two concurrent release actions can't still race for a given version tag, but I'm sure you've thought this through in more detail than I have. Either way I don't think it'll a problem in practice (and we can always add a merge queue if need be).

If you are talking about two concurrent actions targeting the same commit with same tag, then possibly - but wouldn't be a problem as one would fail, we'd still get the release.

If you're talking about two concurrent actions targeting different commits - they should calculate different tags as the logic increments the version number by the number merge commits away from the last tag. In theory, it's possible to release a later version before an easier version, but they should still be on the correct commits.

PR #927, opened 2023-08-10
Update tflint to v0.47.0

theseanything:


terraform/deployments/cluster-services/.tflint.hcl
```diff
@@ -0,0 +1,13 @@
+config {
+  varfile = [
+    "../variables/common.tfvars",
+    "../variables/test/common.tfvars",
+  ]
+}
+
+plugin "aws" {
+  enabled = true
+  version = "0.25.0"
```
sengi, line 10: optional: I don't suppose there's a way for us to avoid pinning this version number in every module's linter config?

.github/workflows/ci-terraform.yml
```diff
@@ -78,26 +78,14 @@ jobs:
...
         tflint --version
-        tflint --init $TFLINT_PLUGINS
-        for d in terraform/deployments/*; do
-          echo "$d"
-          cd "$d"
-          set +e
-          touch tflint.hcl  # For modules which don't have a tflint.hcl.
-          tflint --config=tflint.hcl -f compact --module . $TFLINT_PLUGINS \
+        tflint --init --recursive 
+        tflint --format compact --module --recursive --force \
```
sengi, line 84: Just trying to understand the implications of [`--force`](https://github.com/terraform-linters/tflint/blob/2d73106/docs/user-guide/config.md#force): does it mean "don't fail on warnings ('issues') but still fail on errors"? Or "never fail even if there are errors"? (I think it's the first one but I'm not sure.)

sengi (approved):
Thanks for fixing this!

.github/workflows/ci-terraform.yml
theseanything: Yeah, I was slight unclear on this too. But tested it with a config error and it exited with status code 1.

```
> tflint --format compact --module --recursive --force \
            --enable-rule=terraform_comment_syntax \
            --enable-rule=terraform_deprecated_index \
            --enable-rule=terraform_required_providers \
            --enable-rule=terraform_standard_module_structure \
            --enable-rule=terraform_typed_variables \
            --enable-rule=terraform_unused_declarations \
            --enable-rule=terraform_unused_required_providers
.tflint.hcl:6:3: error - Unsupported argument. An argument named "asds" is not expected here.

> echo $?
1
```

terraform/deployments/cluster-services/.tflint.hcl
sengi: fr don't feel you need to address this in this PR btw — I appreciate we're already at least 2 yaks deep here 😅

theseanything: Ahah yeah - I might come back to this. There wasn't an obvious way in the docs.

sengi: no probs at all — thanks for looking!

PR #932, opened 2023-08-22
Rename github-cli image to toolbox

theseanything:
This name better reflects its multi purpose nature. This image now contains many general purpose tools for a range of different functions, not longer exclusively for github-cli.


sengi (approved):
Nice — thanks for fixing this! 🙌

PR #934, opened 2023-08-23
Add service account IAM role for mirror sync

theseanything:
This Role provides the govuk mirror sync job permissions to access S3 to backup the mirror.

Tested in Staging.

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/govuk_mirror_sync.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/govuk_mirror_sync.tf
@@ -1,0 +1,52 @@
+module "govuk_mirror_sync_iam_role" {
+  source  = "terraform-aws-modules/iam/aws//modules/iam-eks-role"
+  version = "~> 5.28"
+
+  role_name        = "govuk-mirror-sync"
+  role_description = "Role for govuk-mirror-sync to access S3. Corresponds to govuk-mirror-sync k8s ServiceAccount."
+
+  cluster_service_accounts = {
+    "${local.cluster_name}" = ["apps:govuk-mirror-sync"]
+  }
+
+  role_policy_arns = {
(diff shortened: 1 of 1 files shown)


terraform/deployments/govuk-publishing-infrastructure/govuk_mirror_sync.tf
```diff
@@ -0,0 +1,54 @@
...
+  role_policy_arns = {
+    govuk_mirror_sync_policy = aws_iam_policy.govuk_mirror_sync.arn
+  }
+}
+
+resource "aws_iam_policy" "govuk_mirror_sync" {
+  name        = "govuk-mirror-sync"
+  description = "Allows mirror sync process to access S3."
+
+  policy = jsonencode({
```
sengi, line 21: optional: you can get TF to do somewhat more ahead-of-time checking if you represent the IAM policy using a [data.aws_iam_policy_document](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document) and referring to the its `.json` attribute here. (Also makes the syntax for the policy slightly nicer cos it lets you omit Version and Effect.)

sengi (approved):

terraform/deployments/govuk-publishing-infrastructure/govuk_mirror_sync.tf
theseanything: Ah yes - I always forget to do this!

sengi: Oh no worries! Took me quite a while to figure out that that seems to be the safest way to do it. I should go have another look if there's a linter/fixer for it — glad you've reminded me :D

theseanything: Yeah that'd be ace!

PR #936, opened 2023-08-25
Codeql workflows

wjmurphy:
This PR includes a new workflow for running and advanced setup of Codeql. It is a reusable workflow to be called in a project's CI pipeline. 

The configuration of the job ensures that the scan only searches for high or critical severity errors with a high degree of precision. 

There are also updates to the README to explain how to call in a pipeline. 

Trello: https://trello.com/c/nqaYGM9f/3279-implementation-of-codeql

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@
 
 The govuk-infrastructure repo contains Terraform modules for turning up an EKS Kubernetes cluster for GOV.UK.
 
+Also included are reusable workflows for running the code scanning tools Rubocop, Brakeman and the CodeQL Static Application Security Tests (SAST).
+
 ### What's not in this repo
 
 Helm charts for GOV.UK applications are in [alphagov/govuk-helm-charts](https://github.com/alphagov/govuk-helm-charts).
@@ -20,6 +22,16 @@ To install the [currently-used version of Terraform](terraform/.terraform-versio
 brew install tfenv
 cd terraform/
 tfenv install
+```
+
+To use the code scans reusable workflow, add the following job to the jobs section of your CI workflow:
+
+```yaml
+codeql-sast:
+  name: CodeQL SAST scan
+  uses: alphagov/govuk-infrastructure/.github/workflows/codeql-analysis.yml@main
+  permissions:
+    security-events: write
 ```
 
 ## Working on this repo
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,41 @@
+name: "CodeQL Analysis"
+
+on:
+  workflow_call:
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    continue-on-error: true
+    timeout-minutes: 120
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v2
+      with:
+        config: |
+          # initial configurations for CodeQL should be looking for high precision
+          # and high severity issues to ensure developers aren't compromised by false positives.
+
+          query-filters:
+            - include:
+                # alerts should only be raised for security issues to prevent being distracted with none security issues.
+                "tags contain": security
+                # High precision findings give the best likelihood that issues found are true positives.
+                precision:
+                  - high
+                  - very-high
+                # A security severity score of >=7 is classed as high or critical. This is an additional step to
+                # ensure that developers are not distracted by low severity, none issues.
+                # Explanation of security severity scores: https://docs.github.com/en/code-security/code-scanning/managing-code-scanning-alerts/about-code-scanning-alerts#about-security-severity-levels
+                security-severity: /([7-9]|10)\.(\d)+/
+
+    - name: Autobuild
+      uses: github/codeql-action/autobuild@v2
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v2


README.md
```diff
@@ -22,6 +24,20 @@ cd terraform/
 tfenv install
 ```
 
+To use the CodeQL resuable workflow, add the following job to the jobs section of you CI workflow:
```
ChrisBAshton, line 27:
```suggestion
To use the CodeQL reusable workflow, add the following job to the jobs section of your CI workflow:
```

README.md
```diff
@@ -22,6 +24,20 @@ cd terraform/
...
+
+```
+sast:
```
ChrisBAshton, line 30: Could the choice of name be problematic for if/when acquire Snyk in future? At that point we'll presumably have two sast tasks - we should choose a name that's forwards-compatible.

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,38 @@
...
+      config-path:
+        required: true 
+        type: string 
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ${{ (matrix.language == 'swift' && 'macos-latest') || 'ubuntu-latest' }}
+    continue-on-error: true
+    timeout-minutes: ${{ (matrix.language == 'swift' && 120) || 360 }}
```
ChrisBAshton, line 15: It would be good to link to wherever you may have copied / taken inspiration from, for the code above. We don't (currently) use Swift so I'd be minded to remove these conditionals, but it would still be helpful to know what baseline we're building from here.

.github/sast_configurations/codeql.yml
```diff
@@ -0,0 +1,9 @@
+name: "CodeQL configurations"
+
+query-filters:
+  - include:
+      tags contain: security
```
ChrisBAshton, line 5: Is this key name correct? Having a space in it seems dubious.

.github/sast_configurations/codeql.yml
```diff
@@ -0,0 +1,9 @@
...
+      precision:
+        - high
+        - very-high
+      security-severity: /([7-9]|10)\.(\d)+/
```
ChrisBAshton, line 9: It would be good to write up some public justification for these choices (inline comment, commit message, or PR description would be a start). I think you touch on this in one of the PSR reports but we should be able to easily find the reasoning in GitHub too.

ChrisBAshton (commented):
Thanks Will - a few drive-by comments below.

Did you consider the approach recommended by an [earlier PSR report](https://docs.google.com/document/d/19oLbNBqewzLp1pHYWj2Hya6YYEvpuCRsPr9YIRYUFdc/edit), recommending that CodeQL is enabled through the GitHub API (via something like govuk-saas-config)? It feels like a win to be able to roll out CodeQL on behalf of repos, without the need to copy and paste configs and keep them in sync.

wjmurphy:
Hi @ChrisBAshton, yes, I did consider doing it that way. Unfortunately, we need to use an advanced setup for codeQL. This means we need to write our own workflows to include all the additional configurations we would like. As I understand it, enabling via the API would only permit us to use the default settings. Another thing to consider is that we will likely have many more scans running as part of the CI pipeline. Initially these will grow to be SAST and SCA scans with SNYK. But further down the line, we will have IaC, Container and DAST scans too. So I feel a common approach is probably the best way to go. Otherwise we'll be using the govuk-sass-config app for CodeQL and the govuk-infrastructure repo for most other things.

README.md
wjmurphy: Yes I agree, I think a more descriptive name would be appropriate. I'll change it now.

.github/sast_configurations/codeql.yml
wjmurphy: Yes, this is correct. I have tested it to be sure.

.github/sast_configurations/codeql.yml
```diff
@@ -0,0 +1,16 @@
+name: "CodeQL configurations"
```
sengi, line 1: It seems a little odd to me to configure the filters in this repo, since:
-  it's a safe bet that we're gonna want the ability to tune things per project if this gets widely adopted
- with the currently proposed approach, we'd still have to add a YAML file to every repo that wants to use this anyway, yet that file would just be boilerplate and not contain the config

Are we able to override these settings from the repo that calls the workflow?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,38 @@
...
+
+    - name: Checkout configuration files
+      uses: actions/checkout@v3
+        repository: alphagov/govuk-infrastructure
+        path: configuration_files
+
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v2
+      with:
+        languages: ${{ matrix.language }}
```
sengi, line 29: It doesn't look like you've defined a build matrix. Where is `matrix.language` intended to come from?

.github/sast_configurations/codeql.yml
sengi: I'd recommend quoting it in that case. (someday soon we'll have actionlint, which should help with this sort of thing!)
```suggestion
      "tags contain": security
```

.github/sast_configurations/codeql.yml
```diff
@@ -0,0 +1,16 @@
...
+  - include:
+      # alerts should only be raised for security issues to prevent being distracted with none security issues. 
+      tags contain: security
+      # High precision findings give the best likelihood that issues found are true positives.
+      precision:
+        - high
+        - very-high
+      # A security severity score of >=7 is classed as high or critical. This is an additional step to
+      # ensure that developers are not distracted by low severity, none issues. 
+      security-severity: /([7-9]|10)\.(\d)+/
```
sengi, line 16: Does GitHub really not provide a less awful way to specify the minimum severity? This is pretty nasty and it's very hard to tell that it's correct without actually evaluating it.

sengi:
Forgot to mention:

— It looks like the test runs are failing, for example [the latest one](https://github.com/alphagov/govuk-infrastructure/actions/runs/5977691343) is failing with `You have an error in your yaml syntax on line 23`).
- We already have a reusable workflow for Ruby linting. If we broadened that existing workflow to be "Ruby linting and analysis" then we could add the CodeQL jobs to that workflow and then they'd run automatically for all our Ruby projects with no need for additional configuration/boilerplate/pain.

sengi:
Also, how does this relate to the CodeQL findings that already appear in https://github.com/alphagov/support/security/code-scanning? (Would it filter them? duplicate them? not affect them?)

.github/sast_configurations/codeql.yml
wjmurphy: The idea is that developers can add this to their repo as a reusable workflow and the config can still be maintained in a centralized workflow. If it is centralized like this then we are able to easily edit the filters and have them be applied to all the repos using this workflow at the same time. As we mature the process we will want to introduce the lower severity findings and run some of CodeQL's more advanced scans and this should make it easier to edit the filters for everyone simultaneously. 
Allowing tuning on each repo would mean maturing SAST at different paces on each repo rather than doing it synchronously for the entire organization.

.github/workflows/codeql.yml
wjmurphy: My mistake - I had built a matrix but later removed it. Should have removed this too. You can define the languages your project uses in the matrix and this can speed the scan up. Otherwise, it loops through all the languages supported in CodeQL. Given that this workflow will be used in potentially very many repos, I didn't want to risk the required language being omitted.

.github/sast_configurations/codeql.yml
wjmurphy: This is the only way I can see to filter the security severity level. The docs don't appear to provide a simpler method. 
The benefits of focusing on the high and critical issues first should mean that developers won't have so much of their time wasted dealing with false positives. The number of false positives in a project and therefore the value attained from this approach will vary from one to another. It is generally seen as good practice to start in this way and introduce the lower-severity findings further down the line. 

If all this was coming to was turning on CodeQL for GOV.UK repos, then I'd be inclined to agree with you that we should implement the default settings on each repo. Not because it's the right approach, but because the implementation is much simpler. However, down the road, we are going to have more SAST scans provided by SNYK, plus IaC scans, DAST scans, container scans, and possibly more. They will all require a process for building them into the CI pipelines of all the GOV.UK repos. It seems to me that a common approach should be taken for all these things. Then there are the additional processes around the scans. Somewhere the results can be stored, monitored, and analyzed together. This means the workflows will be extended down the line so that these features can be built in. An advanced implementation of CodeQL would be necessary at this point anyway as none of that can be done with a default implementation. We may as well take that approach now or we'll just have the default settings switched on for a while before we switch it off and reengineer it.

wjmurphy:
> Also, how does this relate to the CodeQL findings that already appear in https://github.com/alphagov/support/security/code-scanning? (Would it filter them? duplicate them? not affect them?)

They remain unaffected. No duplicates are made and no alerts already raised are removed in the evert of increased filtering.

wjmurphy:
> Forgot to mention:
> 
> * It looks like the test runs are failing, for example [the latest one](https://github.com/alphagov/govuk-infrastructure/actions/runs/5977691343) is failing with `You have an error in your yaml syntax on line 23`).
> * We already have a [reusable workflow for Ruby linting](https://github.com/alphagov/govuk-infrastructure/blob/main/.github/workflows/rubocop.yml). If we broadened that existing workflow to be "Ruby linting and analysis" then we could add the CodeQL jobs to that workflow and then they'd run automatically for all our Ruby projects with no need for additional configuration/boilerplate/pain.

I've resolved the syntax issue.

OK brilliant, thank you for point this out to me. I was hoping to come across something like this so we could add the SAST workflows into all apps efficiently. I'll make these changes now as it's a much better way of doing it.

wjmurphy:
> Forgot to mention:
> 
>     * It looks like the test runs are failing, for example [the latest one](https://github.com/alphagov/govuk-infrastructure/actions/runs/5977691343) is failing with `You have an error in your yaml syntax on line 23`).
> 
>     * We already have a [reusable workflow for Ruby linting](https://github.com/alphagov/govuk-infrastructure/blob/main/.github/workflows/rubocop.yml). If we broadened that existing workflow to be "Ruby linting and analysis" then we could add the CodeQL jobs to that workflow and then they'd run automatically for all our Ruby projects with no need for additional configuration/boilerplate/pain.

The failing test issue is now resolved. 

I like the idea of using the Rubocop way of implementation - I've suggested an adapted version detailed in a comment below.

wjmurphy:
Building on the idea to embed this in the Rubocop workflow, I've reengineered it so that there is a code-scanning workflow with Rubocop, Brakeman, and CodeQL nested inside. In the future, we can add additional code scanning tools here and modify them without requiring changes to the workflows on other repos. 
I don't think we should embed the CodeQL job in the Rubocop workflow as I think that could be confusing as it doesn't feel like a natural home for it. 

We would need to amend the repos calling the rubocop job at the moment so instead call the code-scans workflow. However, it would be a one-off change that would give us the ability to make changes to code scanning tasks in the future and have the changes instantly effective across all repos.

.github/sast_configurations/codeql.yml
ChrisBAshton, line 1: Is this separate config file necessary, or can it be merged into `.github/workflows/codeql.yml`?

There's only a handful of properties and I think they can all be copied over into that file, under a `with:` key, within the "Initialize CodeQL" block. I don't see much advantage to keeping it separate, it just makes the CodeQL setup harder to reason about.

README.md
```diff
@@ -22,6 +24,18 @@ cd terraform/
...
 
+To use the code scans reusable workflow, add the following job to the jobs section of you CI workflow:
+
+```
+codeql-sast:
+    name: CodeQL SAST scan
+    uses: alphagov/govuk-infrastructure/.github/workflows/code-scans.yml@main
+    permissions:
+        actions: read
+        contents: read
```
ChrisBAshton, line 35: I think we can drop this line of config, as [it's the default setting](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token), even for organisations that have set reusable workflows to their most restricted setting.

ChrisBAshton (commented):
Hi @wjmurphy , apologies it's taken me until the end of the week to find time to look at this.

I think this is shaping up nicely, but have a few comments below (the main one being the last one, about descoping the grouping of workflows under one name. Let's try and ship just the CodeQL workflow for now, we can revisit grouping later).

Also please squash your commits before next review 🙏 Thank you!

.github/sast_configurations/codeql.yml
wjmurphy: These aren't filtering on the same criteria. Problem severity is looking at whether an alert is a warning, recommendation, or error. We want to look at security severity which is either critical, high, medium, or low. This is based on a numerical value which is why the regex is required.

README.md
```diff
@@ -22,6 +24,16 @@ cd terraform/
...
 ```
 
+To use the code scans reusable workflow, add the following job to the jobs section of your CI workflow:
+
+```
+codeql-sast:
+    name: CodeQL SAST scan
+    uses: alphagov/govuk-infrastructure/.github/workflows/codeql-analysis.yml@main
+    permissions:
+        security-events: write
```
ChrisBAshton, line 34: For consistency with the rest of the YAML:

```suggestion
```yaml
codeql-sast:
  name: CodeQL SAST scan
  uses: alphagov/govuk-infrastructure/.github/workflows/codeql-analysis.yml@main
  permissions:
    security-events: write
```

ChrisBAshton (commented):
LGTM 👍 Just a few things before merge:

1) Nitpick comment below
2) Please rewrite the commit message so it describes the what and why of the change (rather than "# This is a combination of 4 commits")
3) Please update the PR description, which still references the separate config YML file
4) Please confirm if you've tried calling this workflow from another repo yet? (I think you can specify this branch by adding `@codeql-workflows` to the end of the path, as per https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow). If you haven't, I think it would be worth trying that in case there are any config tweaks we should make before merging this 👍

ChrisBAshton (approved):
Thanks Will ⭐ Happy to approve on the basis you'll first test calling the reusable workflow from a different repository.

wjmurphy:
I've tested the workflow from another repo and am happy that it is working as expected

PR #938, opened 2023-08-31
Add generic lint GitHub actions

dnkrj:
# What
- Add actions for `yarn run lint:js`, `yarn run lint:scss` and `yarn run lint:prettier`

# Why
- We're introducing new linting to Whitehall, https://github.com/alphagov/whitehall/pull/8178, based on `govuk-frontend`
- The existing lint actions specify the exact command to run
- These new actions allow applications to configure their own linting whilst still running in parallel

### DIFF
--- a/.github/workflows/lint-js.yml
+++ b/.github/workflows/lint-js.yml
@@ -1,0 +1,15 @@
+name: Run lint:js
+
+jobs:
+  run-lint-js:
+    name: Run lint:js
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Setup Node
+        uses: alphagov/govuk-infrastructure/.github/actions/setup-node@main
+
+      - name: Run lint:js
+        run: yarn run lint:js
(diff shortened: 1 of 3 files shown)


theseanything:
I'm wondering what the value is of having these as reusable workflows vs just writing the individual steps in calling workflow (e.g. CI for Whitehall)? They are essential alias commands to running other tools, so limited in the configuration we can have here.

dnkrj:
> I'm wondering what the value is of having these as reusable workflows vs just writing the individual steps in calling workflow (e.g. CI for Whitehall)? They are essential alias commands to running other tools, so limited in the configuration we can have here.

I'd be open to doing that as well.

My concern would be that it makes it difficult to add or remove steps in future:
Ideally a new step is defined here first and then referenced in the application repo when it is ready to use it (`lint:prettier` in this PR for example).
In the case where we have a single action for Whitehall, the application repo would need to be updated first and then the action here - leading to a situation where CI in Whitehall could be more easily broken by a PR here.

Maybe in practice that's the case anyway and we just don't change these things very often but would be interested in your thoughts.

dnkrj:
Linting actions for Whitehall will be defined in the Whitehall repo

PR #939, opened 2023-08-31
Build wget2 binary and add to toolbox image

theseanything:
The wget2 version available for apt package manager is version 1.99, which contains a bug causing a seg fault when it parses HTML with utf-8 encoding. This builds the latest released version of wget2 version 2.01 and manually adds it to the toolbox image with its library dependencies.

Previously tried to compile a static binary, but a was complex due to the many dependencies and the lack of pre-built static libraries available. Also there are dependencies on glibc, which would require using musl as the standard c library instead.


images/toolbox/Dockerfile
```diff
@@ -1,3 +1,22 @@
+ARG wget2_tarball="wget2-2.1.0"
+
+FROM public.ecr.aws/lts/ubuntu:22.04_stable AS builder
+
```
sengi, line 4: If you set `-e` and whatnot in SHELL like you're already doing for the other stage below, you'll get more-helpful error messages if e.g. `apt-get-update` fails.

```suggestion
SHELL ["/bin/bash", "-euo", "pipefail", "-c"]

```

sengi (approved):
Nice!

images/toolbox/Dockerfile
```diff
@@ -35,8 +55,15 @@ RUN curl -Ssf "https://awscli.amazonaws.com/awscli-exe-linux-$(uname -m).zip" \
     chmod +x "${awscli_install_dir}/aws/dist/aws"
 ENV PATH=$PATH:$awscli_install_dir/aws/dist
 
+ARG wget2_tarball
+COPY --from=builder /tmp/$wget2_tarball /tmp/wget
+WORKDIR /tmp/wget
+RUN make install; \
+    rm -rf /tmp/wget
```
sengi, line 62: I wasn't entirely sure off the top of my head whether this would work (i.e. unlinking the current working directory). Spoiler alert: it does!

sengi:
Thanks for trying to compile it statically btw — I agree that would have been cool had it not turned out to be such an unexpected faff. Also totes agree this is a good call 👍

theseanything:
Opted for govuk-mirror instead.

PR #943, opened 2023-09-12
WIP Spiking adding GitHub teams without importing

ChrisBAshton:


ChrisBAshton:
I do wonder what would happen if we merged and applied, and then did another plan later? 🤔 would Terraform 'know' about these repos now, or would it always consider them to be new resources?

theseanything:
> I do wonder what would happen if we merged and applied, and then did another plan later? 🤔 would Terraform 'know' about these repos now, or would it always consider them to be new resources?

It'd try to create new teams on apply and probably fail as teams with same name exist.

PR #944, opened 2023-09-12
Build a container image for MongoDB 2.4.

sengi:
This is just to give us the option to run our antique version of MongoDB on Kubernetes temporarily until we finish the port of Router API to Postgres. It's helpful to have that flexibility because switching off the old Puppet stack quickly is more important than getting Router API on to Postgres quickly.

The idea would be to run a StatefulSet of these and do a quick stop-and-swap while nobody's publishing anything. ([Ah, the old stop-and-swap](https://github.com/alphagov/govuk-infrastructure/assets/1170635/840cdc5f-49e5-4b30-a377-c6941e5a0b57).)

This thing isn't in the serving path and untrusted input doesn't go anywhere near it, otherwise we wouldn't even be considering this gross workaround. If we do it, it's going to have to come with a hard deadline so that we're not left holding a bag of very stale MongoDB.

Hopefully we can just get the Router API -> Postgres port back on track and not need this, but it's good to have options. Perhaps manifesting the horror of this putative workaround will galvanise the porting effort? We really don't want to find ourselves running this image in 2024. Double plus ungood.

Differences from the [last docker-library/mongo image release](https://github.com/docker-library/mongo/tree/4507e97/2.4):

- image manifest format that works with containerd (so it runs on k8s ≥1.24)
- up-to-date base image, libstdc++, etc. etc.
- we don't care about NUMA (our cluster nodes aren't)
- ditched the entrypoint shell script (we don't need it)
- runs as non-root by default (helps when testing on Docker)

Tested: Router end-to-end tests pass with this image. The failing TF Cloud pre-merge check is unrelated to this PR.

### DIFF
--- a/.github/workflows/build-mongodb-image.yml
+++ b/.github/workflows/build-mongodb-image.yml
@@ -1,0 +1,30 @@
+name: Build and publish MongoDB image to ECR
+
+on:
+  workflow_dispatch:
+    inputs:
+      gitRef:
+        description: 'Commit, tag or branch name to deploy'
+        required: true
+        type: string
+        default: main
+
+  push:
+    branches:
+      - main
+    paths:
+      - images/mongodb/Dockerfile
+  
+  schedule:
+    - cron: '1 22 * * 1'
+
+jobs:
+  build-and-push-image:
+    uses: alphagov/govuk-infrastructure/.github/workflows/build-and-push-image.yml@main
+    with:
+      gitRef: ${{ inputs.gitRef || github.ref }}
+      ecrRepositoryName: mongodb
+      dockerfilePath: images/mongodb/Dockerfile
+    secrets:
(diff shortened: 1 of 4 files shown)


.github/workflows/build-mongodb-image.yml
```diff
@@ -0,0 +1,30 @@
+name: Build and publish MongoDB image to ECR
+
+"on":
```
theseanything, line 3: Huh, your right only just realised that normal this would be "true" in YAML. 

V minor: However GitHub Actions doesn't seem to parse it as such, so for consistency could drop the quotes. 
```suggestion
on:
```

theseanything (approved):

.github/workflows/build-mongodb-image.yml
sengi: Ah fair, I think I only did this to keep the linter quiet in my editor. I should just set it up to use actionlint and then it wouldn't complain. Done :)

PR #949, opened 2023-09-29
Add opt-out of repo permission granting

MuriloDalRi:
Adds a mechanism for opting out of the access-granting, e.g. a GitHub topic govuk-sensitive-access would indicate that the repo needs access to be managed manually, and so not to try granting access to the teams. We anticipate using such a mechanism for github-action-dependabot-auto-merge, where access to these teams should not be granted for security reasons.

https://trello.com/c/tbQjgkwR/3249-automate-granting-team-access-to-govuk-tagged-repos-3

TF plan: https://app.terraform.io/app/govuk/workspaces/GitHub/runs/run-JkLULWD25KJRC3Kg


terraform/deployments/github/main.tf
```diff
@@ -56,11 +56,21 @@ data "github_repository" "govuk" {
...
+  name     = each.key
+}
+
 locals {
   deployable_repos = [
     for r in data.github_repository.govuk : r
     if !r.fork && contains(r.topics, "container")
   ]
+
+  permission_repos = [
```
ChrisBAshton, line 70: I'd suggest a slightly more expressive name, e.g.

```suggestion
  auto_configurable_repos = [
```

Repos marked 'sensitive access' are not auto configurable, they need managing carefully & manually. So I think that makes more sense as a name.

ChrisBAshton (approved):
Nitpick comment below, otherwise LGTM 👍

PR #961, opened 2023-10-12
Add optional runbook to slack alerts

KludgeKML:
In https://github.com/alphagov/govuk-helm-charts/pull/1314 we add a runbook annotation, which points to the developer docs related to fixing the problem showing up in the alert. This change allows the annotation to be added to the slack card for the alert.

I'm not 100% sure about the conditional - at the moment this slack is only used for the one type of alert, which as per the other PR will always have a runbook variable (if not a value), so 1) maybe we remove it? 2) not 100% sure if it's the right format.

### DIFF
--- a/terraform/deployments/cluster-services/templates/alertmanager-config.tpl
+++ b/terraform/deployments/cluster-services/templates/alertmanager-config.tpl
@@ -39,6 +39,8 @@ alertmanager:
 
          *Environment:* ${environment}
 
+         *Runbook:* {{ .CommonAnnotations.runbook_url }}
+
          *Expiring tokens:*
 
          {{ range .Alerts -}}


terraform/deployments/cluster-services/templates/alertmanager-config.tpl
```diff
@@ -39,6 +39,8 @@ alertmanager:
 
          *Environment:* ${environment}
 
+         {{ if contains(keys(.CommonAnnotations), "runbook") }}*Runbook:* {{ .CommonAnnotations.runbook }}{{ end }}
```
theseanything, line 42: Probably don't need to bother checking for the annotation, as we know the alert is going to have runbook
```suggestion
         *Runbook:* {{ .CommonAnnotations.runbook }}
```

theseanything (approved):

terraform/deployments/cluster-services/templates/alertmanager-config.tpl
KludgeKML: (Have added and rebased, because I also updated it to runbook_url)

PR #964, opened 2023-10-18
Update Argo CD chart to 5.46.8.

sengi:
This updates Redis to 7.0.13, which fixes CVE-2022-48174 (not a practical issue for us, but helps to keep the noise down).


sengi:
Gotta get around to making this stuff self-hosting at some point so we don't have to keep patching configs like this :/

samsimpson1 (approved):

PR #966, opened 2023-10-23
SCA scans on PR and push to check for new vulnerabilities

wjmurphy:
A reusable workflow for run dependency review scan on PR and push actions to check for new vulnerabilities being added to the supply chain as part of the change. 

https://trello.com/c/BmET7H34/3326-add-dependabot-scan-to-ci-pipelines

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -1,0 +1,37 @@
+name: "Dependabot Scan"
+
+on:
+  workflow_call:
+
+jobs:
+  dependency-review-pr:
+    if: ${{github.event_name == 'pull_request'}}
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,37 @@
...
+
+jobs:
+  dependency-review-pr:
+    if: ${{github.event_name == 'pull_request'}}
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout repository'
+        uses: actions/checkout@v4
+
+      - name: 'Dependency review'
```
MuriloDalRi, line 14: This name is a bit vague, could we change it to something that makes it more obvious what's being reviewed? Same for the one below.

PR #968, opened 2023-10-24
Tidy up reusable actions and examples.

sengi:
- Document reusable actions in a consistent way.
- Mention in the top-level README that we have a library of reusable actions in this repo.
- Update to checkout v4 and turn off checkout progress logspew.
- Remove unnecessary `${{ }}` syntax from `if` conditions.
- Avoid unnecessarily overriding the name of the `checkout` step.
- Prune some comments that were just stating the obvious.


.github/workflows/brakeman.yml
```diff
@@ -1,3 +1,18 @@
...
+#   push:
+#     branches: [main]
+#
+# jobs:
+#   security-analysis:
+#     name: Security Analysis
+#     uses: alphagov/govuk-infrastructure/.github/workflows/brakeman.yml@main
+
+# REUSABLE WORKFLOW
+# -----------------
```
theseanything, line 15: I'm leaning toward removing these usage docs from all the workflows - I don't think they're helpful as you wouldn't use them verbatim, another thing to keep up to date, and it's not difficult to discern the usage from actual workflow definition.

README.md
```diff
@@ -2,9 +2,13 @@
...
 
-The govuk-infrastructure repo contains Terraform modules for turning up an EKS Kubernetes cluster for GOV.UK.
+The govuk-infrastructure repo contains:
 
-Also included are reusable workflows for running the code scanning tools Rubocop, Brakeman and the CodeQL Static Application Security Tests (SAST).
+- [`terraform/`](terraform/): Terraform modules for turning up an Kubernetes
+  cluster on EKS for GOV.UK.
+- [`.github/`](.github/): GitHub Actions and workflows used by other GOV.UK
+  repos, for example release automation, test runners and security analysis
+  tools.
```
theseanything, line 11: Dunno if we also wanna mention it has the Dockerfile/responsible for the build of the toolbox image: https://github.com/alphagov/govuk-infrastructure/blob/main/images/toolbox/Dockerfile

theseanything (approved):

.github/workflows/brakeman.yml
sengi: Done :)

README.md
sengi: Ooh good catch - thanks!

sengi: Done :)

PR #975, opened 2023-10-27
Add slack alert for mirrors in alertmanager

theseanything:
This configures alertmanager to send a slack message when the MirrorFreshnessAlert is firing.

### DIFF
--- a/terraform/deployments/cluster-services/templates/alertmanager-config.tpl
+++ b/terraform/deployments/cluster-services/templates/alertmanager-config.tpl
@@ -21,6 +21,14 @@ alertmanager:
         group_interval: 12h
         active_time_intervals:
         - inhours
+      - match:
+          alertname: MirrorFreshnessAlert
+        receiver: 'slack-mirror-freshness'
+        repeat_interval: 1d
+        group_wait: 12h
+        group_interval: 12h
+        active_time_intervals:
+        - inhours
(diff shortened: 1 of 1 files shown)


terraform/deployments/cluster-services/templates/alertmanager-config.tpl
```diff
@@ -21,6 +21,14 @@ alertmanager:
...
         active_time_intervals:
         - inhours
+      - match:
+          alertname: MirrorFreshnessAlert
+        receiver: 'slack-mirror-freshness'
+        repeat_interval: 1d
+        group_wait: 12h
+        group_interval: 12h
+        active_time_intervals:
+        - inhours
```
sengi, line 31: optional: we probably wanna do this via a label so we don't have to configure individual alerts in the alertmanager config. No probs if you wanna leave that for another day though. Same for the Slack bit I guess.

sengi (approved):

terraform/deployments/cluster-services/templates/alertmanager-config.tpl
theseanything: Yah agree - I think there's defo a much nicer way to configure these alerts more generally. I'm going to merge this in for now, so we have something as I'm running out of time. But on my to do list to revisit this config.

PR #984, opened 2023-10-31
Initial tfc-bootstrap config

marcpomfret:
tfc-bootstrap is run in local mode the as an initial setup for the tfc-configuration workspace.
tfc-configuration workspace is that is then run in remote mode for the set up the rest of the workspaces as per #990

To run this locally you need to be in the dir and the run 
```
terraform login
terraform init
terraform apply
```

https://trello.com/c/M4fXUvda/3340-update-the-govuk-infrastructure-project-to-use-workspacer-to-set-up-workspaces


sengi (approved):
Neat!

PR #990, opened 2023-11-02
TFC Project and Workspace setup - Cluster Infrastructure

marcpomfret:
This Sets up the project and workspaces in TFC for cluster-infrastructure as adding each environments variables to the relevant workspace.
### Running the tfc-configuration workspace

A token variable must be set in on the variables page of the workspace, this needs to either from the owners group or a group that has all workspace and all project permissions, this is to enable the workspaces to be created/edited

https://developer.hashicorp.com/terraform/cloud-docs/users-teams-organizations/permissions

### Workspace setup
Copy workspace creation module from example in tfc-configuration folder

Edit workspace details in this case to create cluster-infrastructure-production

For initial workspace creation if importing state ensure that `execution_mode = "local"`
and that there is a branch set in the vcs_repo `branch = "marc/cluster-infra-int-state"`

Run the TFC job to create the workspace and add the vars to it

Change the branch used to change between state backends (example #1015)

### State Transfer from S3 
While the state is still pointing at `backend s3` run 
`eval $(gds-cli aws govuk-production-admin -e --art 8h)`
`gds-cli aws govuk-${ENV?}-admin -- terraform init -backend-config=${ENV?}.backend -reconfigure -upgrade`

run terraform plan to see if anything needs to be updated

change state to `cloud` block with relevent tags

 `cloud {
    organization = "govuk"
    workspaces {
      tags = ["cluster-infrastructure", "eks", "aws"]
    }
 }`

`terraform login`
`terraform init`

  If there are multiple workspaces already you get these questions 
    Should Terraform migrate your existing state? - yes
    Terraform Cloud requires all workspaces to be given an explicit name. - use the enviroment here (integration, staging, production)
    Would you like to rename your workspaces? 1
    How would you like to rename your workspaces? cluster-infrastructure-*

  If theres only one workspace
    Should Terraform migrate your existing state? - yes
    Terraform Cloud requires all workspaces to be given an explicit name. cluster-infrastructure-production

Check you can see state in the expected TFC workspace if not you might have workspace in the default project, get rid and try again.

Change execution mode to remote 

Run plan to see if it looks the same as the one ran earlier 

Merge #1015 first and then change all of the branches to main before applying in TFC

https://trello.com/c/M4fXUvda/3340-update-the-govuk-infrastructure-project-to-use-workspacer-to-set-up-workspaces


terraform/deployments/tfc-configuration/variables.tf
```diff
@@ -0,0 +1,289 @@
...
+  default     = {}
+}
+
+variable "envvars_ignore_changes" {
+  type        = map(string)
+  description = "Map of sensitive Environment variables to add to Workspace whereby changes made outside of Terraform will be ignored."
+  default     = {}
+}
+
+# #------------------------------------------------------------------------------
```
sengi, line 216: Seems reasonable to keep these commented-out parts in for illustration 👍

terraform/deployments/tfc-configuration/variables.tf
```diff
@@ -0,0 +1,289 @@
...
+  default     = "Default Workspace"
+}
+
+variable "workspace_desc" {
+  type        = string
+  description = "Description of Workspace."
+  default     = "Created by Terraform Workspacer module."
+}
+
+variable "agent_pool_id" {
```
sengi, line 46: Optional: this and a few others (e.g. file_triggers etc.) look unused — are they? If so, is there value in declaring them preemptively or is it cleaner to leave them out until/unless we find a use for them?

sengi (approved):
Awesome, thanks for persevering with this! I appreciate there's a ton of grunge work behind the scenes to get to this point!

terraform/deployments/tfc-configuration/variables.tf
marcpomfret: Yeah i figured it was easier to see them there rather than having to go to the module to look for them

terraform/deployments/tfc-configuration/variables.tf
marcpomfret: There are a few that were not using yet or are the defaults. i'll go through and comment them out for now and then we can add in as needed. Should be easier to see whats been changed that way.

PR #992, opened 2023-11-06
Replace SSL policy to support Forward Secrecy (FS)

MahmudH:
Tenable scan has highlighted, Forward Secrecy (FS) to be implemented for better encryption security.

Trello: https://trello.com/c/RZI303bT/3336-update-load-balancers-security-policy


sengi (approved):
Thanks!

PR #994, opened 2023-11-07
Remove kube-prometheus-stack helm release

samsimpson1:
Now managed by Argo CD

The resource has already been removed from TF state

https://trello.com/c/pOfajLA2/980-install-kube-prometheus-stack-chart-via-argo-cd-instead-of-terraform


sengi (approved):
Awesome!

PR #996, opened 2023-11-07
Rename Dex AlertManager secrets to conform to standard name

samsimpson1:
https://trello.com/c/pOfajLA2/980-install-kube-prometheus-stack-chart-via-argo-cd-instead-of-terraform


sengi (approved):
Neat

PR #997, opened 2023-11-07
Fix IAM policy for aws-load-balancer-controller.

sengi:
This had gotten out of sync with the version of aws-lb-controller that we're running.

This time, just take Amazon's policy file verbatim instead of formatting it as HCL. This way it'll be easier to diff next time.

The `jsonencode(jsondecode())` is not a no-op; it gives us plan-time syntax checking of the json policy file.

Tested: applied in staging; new ingresses now converge again instead of getting stuck with model build errors.


samsimpson1 (approved):

nimalank7 (approved):
LGTM!

PR #1005, opened 2023-11-09
Add environment banner to ArgoCD

csutter:
Adds some hacky custom CSS that inserts an environment banner into the ArgoCD UI to make it easier to differentiate between environments.


sengi (approved):
Cool!

PR #1008, opened 2023-11-10
updated tagging

martynipratt:
Updated tagging in accordance with https://gds-way.cloudapps.digital/manuals/aws-tagging.html


terraform/deployments/govuk-publishing-infrastructure/app_assets_s3.tf
```diff
@@ -2,8 +2,12 @@ resource "aws_s3_bucket" "app_assets" {
   bucket        = "govuk-app-assets-${var.govuk_environment}"
   force_destroy = var.force_destroy
   tags = {
+    Product     = "GOV.UK"
+    System      = "App assets"
+    Service     = "App assets S3"
```
AgaDufrat, line 7: I think we should keep the Service description a bit more general and not duplicate the name of the system. The guidance says:
> Service: used to describe the function of a particular resource (for example: account management, session storage, front end)

What do you think of describing it as `S3` or `Object storage`

AgaDufrat (commented):
I left a couple of inline comments. However, I’m not sure how best to describe the function of the resources a.k.a what the `Service` values should be. My general thought are:
- We should not be duplicating the System name
- They should be consistent 
- I’m not sure if setting it to the resource name adds any value. For example “Security group” for all “aws_security_group" resources. Maybe “virtual firewall” would be more descriptive. 

All of those questions and thoughts make me wonder what the value of this tag is 🤔 

Can we also update the tags for:
- resource "aws_sns_topic" "ecr_scan_topic"
https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/ecr/ecr-scan.tf#L23
- resource "aws_eip" "eks_nat" (so we don’t forget later) https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/cluster-infrastructure/vpc.tf#L106
- resource "aws_efs_file_system" "clamav-db" https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/cluster-infrastructure/clamav_db_efs.tf#L7
- resource "aws_acm_certificate" "cluster_public" https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/cluster-infrastructure/external_dns.tf#L77

terraform/deployments/govuk-publishing-infrastructure/app_assets_s3.tf
martynipratt: Object storage probably best

AgaDufrat (approved):
LGTM now, thanks. 

Could we also swap the email addresses (sorry, I left the [comment on the Trello card](https://trello.com/c/uL3G06hk/3350-implement-tagging#comment-65562c0b59bc460cbcf3a19f) instead of here) and use: 
- govuk-platform-security-reliability@digital.cabinet-office.gov.uk (instead of reliability-engineering) for PSR owned stuff
- govuk-platform-engineering@digital.cabinet-office.gov.uk for PE owned stuff (instead of govuk-replatforming-team)

I think it's good to merge once TF cloud is happy about the new "govuk_environment" variable.

PR #1020, opened 2023-11-22
TFC AWS permission updates required for cluster infrastructure

marcpomfret:


terraform/deployments/tfc-configuration/aws/main.tf
```diff
@@ -83,10 +88,8 @@ resource "aws_iam_policy" "tfc_policy" {
           "iam:*Login*",
           "iam:*Group*",
           "iam:*PermissionsBoundary*",
-          "iam:*Provider*",
           "iam:*User*",
-          "iam:CreateServiceLinkedRole",
-          "iam:PassRole"
+          "iam:CreateServiceLinkedRole"
```
sengi, line 92: mostly note to self: at some point in the near future we should get to the bottom of exactly when/whether PassRole and CreateServiceLinkedRole are needed 🙃 I'm fine with this for now though. (We'd need to spend a bunch of time painstakingly testing turnup-from-scratch in the test account and seeing which permissions it breaks without — and we've more impactful security improvements to make for now!)

sengi (approved):
Thanks for going through these! LGTM

terraform/deployments/tfc-configuration/aws/main.tf
sengi: I'm pretty sure at least one of them is needed at cluster turnup. CreateServiceLinkedRole also definitely used to be, but I dunno if it still is (last time I ran the cluster turnup TF from scratch was about 3 EKS versions ago so who knows 😄)

PR #1021, opened 2023-11-22
Make CloudFront include Host header in cache key

robinjam:
[Trello card](https://trello.com/c/t8NjzYuC/3355-investigate-why-our-failover-cloudfront-distributions-are-broken)

Our CloudFront failover distributions have publicly accessible domain names (e.g. `xyz.cloudfront.net`). If a user visits any of these URLs, the request is forwarded to origin, which returns an HTTP 404 (because the host is unrecognised), which CloudFront caches.

Prior to this PR, future requests made to the same path but with the correct `Host` header (e.g. `www.gov.uk`) would then serve the cached HTTP 404, instead of the correct response.

This PR configures CloudFront to include the value of the `Host` header in the cache key, so that requests made to the distribution with different `Host` headers are considered different requests, and are cached separately.

### DIFF
--- a/terraform/deployments/cloudfront/main.tf
+++ b/terraform/deployments/cloudfront/main.tf
@@ -153,7 +153,10 @@ resource "aws_cloudfront_cache_policy" "no-cookies" {
       cookie_behavior = "none"
     }
     headers_config {
-      header_behavior = "none"
+      header_behavior = "whitelist"
+      headers {
+        items = ["Host"]
+      }
     }
     query_strings_config {
       query_string_behavior = "all"


sengi (commented):
I'd recommend dropping the WAF rule (as it's at best moot but could mask real security issues in future) and just keeping the fix to the cache key (which is the real fix here).

robinjam:
> I'd recommend dropping the WAF rule (as it's at best moot but could mask real security issues in future) and just keeping the fix to the cache key (which is the real fix here).

That's a good shout, I'll do that tomorrow. Changing the cache policy only occurred to me after I'd already added the WAF rule 🤦‍♂️

sengi (approved):
Thanks! LGTM for the header_config part 🚀

PR #1024, opened 2023-11-28
3350 naming

martynipratt:
Updated AWS naming according to https://gds-way.cloudapps.digital/manuals/aws-naming.html


sengi:
Please don't merge this as-is; contains merge markers, lots of fixup commits etc.

The naming convention itself also hasn't been agreed.

PR #1033, opened 2023-12-04
Add firewall rule for Licensify DocDB traffic.

sengi:
Now that we're starting to run Licensify (aka GOV.UK Licensing) in Kubernetes, we need IP connectivity between the cluster nodes and the DocumentDB instances that host the Licensify database.


dj-maisy (approved):
Perfect! 🚀

PR #1037, opened 2023-12-07
Generalise the setup-postgres reusable Action.

sengi:
Let the caller specify the parameters to the Postgres image if they desire, while preserving the existing defaults and behaviour.

The only functional change here is that we automatically track the current minor release of Postgres 13 instead of pinning an obsolete version.

The immediate reason for this is to work around a test which wants a specific Postgres user to exist, but it also enables a bunch of other use cases:

 - using a superuser username/password other than root/root
 - running a Postgres version other than 13
 - more than one Postgres container to coexist in the same test
 - changing the initial database name, listen port, etc.

### DIFF
--- a/.github/actions/setup-postgres/action.yml
+++ b/.github/actions/setup-postgres/action.yml
@@ -1,36 +1,53 @@
-name: 'Setup Postgres'
-description: 'Create a new Postgres database'
+# See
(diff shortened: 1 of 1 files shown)


theseanything (approved):

theseanything:
Thanks for fixing this!

KludgeKML (approved):
Nice!

PR #1044, opened 2023-12-19
Replace SSL policy to support Forward Secrecy (FS)

MahmudH:
We have added a changed SSL policy to support FS in signon and content-publisher. This PR will add it to www as our Tenable scans alerted that FS is not available for GOV.UK

### DIFF
--- a/terraform/deployments/cluster-services/main.tf
+++ b/terraform/deployments/cluster-services/main.tf
@@ -61,7 +61,7 @@ locals {
     "alb.ingress.kubernetes.io/target-type"  = "ip"
     "alb.ingress.kubernetes.io/listen-ports" = jsonencode([{ HTTP = 80 }, { HTTPS = 443 }])
     "alb.ingress.kubernetes.io/ssl-redirect" = "443"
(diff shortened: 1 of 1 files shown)


terraform/deployments/cluster-services/main.tf
```diff
@@ -61,6 +61,6 @@ locals {
     "alb.ingress.kubernetes.io/target-type"  = "ip"
     "alb.ingress.kubernetes.io/listen-ports" = jsonencode([{ HTTP = 80 }, { HTTPS = 443 }])
     "alb.ingress.kubernetes.io/ssl-redirect" = "443"
-    "alb.ingress.kubernetes.io/ssl-policy"   = "ELBSecurityPolicy-TLS-1-2-Ext-2018-06" # No TLS 1.0 or 1.1.
+    "alb.ingress.kubernetes.io/ssl-policy"   = "ELBSecurityPolicy-TLS13-1-2-2021-06" # TLS 1.3, backward compatible with 1.2
```
sengi, line 64: The comment's no longer necessary here.
```suggestion
    "alb.ingress.kubernetes.io/ssl-policy"   = "ELBSecurityPolicy-TLS13-1-2-2021-06"
```

sengi (approved):
Thanks for updating this.

sengi:
BTW cluster-services isn't in TFC yet so you can ignore that failing check.

PR #1045, opened 2023-12-19
Route Staging EKS onto the new NAT Gateways

dj-maisy:
## What?
As setting up a route specifically for Civica (for Licensify) would be too cumbersome, it might just be easier to route all of our traffic immediately through the new NAT Gateways. So let's give that a try!

## TF Output
...can be viewed here (if you have permission):  
https://app.terraform.io/app/govuk/workspaces/cluster-infrastructure-staging/runs/run-kxHrtVBKALuKHihc


sengi (approved):
I think this is the right approach — easy to revert and avoids the complexity of having a weird special-case route.

sengi:
`s/sl.m/route/`?

dj-maisy:
> `s/sl.m/route/`?

Not sure I follow!

sengi:
Just a bit of an odd choice of words! :)

PR #1046, opened 2023-12-20
Workflow users can provide additional image tags

mike29736:
https://trello.com/c/jDSDGNyn/814-measure-and-record-our-publishing-latency-sli

It takes a space-separated string.

This change is motivated by a desire to deploy separate images to each environment but for a non-app project (i.e. doesn't use Argo workflows or `image-tags`).

We only need a single additional image tag on our specific project, but to me, the future proofness of this option felt like it outweighed the additional complexity.

(I've tested this with an empty value, a single tag and multiple tags, and it's shown up with/without them as expected in ECR for all of them.)

### DIFF
--- a/.github/workflows/build-and-push-image.yml
+++ b/.github/workflows/build-and-push-image.yml
@@ -18,6 +18,9 @@ on:
         required: false
         type: string
         default: ${{ github.sha }}
+      additionalImageTags:
+        required: false
+        type: string
     secrets:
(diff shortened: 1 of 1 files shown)


mike29736:
This change is no longer needed for the project that motivated it (https://github.com/alphagov/govuk-helm-charts/pull/1654 gave the project access to `image-tags`)

PR #1047, opened 2023-12-21
Fix the config of the callable SNYK security analysis workflow

wjmurphy:
Fixing the config of the callable workflow as instructed in the docs:

https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow


theseanything (changes requested):
Can you update the commit message to can the description of the change?

AgaDufrat:
Superseded by https://github.com/alphagov/govuk-infrastructure/pull/1049

PR #1054, opened 2024-01-02
Fix don't use licensify EKS routes in integration

theseanything:
This fixes the AWS Route tables to only route to the licensify if the present. We don't have the licensify NAT gateway in intgration, so these routes should still go to the existing NAT.


dj-maisy (approved):
🚀

PR #1058, opened 2024-01-09
Add Clamav image that can run unprivileged

theseanything:
The current official Clamav image cannot be run unprivileged by default. This adds a script and changes the file permissions, so that it can run as non-root. At time of writing, there is a pending PR to add this functionality to the offical image: https://github.com/Cisco-Talos/clamav-docker/pull/33. 

Once this is available we can switch to the offical image.


samsimpson1 (approved):

images/clamav/scripts/unprivileged-entrypoint.sh
sengi, line 1: Do we need any licence attributions on this, or is it original? (beware https://github.com/Cisco-Talos/clamav/tree/main is GPL)

theseanything: I've added one below

sengi (approved):

PR #1078, opened 2024-01-23
Give licensingusers read-only access the cluster

theseanything:
This allows us to limit the permissions for those just managing licensing applications.


sengi (approved):
SGTM for now — presumably we can limit this to just the licensing namespace at some point though?

theseanything:
> SGTM for now — presumably we can limit this to just the licensing namespace at some point though?

Absolutely!

PR #1082, opened 2024-01-23
Create seperate restricted Role for licensingusers

theseanything:
This restrict access of licensingusers to only the licensify namespace and to only perform a limit set of actions.


sengi (approved):
🙌

Issue #1064, opened 2024-01-15
Move the rest of govuk-infrastructure into Terraform Cloud


sengi:

samsimpson1:
All modules have been migrated to TFC, but the documentation is now very out of date and needs rewriting

PR #1087, opened 2024-01-24
Migrate govuk-publishing-infrastructure into TFC

samsimpson1:
#1064

### DIFF
--- a/terraform/deployments/tfc-configuration/govuk-publishing-infrastructure.tf
+++ b/terraform/deployments/tfc-configuration/govuk-publishing-infrastructure.tf
@@ -1,0 +1,31 @@
+module "govuk-publishing-infrastructure-integration" {
+  source  = "alexbasista/workspacer/tfe"
+  version = "0.9.0"
+
+  organization      = var.organization
+  workspace_name    = "govuk-publishing-infrastructure-integration"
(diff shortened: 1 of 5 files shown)


terraform/deployments/tfc-configuration/govuk-publishing-infrastructure.tf
```diff
@@ -0,0 +1,94 @@
...
+
+  project_name = "govuk-infrastructure"
+  vcs_repo = {
+    identifier     = "alphagov/govuk-infrastructure"
+    branch         = "main"
+    oauth_token_id = data.tfe_oauth_client.github.oauth_token_id
+  }
+
+  team_access = {
+    "GOV.UK Non-Production" = "write"
```
sengi, line 54: I think we usually treat staging the same as prod from an access controls perspective. Let's go with just the production group here for now and see if it leads to any friction.
```suggestion
```

(apologies if I'd not spotted this on previous PRs!)

sengi (approved):
Awesome!

terraform/deployments/tfc-configuration/govuk-publishing-infrastructure.tf
samsimpson1: Ah good spot, I'll update the rest of the TF projects too

PR #1094, opened 2024-01-25
Remove terraform-lock module

samsimpson1:
This is no longer necessary with TFC. Resources have already been destroyed

#1064


sengi (approved):
Thanks!

Issue #1011, opened 2023-11-13
Prefer `aws_iam_policy_document` to `policy = jsonencode(...)`.


sengi:
[`aws_iam_policy_document`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document) provides better ahead-of-time checking and helps avoid some [IAM gotchas](https://securitylabs.datadoghq.com/articles/exploring-github-to-aws-keyless-authentication-flaws/#case-study-a-vulnerable-iam-role-in-an-uk-governments-aws-account). It also makes policies easier to read by cutting out boilerplate and simplifying the syntax.

Here's one I made earlier: https://github.com/alphagov/govuk-infrastructure/commit/42402a00

Grep for cases that need fixing:

```sh
git grep 'policy *= *jsonencode'
```

Extra credit: implement a lint rule for this.
Extra extra credit: implement a fixer for this!

PR #1098, opened 2024-01-29
1011 - Use aws_iam_policy_document instead of jsonencode

nimalank7:
Description:
- PR aims to fix https://github.com/alphagov/govuk-infrastructure/issues/1011
- `aws_ecr_lifecycle_policy` in https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/ecr/main.tf#L242 couldn't be refactored as `aws_iam_policy` doesn't have a corresponding `rules` section


sengi (approved):
Wow, that's quite a big PR 😅 Thanks for improving all these!

All null-diff apart from the ECR one, which looks like it's just list/map ordering changes 👍

PR #1101, opened 2024-01-30
Rename Licensify NATs

theseanything:
Having these NATs named in reference to Licensify is confusing, as they are used for all traffic.

### DIFF
--- a/terraform/deployments/cluster-infrastructure/vpc.tf
+++ b/terraform/deployments/cluster-infrastructure/vpc.tf
@@ -99,12 +99,6 @@ resource "aws_eip" "eks_nat" {
   # TODO: depends_on = [aws_internet_gateway.gw] once we've imported the IGW from govuk-aws.
 }
 
-import {
-  for_each = var.eks_licensify_gateways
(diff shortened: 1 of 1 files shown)


terraform/deployments/cluster-infrastructure/vpc.tf
```diff
@@ -99,14 +99,8 @@ resource "aws_eip" "eks_nat" {
   # TODO: depends_on = [aws_internet_gateway.gw] once we've imported the IGW from govuk-aws.
 }
 
-import {
-  for_each = var.eks_licensify_gateways
-  to       = aws_eip.eks_nat[each.key]
-  id       = each.value.eip
-}
-
 resource "aws_nat_gateway" "eks" {
```
samsimpson1, line 102: I'm pretty sure the resource name will need updating to `eks_licensify` for the moved block below to work

theseanything: Ah yeah, I fluffed that names.. Should be fixed now and the terraform plan looks like what we want :)

PR #1102, opened 2024-01-30
Make GitHub CI user token available to Dependabot

brucebolt:
As of version 4 of [`delete-artifact`](https://github.com/GeekyEggo/delete-artifact), a GitHub API token is required to remove artifacts, which we are doing as part of the CI pipeline.

This secret is included for access by GitHub Actions triggered events, but is not available to Dependabot.

Therefore this step fails on PRs raised by dependabot.

This makes the secret accessible to dependabot.

[Trello card](https://trello.com/c/ngFoCxYX)


brucebolt:
Thanks @sengi.  It turns out the artifact from this pipeline is tiny (1.15KB) and we have our GitHub org set up to only retain them for 90 days, so I'm just going to remove this step as it feels redundant (and GitHub are already clearing these up anyway).

PR #1106, opened 2024-01-31
Update applying terraform documentation for Terraform Cloud

### LINKED ISSUE #1064: Move the rest of govuk-infrastructure into Terraform Cloud

### LINKED ISSUE #1090: Fix cloudfront module always wanting to make changes
The cloudfront module always wants to make the following change to both assets/www cloudfront distributions, even if nothing has changed:

```
default_cache_behavior { 
  default_ttl: 0 -> 86400
  max_ttl: 0 ->  31536000
}
```

samsimpson1:
Also make some small fixes to terraform wanting to always make changes

Resolves #1064
Resolves #1090

### DIFF
--- a/terraform/docs/applying-terraform.md
+++ b/terraform/docs/applying-terraform.md
@@ -23,7 +23,7 @@ start for your branch.
 When turning up from scratch, deploy the root modules in this order:
 
 1. [`tfc-bootstrap`](../deployments/tfc-bootstrap): bootstraps Terraform Cloud and creates `tfc-configuration`
-workspace which is used to manage the other Terraform module workspaces.
+workspace which manages the other Terraform module workspaces.
 2. [`tfc-configuration`](../deployments/tfc-configuration): creates TFC workspaces for each terraform module and environment.
 3. [`ecr`](../deployments/ecr) (test and production accounts only): creates the ECR container registry from
    which the cluster pull container images. There is a single registry for all
--- a/terraform/docs/applying-terraform.md
+++ b/terraform/docs/applying-terraform.md
@@ -31,10 +31,9 @@ workspace which manages the other Terraform module workspaces.
    having to copy images between registries), so this module is not deployed
    per-environment.
 4. [`cluster-infrastructure`](../deployments/cluster-infrastructure): creates the AWS resources for the cluster.
-5. Delete the `aws-auth` configmap by running `gds aws govuk-${ENV?}-admin -- aws eks update-kubeconfig --name govuk && kubectl -n kube-system delete cm aws-auth`. This is a workaround for the problem that one of the AWS-managed EKS addons creates a default aws-auth configmap which then either needs to be imported into Terraform or deleted.
-6. [`govuk-publishing-infrastructure`](../deployments/govuk-publishing-infrastructure): creates AWS resources specific to the GOV.UK apps where we are not yet
+5. [`govuk-publishing-infrastructure`](../deployments/govuk-publishing-infrastructure): creates AWS resources specific to the GOV.UK apps where we are not yet
 able to manage those resources via Kubernetes.
-7. [`cluster-services`](../deployments/cluster-services): deploys the base services into the cluster.
+6. [`cluster-services`](../deployments/cluster-services): deploys the base services into the cluster.
 
 ### Bootstrapping Terraform Cloud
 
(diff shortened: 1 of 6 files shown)


terraform/docs/applying-terraform.md
```diff
@@ -7,60 +7,49 @@ The EKS cluster is deployed via Terraform in two stages. See [adr-3] for backgro
...
+start for your branch.
 
 When turning up from scratch, deploy the root modules in this order:
 
-1. [`terraform-lock`](../deployments/terraform-lock): creates the
-   DynamoDB table which Terraform uses to control concurrent access to the
-   state files in S3.
-1. [`ecr`](../deployments/ecr) (test and production accounts only): creates the ECR container registry from
+1. [`tfc-bootstrap`](../deployments/tfc-bootstrap): bootstraps Terraform Cloud and creates `tfc-configuration`
+workspace which is used to manage the other Terraform module workspaces.
```
sengi, line 26: optional/trivial: passive voice doesn't add much here so prefer active (debatable rule of thumb but v common in tech writing style guides including [ours](https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style))
```suggestion
workspace which manages the other Terraform module workspaces.
```

terraform/docs/applying-terraform.md
```diff
@@ -7,60 +7,49 @@ The EKS cluster is deployed via Terraform in two stages. See [adr-3] for backgro
...
    having to copy images between registries), so this module is not deployed
    per-environment.
-1. [`cluster-infrastructure`](../deployments/cluster-infrastructure): creates the AWS resources for the cluster.
-1. Delete the `aws-auth` configmap by running `gds aws govuk-${ENV?}-admin -- aws eks update-kubeconfig --name govuk && kubectl -n kube-system delete cm aws-auth`. This is a workaround for the problem that one of the AWS-managed EKS addons creates a default aws-auth configmap which then either needs to be imported into Terraform or deleted.
-1. [`govuk-publishing-infrastructure`](../deployments/govuk-publishing-infrastructure): creates AWS resources specific to the GOV.UK apps where we are not yet
+4. [`cluster-infrastructure`](../deployments/cluster-infrastructure): creates the AWS resources for the cluster.
+5. Delete the `aws-auth` configmap by running `gds aws govuk-${ENV?}-admin -- aws eks update-kubeconfig --name govuk && kubectl -n kube-system delete cm aws-auth`. This is a workaround for the problem that one of the AWS-managed EKS addons creates a default aws-auth configmap which then either needs to be imported into Terraform or deleted.
```
sengi, line 34: :sadpanda: Off-topic but I wonder if recent versions of TF give us the means to adopt the existing resource by name without a separate import step 🤔 (don't worry about it for now, just thinking out loud here)

terraform/docs/example-tfc-role-policy.json
sengi, line 1: optional: is it worth switching these to [aws_iam_policy_document](https://github.com/alphagov/govuk-infrastructure/issues/1011) syntax?

sengi (approved):
Awesome, thanks for fixing these docs!

terraform/docs/example-tfc-role-policy.json
sengi: (counterpoint: there are def some cases where we're just taking AWS-provided policy documents verbatim, so makes sense to keep those ones in the same format that AWS provides)

sengi: I think I just talked myself out of this one haha. Could maybe mention that if it's a policy that we maintain ourselves (as opposed to copied verbatim from Amazon/vendor docs) then we recommend using aws_iam_policy_document for better validation — up to you though

samsimpson1: These policies aren't taken from the AWS docs, they're pretty much what we're using at the moment. Unfortunately they can't really be managed by TF because it needs to be in place before TF can do anything on the AWS accounts

terraform/docs/applying-terraform.md
samsimpson1: Hmm, I wonder if using an `import` block here would work, since they're idempotent. Something like:

```hcl
import {
  id = "kube-system/aws-auth"
  to = kubernetes_config_map.aws_auth
}
```

samsimpson1: TF seems to be happy with the import block, so I've added it and removed that step from the docs

sengi: Awesome!

PR #1109, opened 2024-02-01
Add IAM role for Github Actions ECR push workflows

samsimpson1:
This essentially duplicates the IAM user defined [here](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/ecr/main.tf#L86), which will be removed once all workflows are migrated to using OIDC

### DIFF
--- a/terraform/deployments/ecr/gha-iam-role.tf
+++ b/terraform/deployments/ecr/gha-iam-role.tf
@@ -1,0 +1,64 @@
+locals {
+  github_oidc_arn = "arn:aws:iam::${data.aws_caller_identity.current.account_id}:oidc-provider/token.actions.githubusercontent.com"
+}
+
+data "aws_iam_openid_connect_provider" "github_oidc" {
+  arn = local.github_oidc_arn
+}
+
+data "aws_iam_policy_document" "ecr_role_permissions" {
+  statement {
+    actions = [
+      "ecr:GetDownloadUrlForLayer",
+      "ecr:BatchGetImage",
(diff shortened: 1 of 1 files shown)


terraform/deployments/ecr/gha-iam-role.tf
```diff
@@ -0,0 +1,64 @@
...
+    condition {
+      test     = "StringEquals"
+      variable = "token.actions.githubusercontent.com:aud"
+      values   = data.aws_iam_openid_connect_provider.github_oidc.client_id_list
+    }
+    condition {
+      test     = "StringLike"
+      variable = "token.actions.githubusercontent.com:sub"
+      values   = ["repo:alphagov/*"]
+    }
```
sengi, line 50: I always have to check the docs for these — IIUC it's an AND so lgtm 👍

terraform/deployments/ecr/gha-iam-role.tf
```diff
@@ -0,0 +1,64 @@
...
+
+data "aws_iam_policy_document" "ecr_role_permissions" {
+  statement {
+    actions = [
+      "ecr:GetDownloadUrlForLayer",
+      "ecr:BatchGetImage",
+      "ecr:BatchCheckLayerAvailability",
+      "ecr:DescribeImages",
+      "ecr:PutImage",
+      "ecr:InitiateLayerUpload",
```
sengi, line 17: 🐔🥚🚀

sengi (approved):
Thanks!

Issue #1113, opened 2024-02-01
Use OIDC connection for Github Actions to authenticate with AWS


samsimpson1:
We're using long-lived credentials at the moment, we should move to using an IAM role and the Github Actions -> AWS OIDC connection

PR #1115, opened 2024-02-02
Update build-and-push-image.yml workflow to use OIDC connection

samsimpson1:
Tested with whitehall and images in govuk-infrastructure 

#1113


sengi (approved):
🎉

PR #1116, opened 2024-02-02
Add tags to AWS resources

MuriloDalRi:
https://trello.com/c/xnCICnzN/3369-define-owners-of-aws-resources-3


terraform/deployments/cloudfront/main.tf
```diff
@@ -301,6 +301,13 @@ resource "aws_wafv2_web_acl" "cdn_poc_govuk" {
     cloudwatch_metrics_enabled = true
     metric_name                = "cdn_poc_govuk"
   }
+
+  tags = {
+    Product     = "GOV.UK"
+    System      = "Cloudfront"
+    Environment = "${var.govuk_environment}"
```
sengi, line 308: These all contain unnecessary substitutions / format strings. Please fix.

terraform/deployments/cloudfront/main.tf
```diff
@@ -301,6 +301,13 @@ resource "aws_wafv2_web_acl" "cdn_poc_govuk" {
     cloudwatch_metrics_enabled = true
     metric_name                = "cdn_poc_govuk"
   }
+
+  tags = {
```
sengi, line 305: Please use [provider default tags](https://developer.hashicorp.com/terraform/tutorials/aws/aws-default-tags) instead of tagging every resource individually.

sengi (changes requested):

samsimpson1:
We now have default tags defined in the AWS providers so most of these should already be set: https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/cluster-infrastructure/main.tf#L35-L43

MuriloDalRi:
Thanks @sengi and @samsimpson1 , I've removed most of the tags now

terraform/deployments/govuk-publishing-infrastructure/locations_api_s3.tf
```diff
@@ -2,8 +2,11 @@ resource "aws_s3_bucket" "location_api_import_csvs" {
   bucket        = "govuk-${var.govuk_environment}-locations-api-import-csvs"
   force_destroy = var.force_destroy
   tags = {
-    System = "Locations API"
-    Name   = "CSVs for importing postcode information into Locations API"
+    System      = "Locations API"
+    Name        = "CSVs for importing postcode information into Locations API"
+    Product     = "GOV.UK"
+    Environment = var.govuk_environment
```
sengi, line 8: These should be inherited, no? Same for all the others below.
```suggestion
```

MuriloDalRi: Ah yes, Monday!

sengi: Haha no worries :)

sengi (approved):
Cool, thanks for fixing.

PR #1124, opened 2024-02-13
Update to Argo CD 2.10.

sengi:
The 6.0 Helm chart has some breaking changes, so this involves some updates to the config.

https://github.com/argoproj/argo-helm/tree/main/charts/argo-cd#600


theseanything (approved):
Thanks!

PR #1142, opened 2024-02-27
Enable caching for image builds

theseanything:
This replaces our custom implementation to build and push images with [Docker's own GitHub Action](https://github.com/docker/build-push-action). From this we gain ability to cache layers using GitHub's cache and to decrease workflow runtime. This change removes the "additionalBuildArgs" input, which was only being used by Licensify. This is replaces with "buildArgs" which is more specific in its usage.

We no longer need to check if the existing image is already built as caching prevent us re-building an image.


sengi (approved):
Awesome!

PR #1157, opened 2024-02-29
Remove RBAC resources for argo workflows

theseanything:
These resources aren't being used. The default Role and Service Accounts are managed within cluster-services.


sengi (approved):
Great find! Thanks so much for cleaning these up!

Issue #1108, opened 2024-01-31
Snyk reusable workflow downloads + installs Snyk binaries on every run.


sengi:
This burns resources and runner time (which we pay for when we go over our included quota, which we've been doing lately).

It also occasionally causes flakes (unsurprising — if you download a 50M file enough times you're bound to see it fail once in a while), which is annoying and an unnecessary test dependency.

We need to be [caching](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows) this stuff so that we're not being wasteful with runner minutes (and other resources).

<details>
<summary>[Example](https://github.com/alphagov/signon/actions/runs/7729965195/job/21074309271#step:3:17) where the download failed</summary>

```
Run /home/runner/work/_actions/snyk/actions/master/setup/setup_snyk.sh latest Linux
  /home/runner/work/_actions/snyk/actions/master/setup/setup_snyk.sh latest Linux
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
Installing the latest version of Snyk on Linux
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
 83 44.2M   83 37.0M    0     0  36.1M      0  0:00:01  0:00:01 --:--:-- 36.1M
 97 44.2M   97 43.1M    0     0  40.0M      0  0:00:01  0:00:01 --:--:-- 39.9M
curl: (56) OpenSSL SSL_read: Connection reset by peer, errno 104
Error: Process completed with exit code 56.
```
</details>

PR #1167, opened 2024-03-07
Make Snyk SCA scan failures block builds

robinjam:
[Trello card](https://trello.com/c/I5jfSH7Z/3445-3-ensure-failed-sca-scans-block-merging)

This is a requirement before we can implement [RFC-167][], "Allow automatic patching of all dependencies".

If this change causes problems on any our repos, we could temporarily patch those repos (and those repos only) to refer to the previous version of this workflow instead of `@main`. For example:

```yaml
jobs:
  snyk-security:
    name: SNYK security analysis
    uses: alphagov/govuk-infrastructure/.github/workflows/snyk-security.yml@533143ddd823518b97165a4a80c2995e9ffe980a
    secrets: inherit
```

[RFC-167]: https://github.com/alphagov/govuk-rfcs/blob/main/rfc-167-auto-patch-dependencies.md#sufficient-security-scanning

### DIFF
--- a/.github/workflows/snyk-security.yml
+++ b/.github/workflows/snyk-security.yml
@@ -25,7 +25,6 @@ jobs:
       - uses: actions/checkout@v4
       - uses: snyk/actions/setup@master
       - name: Snyk Supply Chain Test
-        continue-on-error: true
         run: |
           snyk auth ${{ secrets.SNYK_TOKEN }}
           snyk test --severity-threshold=high --all-projects --detection-depth=1


sengi:
Just fyi these checks still [flake a bit](https://github.com/alphagov/govuk-infrastructure/issues/1108). Dunno if that's gonna be a problem (happy to try it and find out — we can always revert).

ChrisBAshton (approved):
LGTM 👍 thanks both. Let's keep an eye, we can always revert (and in the meantime should probably get a PSR card written to look at caching dependencies)

sengi:
Last I looked, the flakiness was within the third-party action.

robinjam:
Going to merge this for now, if it's flaky then it's easy enough to re-run failed jobs.

I don't know if it's possible to get GH to cache the Snyk binary without forking the `snyk/actions/setup` action, but I'll have a quick look and maybe write up a card for PSR to do a spike if it looks doable.

PR #1170, opened 2024-03-11
Update staging k8s cluster version to 1.29

nimalank7:
Description:
- Upgrade to version 1.29 has already happened in EKS so this resolves the Terraform diff


sengi (approved):

sengi:
Since they're all already updated, you could do all three of these in one PR tbh. No problem though 😃

PR #1171, opened 2024-03-11
Update prod k8s cluster version to 1.29

nimalank7:
Description:
- Upgrade to version 1.29 has already happened in EKS so this resolves the Terraform diff


sengi (approved):
Yay!

PR #1173, opened 2024-03-13
Add reusable workflow to auto-release rubygems

robinjam:
[Trello card](https://trello.com/c/bYVP288R/3447-5-automate-release-of-new-gem-versions)

This reusable workflow, intended to be added to Ruby gem repos on a weekly schedule, will automatically raise a PR to perform a patch-level version bump if (and only if) the gem has unreleased changes, and all of those changes were authored by Dependabot. After a developer approves and merges that PR, the [publish-rubygem workflow](https://github.com/alphagov/govuk-infrastructure/blob/main/.github/workflows/publish-rubygem.yml) will automatically publish the next release to Rubygems.

This has been [tested on a dummy repo](https://github.com/robinjam/auto-release-test/pull/3) outside of alphagov. My intention is to first roll it out to PSR-owned repos, so that we can debug any teething issues before rolling it out more widely.

### DIFF
--- a/.github/workflows/autorelease-rubygem.yml
+++ b/.github/workflows/autorelease-rubygem.yml
@@ -8,10 +8,9 @@
 #
 # jobs:
 #   autorelease:
-#     permissions:
-#       contents: write
-#       pull-requests: write
 #     uses: alphagov/govuk-infrastructure/.github/workflows/autorelease-rubygem.yml@main
+#     secrets:
+#       GH_TOKEN: ${{ secrets.GOVUK_CI_GITHUB_API_TOKEN }}
 
 name: Auto-release Rubygem
 
@@ -22,19 +21,20 @@ on:
         required: false
         type: string
         default: ${{ github.event.repository.name }}
+    secrets:
+      GH_TOKEN:
+        required: true
 
 jobs:
   release:
     name: Release Gem
     runs-on: ubuntu-latest
-    permissions:
-      contents: write
-      pull-requests: write
     steps:
     - uses: actions/checkout@v4
       with:
         fetch-depth: 0
         show-progress: false
+        token: ${{ secrets.GH_TOKEN }}
     - uses: ruby/setup-ruby@v1
       with:
         rubygems: latest
@@ -46,7 +46,7 @@ jobs:
       name: "Auto-release rubygem"
       env:
         GEM_NAME: ${{ inputs.gem_name }}
-        GH_TOKEN: ${{ github.token }}
+        GH_TOKEN: ${{ secrets.GH_TOKEN }}
       run: |
         set -euo pipefail
 
--- a/.github/workflows/autorelease-rubygem.yml
+++ b/.github/workflows/autorelease-rubygem.yml
@@ -1,0 +1,96 @@
+# USAGE
+# -----
+#
+# on:
+#   workflow_dispatch: {}
+#   schedule:
+#     - cron: '00 13 * * 2'
+#
+# jobs:
+#   autorelease:
+#     permissions:
+#       contents: write
+#       pull-requests: write
+#     uses: alphagov/govuk-infrastructure/.github/workflows/autorelease-rubygem.yml@main
+
+name: Auto-release Rubygem
+
+on:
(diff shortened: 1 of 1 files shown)


.github/workflows/autorelease-rubygem.yml
```diff
@@ -0,0 +1,96 @@
...
+        fi
+
+        if [ "$(git log v$CURR_VERSION..main --no-merges --format=format:%an | sort | uniq)" != "dependabot[bot]" ]; then
+          echo "Gem has no unreleased changes, or those changes include commits from users other than Dependabot. Exiting."
+          exit 0
+        fi
+
+        find . -type f -iname version.rb -exec sed -i "s/$CURR_VERSION/$NEXT_VERSION/g" {} \;
+
+        printf "# v$NEXT_VERSION\n\n* Update dependencies\n\n" | cat - CHANGELOG.md > NEW_CHANGELOG.md
```
AgaDufrat, line 81: I think we don't tend to put the `v` in the headers.

```suggestion
        printf "# $NEXT_VERSION\n\n* Update dependencies\n\n" | cat - CHANGELOG.md > NEW_CHANGELOG.md
```

.github/workflows/autorelease-rubygem.yml
```diff
@@ -0,0 +1,96 @@
...
+        git checkout -b $BRANCH
+        git add .
+        git commit -m "$PR_TITLE"
+        git push origin $BRANCH
+
+        gh pr create -H $BRANCH --title "$PR_TITLE" --body "$PR_DESCRIPTION"
+    - if: github.sha != steps.fetch_default_branch_head.outputs.sha
+      name: "Skip release: commit is not HEAD of default branch"
+      run: |
+        echo "Skipping release because commit (${GITHUB_SHA}) isn't the HEAD of the default branch (${{ steps.fetch_default_branch_head.outputs.sha }})"
```
AgaDufrat, line 96:
```suggestion
        echo "Skipping release because commit (${github.sha}) isn't the HEAD of the default branch (${{ steps.fetch_default_branch_head.outputs.sha }})"
```

AgaDufrat (approved):
Looks great, thanks!

.github/workflows/autorelease-rubygem.yml
```diff
@@ -0,0 +1,96 @@
...
+      with:
+        fetch-depth: 0
+        show-progress: false
+    - uses: ruby/setup-ruby@v1
+      with:
+        rubygems: latest
+        bundler-cache: true
+    - name: "Determine HEAD of default branch"
+      id: fetch_default_branch_head
+      run: echo "sha=$(git ls-remote origin HEAD | cut -f 1)" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
robinjam: Good catch! IIRC this part was copypasta from a different workflow which uses a shallow clone (hence `rev-parse` wouldn't work there).

.github/workflows/autorelease-rubygem.yml
robinjam: I think this would need double squiggly brackets (e.g. `${{ github.sha }}`), but I'm not certain (and the environment variable definitely works).

robinjam:
Thanks Aga!

> Covers a wider range of scenarios, which we may want to skip.

Not sure what you mean here? Sorry

> I wonder if we need the second conditional at all 🤔. I might be good to keep it there for testing.

The 2nd conditional is there in case we have any gems where the version _isn't_ sourced from a file called `version.rb`. We've had at least 1 of those in the past, where the version was read from inside a git submodule, though I can't remember what it was called (and it's been archived now).

As a general rule I tried to add a conditional to verify every assumption my script makes about the structure of our gem repos.

.github/workflows/autorelease-rubygem.yml
AgaDufrat: Ignore it then!

AgaDufrat:
> > Covers a wider range of scenarios, which we may want to skip.

> Not sure what you mean here? Sorry 😞

Issue #1127, opened 2024-02-15
Move resources out of govuk-aws and retire it


samsimpson1:
Move all of the still in-use resources out of govuk-aws and into this repository so everything is managed in one place by Terraform Cloud
```[tasklist]
### Remaining projects
- [x] app-publishing-amazonmq
- [x] infra-csp-reporter
- [ ] infra-fastly-logs
- [x] infra-monitoring
- [x] infra-networking
- [x] infra-public-wafs
```

jasonBirchall:
Sam to chat with Murilo about this async.

samsimpson1:
All resources that were managed in govuk-aws have been moved over to govuk-infrastructure, and the old repos have been archived.

PR #1187, opened 2024-03-19
Add vpc module

samsimpson1:
#1127

### DIFF
--- a/terraform/deployments/vpc/variables.tf
+++ b/terraform/deployments/vpc/variables.tf
@@ -9,12 +9,6 @@ variable "aws_region" {
   default     = "eu-west-1"
 }
 
-variable "stackname" {
-  type        = string
-  description = "Stackname"
-  default     = ""
-}
-
 variable "vpc_cidr" {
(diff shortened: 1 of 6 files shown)


terraform/deployments/vpc/variables.tf
```diff
@@ -0,0 +1,32 @@
...
+  type        = string
+  description = "AWS region"
+  default     = "eu-west-1"
+}
+
+variable "stackname" {
+  type        = string
+  description = "Stackname"
+  default     = ""
+}
```
sengi, line 16: optional: we might be able to drop "stackname" already (that's the legacy "blue" name label that appears in some domain names) — no worries if not though.

sengi (approved):
Yay!

terraform/deployments/vpc/variables.tf
samsimpson1: It wasn't in use anyway, I just forgot to remove it from the terraform that was copied from the old repo

PR #1198, opened 2024-04-03
Exclude vendor files from automated PR

MuriloDalRi:
The issue, which can be seen in [this PR](https://github.com/alphagov/rubocop-govuk/pull/368), is that setup-ruby configures bundler to install gems in vendor/bundle, and this workflow just runs git adds everything.

Adding `git reset -- vendor/bundle` should exclude all of those. 

Thanks @robinjam for suggesting this fix.

### DIFF
--- a/.github/workflows/autorelease-rubygem.yml
+++ b/.github/workflows/autorelease-rubygem.yml
@@ -86,6 +86,7 @@ jobs:
 
         git checkout -b $BRANCH
         git add .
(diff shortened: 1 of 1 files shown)


sengi (approved):
Seems reasonable to me. Ideally these repos would also have `vendor/bundle` in `.gitignore` to prevent this, but seems wise not to rely on that.

Issue #1201, opened 2024-04-08
Update nodes to Amazon Linux 2023.


sengi:
This gets us kernel 6 as well as a bunch of other good stuff.

https://docs.aws.amazon.com/linux/al2023/ug/compare-with-al2.html 

https://aws.amazon.com/blogs/containers/amazon-eks-optimized-amazon-linux-2023-amis-now-available/

sengi:
Oh what a lovely surprise. Thanks, Amazon 🙃 https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.7/deploy/installation/#using-the-amazon-ec2-instance-metadata-server-version-2-imdsv2

I guess I'd better go figure out how to update the launch template.

sengi:
Or maybe let's just set `--aws-region` and `--aws-vpc-id` and get rid of the somewhat surprising dependency on the instance metadata service. Getting the VPC ID into the cluster neatly might be a pain. Let's find out.

sengi:
Looks good in production so far.

PR #1228, opened 2024-04-09
Enable use_custom_launch_template in aws-eks module.

sengi:
The terraform-aws-eks module's [use_custom_launch_template](https://github.com/terraform-aws-modules/terraform-aws-eks/tree/v19.21.0/modules/eks-managed-node-group#input_use_custom_launch_template) param determines whether the module sets its own (sensible) defaults for the launch template or uses the launch template that the EKS API creates.

It turns out that the EKS API's default launch template doesn't play nice with Amazon Linux 2023 because `MetadataOptions.http-put-response-hop-limit` defaults to `1`, which plays havoc with AWS SDK workloads that use instance profile authentication because they can't talk to IMDS.

I suspect this also affects workloads that use IRSA or Pod Identity but I haven't tested that.

The aws-eks module works around this, but only if you let it use its custom launch template.

Just the integration cluster for now, because AL2023 is only rolled out there so far.

#1201

### DIFF
--- a/terraform/deployments/cluster-infrastructure/main.tf
+++ b/terraform/deployments/cluster-infrastructure/main.tf
@@ -39,7 +39,7 @@ locals {
       min_size                   = var.workers_size_min
       instance_types             = var.workers_instance_types
       disk_size                  = var.node_disk_size
-      use_custom_launch_template = false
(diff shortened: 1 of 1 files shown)


dj-maisy (approved):
:rocket:

PR #1231, opened 2024-04-10
Reinstate node disk size in custom launch template.

sengi:
Missed from #1228.

https://www.github.com/terraform-aws-modules/terraform-aws-eks/blob/v19.21.0/modules/eks-managed-node-group/main.tf#L332

### DIFF
--- a/terraform/deployments/cluster-infrastructure/main.tf
+++ b/terraform/deployments/cluster-infrastructure/main.tf
@@ -34,13 +34,25 @@ locals {
       name_prefix = var.cluster_name
       # TODO: set iam_role_permissions_boundary
       # TODO: apply provider default_tags to instances; might need to set launch_template_tags.
-      desired_size               = var.workers_size_desired
-      max_size                   = var.workers_size_max
-      min_size                   = var.workers_size_min
-      instance_types             = var.workers_instance_types
+      desired_size   = var.workers_size_desired
+      max_size       = var.workers_size_max
+      min_size       = var.workers_size_min
+      instance_types = var.workers_instance_types
+      update_config  = { max_unavailable = 1 }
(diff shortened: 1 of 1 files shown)


terraform/deployments/cluster-infrastructure/main.tf
```diff
@@ -34,13 +34,25 @@ locals {
...
+      # TODO(#1201): remove disk_size and use_custom_launch_template after AL2023 rollout.
+      use_custom_launch_template = var.govuk_environment == "integration"
       disk_size                  = var.node_disk_size
-      use_custom_launch_template = var.govuk_environment == "integration" # TODO(#1201): remove with AL2023 rollout.
-      update_config              = { max_unavailable = 1 }
+      block_device_mappings = {
+        xvda = {
+          device_name = "/dev/xvda"
+          ebs = {
+            volume_size           = 60
```
dj-maisy, line 49: Do we want to set this to `var.node_disk_size` or shall we leave this hard-coded for now?

dj-maisy (approved):
One optional comment otherwise all good. 👍

terraform/deployments/cluster-infrastructure/main.tf
dj-maisy: Oh - also, should we add the same mapping to the ARM group?

sengi: Whoops, good catches, thanks!

sengi:
Checked what's up with the `EbsOptimized=false` thing and it's fine; [it's a no-op for modern instance types](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ebs-optimized.html#current).

PR #1234, opened 2024-04-10
Allow TFC to configure AWS Glue and EKS nodegroups + spring cleaning.

sengi:
- Give Terraform Cloud permissions to:
  - create EKS managed node groups
  - manage AWS Glue catalogs/schemas (we'll need it when we port the CDN log ingestion stuff over from alphagov/govuk-aws)
- Add AWS tags for the couple of resources that the tfc-configuration module manages.

Plus some minor syntactic cleanup while we're there. See commits for detail.

### DIFF
--- a/terraform/deployments/tfc-configuration/aws/main.tf
+++ b/terraform/deployments/tfc-configuration/aws/main.tf
@@ -10,21 +10,16 @@ resource "aws_iam_openid_connect_provider" "tfc_provider" {
 
 data "aws_iam_policy_document" "tfc_role" {
   statement {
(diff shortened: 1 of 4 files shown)


samsimpson1 (approved):

sengi:
force-push to resolve trivial merge conflict (provider version update)

sengi:
Heh, turns out the `aws` module is really a standalone root module and not a submodule, so the provider config was needed after all.

I'll see if I can clean that up further in a followup PR.

For now, this all applies cleanly though.

PR #1246, opened 2024-04-16
Add an Aurora postgresql DB for GOV.UK chat

rtrinque:
Add an Aurora postgresql DB for GOV.UK chat

### DIFF
--- a/terraform/deployments/cluster-infrastructure/variables.tf
+++ b/terraform/deployments/cluster-infrastructure/variables.tf
@@ -147,6 +147,30 @@ variable "grafana_db_seconds_until_auto_pause" {
   default     = 7200
 }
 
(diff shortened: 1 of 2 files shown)


terraform/deployments/cluster-infrastructure/variables.tf
```diff
@@ -147,6 +147,30 @@ variable "grafana_db_seconds_until_auto_pause" {
   default     = 7200
 }
 
+variable "chat_db_min_capacity" {
+  type        = number
+  description = "Minimum capacity of the Grafana RDS Aurora Serverless database."
```
kevindew, line 152: A few references to Grafana lingering in this file.

rtrinque: Fixed

PR #1253, opened 2024-04-18
Use the same IAM role for amd64 and arm64 nodes.

sengi:
We don't want separate IAM roles for each node group. The only difference between them is the machine architecture; we want to treat them identically in terms of IAM permissions.


dj-maisy (approved):

sengi:
Dunno how smooth this is gonna to be to roll out; if it's rough in integration then I'll stick some environment guards in afterwards and figure out a better way to roll it out to prod.

sengi:
Diff looks much less scary after sprinkling some `moved`.

PR #1269, opened 2024-04-22
TFC IAM permissions for RDS Performance Insights.

sengi:
Turns out this is yet another case that involves passing IAM roles around.


rtrinque (approved):
Looks good

PR #1270, opened 2024-04-22
Avoid "special" characters in RDS passwords.

sengi:
This can make the RDS API fall over with excuses like `The parameter MasterUserPassword is not a valid password. Only printable ASCII characters besides '/', '@', '"', ' ' may be used.`


rtrinque (approved):
Looks good

PR #1273, opened 2024-04-23
Add GCP OIDC connection resources and import a GCP bucket

samsimpson1:
This adds the initial config for a TFC -> GCP OIDC connection. It has already been applied in integration, but not tested yet.

It also imports a GCP bucket from govuk-aws as a test

TODO: rename the vpc deployment to something more appropriate

#1127

### DIFF
--- a/terraform/deployments/tfc-aws-config/gcp_oidc.tf
+++ b/terraform/deployments/tfc-aws-config/gcp_oidc.tf
@@ -1,0 +1,63 @@
+data "google_project" "project" {}
+
+resource "google_iam_workload_identity_pool" "tfc" {
+  workload_identity_pool_id = "terraform-cloud-${var.govuk_environment}"
+  display_name              = "Terraform Cloud (${var.govuk_environment})"
+  description               = "Identity pool for Terraform Cloud connection to GCP (${var.govuk_environment})"
+}
+
+resource "google_iam_workload_identity_pool_provider" "tfc" {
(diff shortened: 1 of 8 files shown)


terraform/deployments/tfc-aws-config/gcp_oidc.tf
```diff
@@ -0,0 +1,63 @@
...
+resource "google_iam_workload_identity_pool" "tfc" {
+  workload_identity_pool_id = "terraform-cloud-${var.govuk_environment}"
+  display_name              = "Terraform Cloud (${var.govuk_environment})"
+  description               = "Identity pool for Terraform Cloud connection to GCP (${var.govuk_environment})"
+}
+
+resource "google_iam_workload_identity_pool_provider" "tfc" {
+  workload_identity_pool_id          = google_iam_workload_identity_pool.tfc.workload_identity_pool_id
+  workload_identity_pool_provider_id = "terraform-cloud-${var.govuk_environment}"
+  attribute_condition                = "assertion.sub.startsWith(\"organization:govuk:\")"
```
sengi, line 12: Do we also need to assert on audience (`aud`) or is this sufficient? (apologies if this is a poorly-formed question!)

sengi (approved):

terraform/deployments/tfc-aws-config/gcp_oidc.tf
sengi: Ah cool, sounds like aud isn't very useful here then! Thanks!

PR #1274, opened 2024-04-25
Fix tagging the sha for multi-arch images 

theseanything:
For manual deploys, images need to be tagged by the git sha. This however was broken as the manifest job didn't calculate the local sha for current commit. This is fixed by pulling it as an output of the image build job.

This also removes the custom createdDate label, as this should be the time of image build rather than commit date according to OCI specs.

Also remove the extra "sha-" prefix tags, as these aren't used by anything.


sengi (approved):
Thanks! 🙇

PR #1275, opened 2024-04-25
Import database backup buckets from govuk-aws and small GCP changes

samsimpson1:
GCP changes won't work properly until the staging project is sorted out

#1127

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/db_backup_iam.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/db_backup_iam.tf
@@ -1,0 +1,54 @@
+locals {
+  db_backup_service_account_name = "db-backup"
+  env_also_reads_from = {
+    production  = "govuk-production-database-backups-replica"
+    staging     = "govuk-production-database-backups"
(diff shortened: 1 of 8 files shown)


terraform/deployments/govuk-publishing-infrastructure/db_backup_iam.tf
```diff
@@ -0,0 +1,54 @@
+locals {
+  db_backup_service_account_name = "db-backup"
+  env_also_reads_from = {
+    "production"  = "govuk-production-database-backups-replica"
+    "staging"     = "govuk-production-database-backups"
+    "integration" = "govuk-staging-database-backups"
```
sengi, line 6: optional/trivial: can drop some unnecessary quotes here (dunno why tf fmt doesn't deal with this automatically 🤷)
```suggestion
    production  = "govuk-production-database-backups-replica"
    staging     = "govuk-production-database-backups"
    integration = "govuk-staging-database-backups"
```

sengi (approved):
Nice! Thanks for persevering with the GCP auth stuff! 🙇

PR #1276, opened 2024-04-25
Roll out Amazon Linux 2023 nodes to production.

sengi:
Need to get this out of the way so it doesn't block turning up ARM nodes in prod. It's been working great in integration/staging for the past few days.

#1201


dj-maisy (approved):
👍

PR #1286, opened 2024-05-03
Import transition RDS instance

samsimpson1:
This will need to wait until next week, it will take a couple of hours to do the migration and it will result in a bit of downtime (changing subnet groups means having to move the instance out of the VPC and back in again, thanks AWS)

#1127


sengi (approved):
Thanks! 🙇

sengi:
Sorry for not updating those instance types btw. Good catch.

PR #1291, opened 2024-05-08
Try and tag builds on main as "latest".

dj-maisy:
## What?
Some of our helper/sidecar images don't have explicit version tags and instead rely on pulling "latest" which is causing us some issues with our multi-arch images. Also we aren't getting the latest image, which is concerning.

Let's tag all images that are built from the `main` branch of a repo, to make this a more consistent experience.

### DIFF
--- a/.github/workflows/build-and-push-multiarch-image.yml
+++ b/.github/workflows/build-and-push-multiarch-image.yml
@@ -187,6 +187,7 @@ jobs:
           labels: |
             org.opencontainers.image.vendor=GDS
           tags: |
+            type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', 'main') }}
             type=raw,priority=500,value=${{ inputs.gitRef }},enable=${{ startsWith(inputs.gitRef, 'v') }}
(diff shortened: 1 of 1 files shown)


theseanything (approved):

.github/workflows/build-and-push-multiarch-image.yml
```diff
@@ -187,6 +187,7 @@ jobs:
           labels: |
             org.opencontainers.image.vendor=GDS
           tags: |
+            type=raw,value=latest,enable=${{ github.ref == format('refs/heads/{0}', 'main') }}
```
sengi, line 190: The format looks redundant here, I think?

sengi (approved):

.github/workflows/build-and-push-multiarch-image.yml
dj-maisy: Just going by what the documentation says.

sengi: https://docs.github.com/en/actions/learn-github-actions/expressions#format

sengi: `format('refs/heads/{0}', 'main')` is always gonna evaluate to `'refs/heads/main'`.

PR #1294, opened 2024-05-09
Fix unstable GitHub API token permissions.

sengi:
bbab2de / #1256 introduced a second TF resource managing the same underlying object, which led to unstable results on apply, causing GH Actions to break more or less at random.


MuriloDalRi (approved):
👀

PR #1296, opened 2024-05-09
Send Amazon Inspector image scan results to PlatEng.

sengi:
We're paying for this, so we might as well have a look at the results.

Also set the threshold to crit for now. Remains to be seen if there's value in these scan results, but there's really only one way to find out!

It's also quite likely that the "Enhanced" option overlaps with GitHub's scanning features, but let's see.

https://docs.aws.amazon.com/inspector/latest/user/scanning-ecr.html


theseanything (approved):

sengi:
tf fmt 🙃

PR #1300, opened 2024-05-09
Replace ssm for secrets-manager for opensearch

AlexAvlonitis:


sengi (approved):

rtrinque (approved):
Looks good

terraform/deployments/opensearch/main.tf
```diff
@@ -162,10 +162,12 @@ resource "aws_opensearch_domain" "opensearch" {
 CONFIG
 }
 
-resource "aws_ssm_parameter" "opensearch_master_user" {
-  name        = "/service/${var.service}/MASTER_USER"
-  description = "opensearch_password for ${var.service} domain"
-  type        = "SecureString"
-  value       = "${local.master_user},${random_password.password.result}"
```
sengi, line 169: Please consider using JSON for the two fields here, since secretsmanager is designed to be used that way. (Try it in the UI first and you'll see what I mean.)

terraform/deployments/opensearch/main.tf
```diff
@@ -162,10 +162,12 @@ resource "aws_opensearch_domain" "opensearch" {
...
+resource "aws_secretsmanager_secret" "opensearch_passwords" {
+  name = "opensearch-admin-passwords"
```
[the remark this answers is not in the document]
AlexAvlonitis: oh great thanks, shall i name it `govuk/govuk-chat/opensearch` to follow the convention of the rest of the govuk-chat secrets in secrets-manager, or is it unrelated?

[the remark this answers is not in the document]
sengi: Oh sorry didn't see there were already some in there — for sure yeah please use the same prefix 👍

terraform/deployments/opensearch/main.tf
AlexAvlonitis: This part is deleted, unless i'm missing something

sengi: Oh whoops, I must have fumbled the GitHub UI there 😅 Was meant to be on what's now line 171 — still valid tho I think?

sengi: You probably want something like `secret_string = jsonencode({password = random_password.password.result})`

AlexAvlonitis: oh yup i see what you mean, good point, i'll take a look

sengi: Otherwise you'll find some clients will throw a wobbly trying to read the secretversion. (Might get away with a plain value, but it's way easier to extend and change later if you use JSON, plus you avoid any pontential compatibility issues.)

AlexAvlonitis: awesome thanks for the feedback, i added the username as well since it's available (hope it's ok)

sengi: thanks! 🙇

terraform/deployments/opensearch/main.tf
```diff
@@ -162,10 +162,15 @@ resource "aws_opensearch_domain" "opensearch" {
...
-  value       = "${local.master_user},${random_password.password.result}"
+resource "aws_secretsmanager_secret" "opensearch_passwords" {
+  name = "govuk/chat/opensearch"
+}
+
+resource "aws_secretsmanager_secret_version" "opensearch_passwords" {
+  secret_id     = aws_secretsmanager_secret.opensearch_passwords.id
+  secret_string = jsonencode({
+    "username" = local.master_user
+    "password" = random_password.password.result
```
sengi, line 173: trivial thing: the quotes are redundant (dunno why `tf fmt` doesn't fix this, seems to be a bug)
```suggestion
    username = local.master_user
    password = random_password.password.result
```

sengi (approved):

PR #1308, opened 2024-05-15
Output SARIF or JSON depending on repo visibility

MahmudH:
Trello: https://trello.com/c/9K4r5oJk/3501-update-brakeman-workflow-to-work-with-private-repos-5

This PR checks whether the repository is `public` or `private`. If the repo is public then the output is `SARIF` so that it is uploaded to code scanning, if the repo is `private` then the output is `JSON` and it is uploaded as an artifact. This work is needed because the repos that are `private` do not have GitHub Advanced Security enabled, therefore, we cannot upload to code scanning. 

The steps in the workflow are illustrated below:

Public repo `support`:
Screenshot 2024-05-15 at 15 31 04

Private repo `govuk-chat`:
Screenshot 2024-05-15 at 15 38 53

The condition cannot be simplified further to combine the file checks, this is because Github will deprecate the [set-output](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) command, this is why `$GITHUB_OUTPUT` needs to be used. If the output isn't set to true or false, then **ALL** steps are run even when they should not. This is why I've had to do the following:

```
     - name: Check if SARIF file exists
        id: sarif_check
        run: |
          if [ -f brakeman.sarif ]; then
            echo "sarif_exists=true" >> $GITHUB_OUTPUT
          else
            echo "sarif_exists=false" >> $GITHUB_OUTPUT
          fi
```

### DIFF
--- a/.github/workflows/brakeman.yml
+++ b/.github/workflows/brakeman.yml
@@ -20,14 +20,15 @@ jobs:
         with:
           bundler-cache: true
 
-      - name: Run Brakeman
+      - name: Run Brakeman SARIF
         continue-on-error: true
-        run: |
-          if [ "${{ github.repository_visibility }}" == 'public' ]; then
-            bundle exec brakeman . --except CheckRenderInline --quiet -f sarif >> brakeman.sarif
-          else
-            bundle exec brakeman . --except CheckRenderInline --quiet -f json >> brakeman.json
-          fi
+        if: ${{ github.repository_visibility == 'public' }}
+        run: bundle exec brakeman . --except CheckRenderInline --quiet -f sarif >> brakeman.sarif
+
+      - name: Run Brakeman JSON
+        continue-on-error: true
+        if: ${{ github.repository_visibility == 'private' }}
+        run: bundle exec brakeman . --except CheckRenderInline --quiet -f json >> brakeman.json
 
       - name: Check if SARIF file exists
         id: sarif_check
--- a/.github/workflows/brakeman.yml
+++ b/.github/workflows/brakeman.yml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Run Brakeman JSON
         continue-on-error: true
-        if: ${{ github.repository_visibility == 'private' }}
+        if: ${{ github.repository_visibility == 'private' || github.repository_visibility == 'internal' }}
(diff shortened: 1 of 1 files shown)


sengi:
Why don't we have Advanced Security enabled on these repos?

Presumably this is just a temporary fix until we do?

.github/workflows/brakeman.yml
```diff
@@ -22,10 +22,31 @@ jobs:
 
       - name: Run Brakeman
         continue-on-error: true
-        run: bundle exec brakeman . --except CheckRenderInline --quiet -f sarif >> brakeman.sarif
+        run: |
+          if [ "${{ github.repository_visibility }}" == 'public' ]; then
```
sengi, line 26: While this particular example isn't dangerous (I think, assuming there isn't any way for an outside user to influence the value of `github.repository_visibility`), it's best to avoid substituting values directly into shell commands in GitHub Actions like this because it's all too easy to create command-injection vulnerabilities that way.

sengi, line 26: It might be clearer and cleaner to make these separate steps and use GitHub Actions' `if:` rather than doing it in shell.

sengi: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#example-of-a-script-injection-attack

MuriloDalRi (approved):

.github/workflows/brakeman.yml
```diff
@@ -20,12 +20,34 @@ jobs:
...
+      - name: Run Brakeman SARIF
         continue-on-error: true
+        if: ${{ github.repository_visibility == 'public' }}
         run: bundle exec brakeman . --except CheckRenderInline --quiet -f sarif >> brakeman.sarif
 
-      - name: Upload result to Github Code Scanning
-        if: always()
+      - name: Run Brakeman JSON
+        continue-on-error: true
+        if: ${{ github.repository_visibility == 'private' }}
```
AgaDufrat, line 30: We also want any internal repos to output JSON. For example govuk-rota-announcer 
```suggestion
        if: ${{ github.repository_visibility == 'private'  || github.repository_visibility == 'internal' }}
```

PR #1310, opened 2024-05-15
Update sentry config to add places-manager, remove imminence

KludgeKML:
Questions - do we need to rationalise teams? (find-and-view shouldn't exist?)


sengi (approved):
DSNs are per project. Up to you if you wanna rename and keep the history or just let it create a new one.

PR #1311, opened 2024-05-16
Remove old web custom groups from sentry projects.

KludgeKML:
We're trying to move to a single govuk-developers group in Sentry, remove the govuk-find-and-view and govuk-content-interactions-on-platform teams as project owners (which are now obsolete as names anyway)


sengi (approved):
Thanks!

PR #1316, opened 2024-05-17
Increase chat Opensearch data nodes type to m6g.2xlarge

rtrinque:
We're experiencing bad performance with a populated search index. Moving to m6g.2xlarge for data nodes will increase memory per node from 8G to 32G. Which should improve performance.

### DIFF
--- a/terraform/deployments/tfc-configuration/opensearch.tf
+++ b/terraform/deployments/tfc-configuration/opensearch.tf
@@ -38,7 +38,7 @@ module "opensearch-integration" {
     ebs_enabled              = true
     ebs_volume_size          = 45
     service                  = "chat"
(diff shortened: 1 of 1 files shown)


kevindew (approved):
Sounds reasonable to me. 

I guess it might be expensive? Perhaps we may want to reduce our instance count in integration if that is a concern

sengi:
You probably don't need dedicated masters in integration either.

sengi:
Sadly it looks like the way we're passing the tfvars here is [defeating Terraform Cloud's cost estimation feature](https://app.terraform.io/app/govuk/workspaces/tfc-configuration/runs/run-DJJ3pJGUCVKNdFhe) 😢

PR #1321, opened 2024-05-20
Create a shared action for running a node task

kevindew:
This creates a new reusable GitHub workflow that can be used to run a node JS task from a repository. By task I expect it to be something that is defined inside a package.json scripts object (e.g. yarn run lint:js).

It then utilises this workflow on the standardx and stylelint workflows to simplify them.

The motivation for adding this was due to on govuk-chat deviating from standardx to standard and then having to define our own workflow [1]. I looked and saw other repos had also defined their own near identical workflows [2] [3] so felt the govuk-chat need was not unique.

A demo of this working from a branch is available at: https://github.com/alphagov/govuk-chat/pull/236

[1]: https://github.com/alphagov/govuk-chat/pull/235/commits/fbb629e60854fc51808abb8867d84679ca30e1f9
[2]: https://github.com/alphagov/whitehall/blob/main/.github/workflows/lintjs.yml
[3]: https://github.com/alphagov/whitehall/blob/main/.github/workflows/lintprettier.yml

### DIFF
--- a/.github/workflows/node-script.yml
+++ b/.github/workflows/node-script.yml
@@ -1,0 +1,25 @@
+name: Run Node.js Script
+run-name: Run ${{ inputs.script }}
+
+on:
+  workflow_call:
+    inputs:
+      script:
+        description: 'Script to execute (typically `yarn run <script>`)'
+        required: true
+        type: string
+
+jobs:
+  run-script:
+    name: Run Node.js Script
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          show-progress: false
+
+      - name: Setup Node
+        uses: alphagov/govuk-infrastructure/.github/actions/setup-node@main
+
+      - name: Run script
+        run: ${{ inputs.script }}
--- a/.github/workflows/standardx.yml
+++ b/.github/workflows/standardx.yml
@@ -12,14 +12,6 @@ on:
 jobs:
   run-standardx:
     name: Run Standardx
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-        with:
-          show-progress: false
-
-      - name: Setup Node
-        uses: alphagov/govuk-infrastructure/.github/actions/setup-node@main
-
-      - name: Run Standardx
-        run: yarn run standardx ${{ inputs.files }}
+    uses: alphagov/govuk-infrastructure/.github/workflows/node-script.yml@main
+    with:
+      script: yarn run standardx ${{ inputs.files }}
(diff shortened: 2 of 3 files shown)


.github/workflows/node-task.yml
```diff
@@ -0,0 +1,24 @@
+name: Run Node.js Task
+
+on:
+  workflow_call:
+    inputs:
+      task:
+        description: 'Task to execute (typically `yarn run <task>`)'
```
theseanything, line 7:
```suggestion
      script:
        description: 'Script to execute (typically `yarn run <script>`)'
```
Wonder if it's helpful to keep terminology consistent with [yarn docs](https://classic.yarnpkg.com/lang/en/docs/cli/run/)?

.github/workflows/node-task.yml
```diff
@@ -0,0 +1,24 @@
+name: Run Node.js Task
```
theseanything, line 1:
```suggestion
name: Run yarn script
```

Also wonder if we can add a `run-name` here to output which script we're running?

theseanything: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#run-name

theseanything (approved):
Not entirely convinced this a useful abstraction - it saves a few lines, however then makes it quite painful to handle and configuration or options should we need to pass to the setup node or checkout repo steps. It's also quite cumbersome to actually see what's going on. I.e. which version of node are we using, we need to follow two layers of workflow calls. But that aside I'm not that bothered.

.github/workflows/node-task.yml
kevindew: Sounds good 👍

.github/workflows/node-task.yml
kevindew: I've tried this but not been able to see any evidence of it being used. 

Example run: https://github.com/alphagov/govuk-chat/actions/runs/9762483958/job/26946059073

It looks like it uses the name value from the job rather than the workflow file in a workflow call so this has no effect.

kevindew:
Thanks for the review @theseanything. So the goal isn't to save the lines of those files but to avoid having to write an entirely new action for an app that needs to differ slightly: e.g: https://github.com/alphagov/govuk-chat/blob/19650fd2abfc60682d372cfd08d3464538dd0d0a/.github/workflows/ci.yml#L34-L46 is there are a more conventional approach we can take that you think is more useful?

An alternative is that we could have an action to run a script and then having full actions for stylelint and standardx, though it does seem a bit superfluous and would lose the implicit usage of the script one so I think that'd be only something to try if the layering caused problems.

I've updated the branch, just need to work out whether to proceed further with `run-name`

theseanything:
@kevindew sorry I haven't had time to think about this. Is this still needed/urgent, or can we close this for now and revisit later?

theseanything:
Closing for now

PR #1331, opened 2024-05-28
Revert "Remove use of pact broker secrets for gds-api-adapters"

brucebolt:
It turns out these organisation level secrets are required and were previously set in the UI.  Therefore we need to retain access to them in the relevant repos.

Reverts alphagov/govuk-infrastructure#1327


AgaDufrat (approved):
🤞

PR #1332, opened 2024-05-29
Fix autorelease workflow

MuriloDalRi:
This was changed to appease a new linter but broke the code.

https://trello.com/c/TxQI9DMy/3542-fix-autorelease-workflow


MahmudH (approved):
Makes sense. LGTM 👍🏽

PR #1336, opened 2024-06-06
Create tfc workspace for GOV.UK App backend

jflm:
We would like a Terraform Cloud workspace to deploy resources into the Integration account while we carry out some spikes and investigations. I _think_ all that is needed is a new file in `tfc-configuration`, but any further input would be appreciated.

Currently this file makes reference to a separate repo ([govuk-mobile-backend](https://github.com/alphagov/govuk-mobile-backend)) where we are putting resources (for a few reasons, predominantly that we don't want to 'pollute' main production GOV.UK infra with what we're doing, and also that repo is going to contain a mix of tech approaches that are still being investigated). So I am assuming this will work but again happy to be corrected.

### DIFF
--- a/terraform/deployments/tfc-configuration/govuk-app.tf
+++ b/terraform/deployments/tfc-configuration/govuk-app.tf
@@ -1,0 +1,33 @@
+module "govuk-app-integration" {
+  source  = "alexbasista/workspacer/tfe"
+  version = "0.10.0"
+
+  project_name = "govuk-mobile-backend"
+  vcs_repo = {
+    identifier     = "alphagov/govuk-mobile-backend"
+    branch         = "main"
+    oauth_token_id = data.tfe_oauth_client.github.oauth_token_id
+  }
+
+  organization        = var.organization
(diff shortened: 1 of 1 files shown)


terraform/deployments/tfc-configuration/govuk-app.tf
```diff
@@ -0,0 +1,33 @@
+module "govuk-app-integration" {
+  source  = "alexbasista/workspacer/tfe"
+  version = "0.10.0"
+
+  project_name = "govuk-mobile-backend"
+  vcs_repo = {
+    identifier     = "alphagov/govuk-mobile-backend"
+    branch         = "main"
+    oauth_token_id = data.tfe_oauth_client.github.oauth_token_id
+  }
```
theseanything, line 10: Assuming that the terraform is living in this repo, instead of govuk-infrastructure.

theseanything (approved):

jflm:
Closing for the time being as we will just run Terraform locally. Can re-visit in future if necessary.

PR #1340, opened 2024-06-17
Update autorelease-rubygem.yml

MuriloDalRi:
Split check into two for easier debugging. We want to know the exact reason a PR has not been raised.

### DIFF
--- a/.github/workflows/autorelease-rubygem.yml
+++ b/.github/workflows/autorelease-rubygem.yml
@@ -61,8 +61,13 @@ jobs:
           exit 0
         fi
 
+        if [ -z "$(git log "v$curr_ver..main" --no-merges --format=format:%an)" ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/autorelease-rubygem.yml
```diff
@@ -61,8 +61,13 @@ jobs:
           exit 0
         fi
 
+        if [ -z "$(git log "v$curr_ver..main" --no-merges --format=format:%an | sort | uniq)" ]; then
```
sengi, line 64: Given that you're just checking for zero-length output, the `sort` and `uniq` are redundant and a little confusing.

MuriloDalRi: Ahh yeah that was a bad copy paste situation, have removed it

sengi (approved):

PR #1342, opened 2024-06-18
Add workflow to enforce gem version bumps

MuriloDalRi:
This workflow will run when opening a pull request on gem repositories. It checks if there are changes to the CHANGELOG.md and version.rb or *.gemspec files. If those files have not been changed the workflow will fail.

Code changes are often merged into gems without the gem version being bumped. This leads to changes not being released quickly and sometimes piling up.

https://trello.com/c/R1kkdVjt/3560-re-evaluate-effectiveness-of-gem-auto-release-workflow-3

### DIFF
--- a/.github/workflows/gem-bump-checker.yml
+++ b/.github/workflows/gem-bump-checker.yml
@@ -1,0 +1,38 @@
+name: Gem Bump Checker
+
+on:
+  workflow_call:
+    secrets:
+      GH_TOKEN:
+        required: true
+
+jobs:
+  check-files:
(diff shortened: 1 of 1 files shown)


AgaDufrat (approved):
LGMT. Let's try it out with our gems. I can see people objecting to such frequent releases especially, where gems have slightly [different release process](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/publishing-to-rubygems.md). However this workflow is not mandatory and won't block merging but will make it easier for the maintainers, if they choose to use it.

PR #1346, opened 2024-06-20
Explictly list GH repos that can write to ECR.

sengi:
We were using GitHub Topics (labels that can be applied to a repo) to determine which repos should be permitted to write to our ECR container registry.

Turns out this is inadequate, because anyone in the alphagov GitHub org who can create repos (roughly 700 people, most of whom don't work on the GOV.UK web CMS) can create a repo and set whatever topics they like on it, then get write access to our private ECR.

For now, let's just go back to maintaining a list of authorised repos. That shouldn't be too toily in the short term, since we don't create new repos all that often.


theseanything (approved):

samsimpson1 (approved):
Is this data source and the github provider in general required in this module anymore?

https://github.com/alphagov/govuk-infrastructure/pull/1346/files#diff-55010582a0ff51a6fba1e59d90a4d7ba08925e8d7362a9bbc2d7d3fc0ef90d48R49

sengi:
dammit 😭

PR #1349, opened 2024-06-20
Revert "Explictly list GH repos that can write to ECR."

sengi:
Ughhh the policy doesn't fit in the stoopid 2 kbyte limit. Whyyyy Amazon whyyyy 😭 

\old_man_yells_at_cloud

Reverts alphagov/govuk-infrastructure#1346


theseanything (approved):
😢

PR #1350, opened 2024-06-20
Split IAM policy for ECR push per GitHub repo.

sengi:
This generates a nasty quantity of IAM guff, but this is the next simplest thing I can think of since the 2 kbyte assume-role/trust policy limit scuppers the obvious solution of just listing all the repos in a condition.

This way also restricts a GH repo to writing to its own repo in the container registry, so at least we get something in return for the clutter.

Second attempt at #1346.


sengi:
Ah the images that come from alphagov/govuk-infrastructure are an edge case :/

theseanything (approved):
Not as horrible as I thought terraform wise.

sengi:
I wonder if we can just (hah) rename those three images like `toolbox` -> `govuk-infrastructure/toolbox` 🤔

sengi:
Or just do the dumb thing and unroll a special-case policy/role/attachment for govuk-infrastructure (and maybe but probably not clean it up later 🙈)

sengi:
Ah Licensify has the same thing. Ugh I guess I just need to make this a map.

sengi:
Ok I think that should cover the existing permissions now.

sengi:
Copying this here from Slack since it's relevant:

> sengi:
> 
> Medium-term we'd probably be better off having container builds go to GitHub Packages (ghcr.io) and make ECR a pull-through cache of ghcr. Then we wouldn't need a shedload of IAM garbage just to stop Actions from being able to write under other repos in the registry.
>
> and that has the major bonus that the built images get the same visibility as the source code, which should help a lot with lowering the barrier to entry for development
>
> ...
>
> It'd be really cool (from a security perspective at least) if with that plus switching from webhooks to having Argo watch for tag changes (Sean's suggestion from earlier) we could get rid of most (all?) of the authnz complexity and coupling between GitHub and AWS

PR #1357, opened 2024-06-24
Refactor variable into local variable for opensearch snapshots.

ianhowell-gds:
Having the variable in `tfc-configuration`, the ARN colons are replaced by equals signs by the `variable-set` module. This change makes it all a bit clean and tidier by having all code together.
modified:   iam.tf
new file:   locals.tf
modified:   variables.tf

## Trello Card
https://trello.com/c/zIiHKPHx/1348-configure-daily-data-sync-of-opensearch-data-from-production-to-staging-and-integration


sengi:
This seems like something that shouldn't need to be worked around 🤔 cc @samsimpson1 in case you have any ideas

PR #1358, opened 2024-06-27
Use preinstalled Docker from the runner image.

sengi:
We no longer need to install Docker ourselves now that https://github.com/actions/partner-runner-images/tree/main/images are available.

Tested: https://github.com/alphagov/release/actions/runs/9697881909


dj-maisy (approved):
🚀

PR #1364, opened 2024-07-03
Fix autorelease bug

MuriloDalRi:
The workflow is not working for repos where the version is set in the gemspec file.

https://trello.com/c/R1kkdVjt/3560-re-evaluate-effectiveness-of-gem-auto-release-workflow-3


MahmudH (approved):
LGTM 👍🏽

AgaDufrat:
It wasn't working when a version was set in `version.rb` either. This should fix them both 🤞

PR #1365, opened 2024-07-03
Push to images to ghcr.io

theseanything:
This enables the image build and push workflows to also push images to ghcr.io (whilst continuing to push to ECR). This allows us to configure pull through cache for ECR, instead having to push directly to ECR.

Future work should rename the inputs to remove references to "ecr".

Calling workflows need to be updated to pass the `packages: write` permission, before this can be merged.

Tested on [Release](https://github.com/alphagov/release/pkgs/container/release).

### DIFF
--- a/.github/workflows/build-and-push-image.yml
+++ b/.github/workflows/build-and-push-image.yml
@@ -32,6 +32,7 @@ jobs:
     permissions:
       id-token: write
       contents: read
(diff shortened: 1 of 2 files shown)


dj-maisy (approved):
:octocat:

sengi (approved):

PR #1370, opened 2024-07-08
Replace a bunch of glue actions with super-linter.

sengi:
See https://github.com/alphagov/govuk-ruby-images/pull/103.


samsimpson1 (approved):

sengi:
Thanks! I'll fix up the existing lints (wasn't expecting so many!) and/or the config before merging.

PR #1374, opened 2024-07-08
Migrate GitHub repo config

MahmudH:
Migrate Github repo config from Ruby to Infrastructure as Code. 

Adds merge settings and branch protection resource settings.

Trello: https://trello.com/c/zXGouwXs/3575-investigate-migrating-github-repos-config-to-govuk-infrastructure-3-days
Closes https://github.com/alphagov/govuk-infrastructure/issues/1398

### DIFF
--- a/terraform/deployments/github/imports.tf
+++ b/terraform/deployments/github/imports.tf
@@ -1226,3 +1226,8 @@ import {
   to = github_branch_protection.govuk_repos["whitehall"]
   id = "whitehall:main"
 }
+
+import {
+  to = github_repository.govuk_repos["govuk-crd-library"]
+  id = "govuk-crd-library"
+}
--- a/terraform/deployments/github/imports.tf
+++ b/terraform/deployments/github/imports.tf
@@ -430,16 +430,6 @@ import {
 }
 
 import {
-  to = github_repository.govuk_repos["govuk-display-screen"]
-  id = "govuk-display-screen"
-}
-
-import {
-  to = github_branch_protection.govuk_repos["govuk-display-screen"]
-  id = "govuk-display-screen:main"
-}
-
-import {
   to = github_repository.govuk_repos["govuk-dns-tf"]
   id = "govuk-dns-tf"
 }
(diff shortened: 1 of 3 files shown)


terraform/deployments/github/main.tf
```diff
@@ -115,6 +115,44 @@ resource "github_team_repository" "govuk_repos" {
...
+
+  required_status_checks {
+    strict = true # This ensures the branch is up to date before merging
+    contexts = [
+      "Lint SCSS / Run Stylelint",
+      "Security Analysis / Run Brakeman",
+      "Lint Ruby / Run RuboCop",
+      "Lint JavaScript / Run Standardx",
+      "Test JavaScript / Run Jasmine",
+      "Test Ruby / Run RSpec"
```
AgaDufrat, line 151: Unfortunately the checks are not the same for every repo, see: https://github.com/alphagov/govuk-saas-config/blob/main/github/repo_overrides.yml

MuriloDalRi:
The plan is looking better, I think the errors are because those 3 repos have been archived so they can be removed from the JSON.

terraform/deployments/github/imports.tf
theseanything, line 1: Instead of writing each import statements you could use a for_each https://developer.hashicorp.com/terraform/language/import#import-multiple-instances-with-for_each

terraform/deployments/github/repos.yml
```diff
@@ -0,0 +1,651 @@
...
+repos:
+  accessible-autocomplete-multiselect: {}
+
+  account-api:
+    required_status_checks:
+      standard_contexts: *standard_govuk_rails_checks
+      additional_contexts:
+        - Test Ruby
+        # Status checks missing from here but present in [ci.yml](https://github.com/alphagov/account-api/blob/main/.github/workflows/ci.yml):
+        # - Run Pact tests
```
theseanything, line 20: Looks like lots of addition contexts are still missing from this config

terraform/deployments/github/repos.yml
```diff
@@ -0,0 +1,651 @@
...
+  licensify: {}
+  public-asset-checker: {}
+
+  ##### GOVUK-SAAS-CONFIG TEST REPOS _ONLY_ BELOW THIS LINE #####
+
+  repo-for-govuk-saas-config-automated-tests:
+    up_to_date_branches: true
+
+  xx-test-fixture-with-allow-squash-merge:
+    allow_squash_merge: true
```
theseanything, line 651: Do we need these any more?

terraform/deployments/github/imports.tf
MahmudH: I followed this method [here](https://github.com/alphagov/govuk-infrastructure/pull/948) because of the imports.tf will be removed after this PR is merged.

terraform/deployments/github/repos.yml
MahmudH: You're right we don't, will remove it.

theseanything:
@MahmudH I don't think you've address the previous comments and it looks like terraform is still making lots of changes. This should be a no-op.

theseanything (approved):

terraform/deployments/github/repos.yml
```diff
@@ -0,0 +1,858 @@
...
+      standard_contexts: *standard_security_checks
+      additional_contexts:
+        - test
+
+  gds_zendesk:
+    required_status_checks:
+      standard_contexts: *standard_security_checks
+      additional_contexts:
+        - test
+
```
AgaDufrat, line 228: We archived this repo.

```suggestion
```

MahmudH: I removed the config. Thanks!

PR #1393, opened 2024-07-17
Tag RDS instances with their project

richardTowers:
This will be used to differentiate publishing databases from web / ai / app etc. in cost explorer.

project feels like the most appropriate of the 8 cost allocation tags we currently have in the organisation (aws:createdBy, Customer, Environment, Name, Service, Source, chargeable_entity, project)

These will have to be set in the `databases` variable in Terraform Cloud in each environment.

We should prefix tag values with "GOV.UK - " so that they're easily distinguishable from values elsewhere in the AWS Organisation when looking at cross account bills.


dj-maisy (approved):
🚀

PR #1396, opened 2024-07-19
Change SSE for S3 bucket from "aws:kms" to "AES256" 

ianhowell-gds:
Change SSE for S3 bucket from "aws:kms" to "AES256" so that cross account decryption of files can take place.
This is required as snapshots taken from Production need to be imported into Staging, and then from Staging into Integration.


rtrinque (approved):
Looks good

PR #1399, opened 2024-07-23
Fix names of kubernetes nodes

richardTowers:
Currently in AWS each of the arm nodes is just called "arm", and each of the non-arm nodes is called "main".

image

It looks like the eks module is pulling these from the key (main / arm) in this map.

We switched from using name to name_prefix in fa08608c, but the [documentation for the eks module](https://github.com/terraform-aws-modules/terraform-aws-eks/tree/master/modules/eks-managed-node-group#input_name) doesn't mention name_prefix - it only has name and use_name_prefix.

Switching to name should hopefully mean the module will name the instances with something more descriptive than "main". Specifically, this PR will call the nodes `govuk-k8s-node` / `govuk-k8s-node-arm`.

### DIFF
--- a/terraform/deployments/cluster-infrastructure/main.tf
+++ b/terraform/deployments/cluster-infrastructure/main.tf
@@ -58,7 +58,7 @@ locals {
   arm_managed_node_group = {
     arm = {
       ami_type              = "AL2023_ARM_64_STANDARD"
-      name                  = "${var.cluster_name}-k8s-node-arm"
+      name                  = "${var.cluster_name}-k8s-node"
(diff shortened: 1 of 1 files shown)


richardTowers:
This still doesn't seem to be setting the names with prefixes:

image

richardTowers:
Ah, it does set it as a prefix for the node_group itself, just not in the launch templates of the actual nodes.

image

That's okay. Don't think we really need prefix-style names for the nodes. More convenient in billing if they've all got the same name.

theseanything (approved):

PR #1403, opened 2024-07-25
Manage licensify documentdb clusters

samsimpson1:
Imported from app-licensify-documentdb in govuk-aws

The root password will be manually imported before applying.

#1127

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/licensify_documentdb.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/licensify_documentdb.tf
@@ -1,0 +1,84 @@
+data "terraform_remote_state" "infra_security" {
+  backend = "s3"
+
+  config = {
+    bucket = "${var.govuk_aws_state_bucket}"
+    key    = "govuk/infra-security.tfstate"
+    region = "eu-west-1"
+  }
+}
+
+resource "random_password" "licensify_documentdb_master" {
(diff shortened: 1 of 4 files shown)


terraform/deployments/govuk-publishing-infrastructure/licensify_documentdb.tf
```diff
@@ -0,0 +1,84 @@
...
+import {
+  to = aws_docdb_cluster.licensify_cluster
+  id = "licensify-documentdb-${var.govuk_environment}"
+}
+
+resource "aws_docdb_cluster_instance" "licensify_cluster_instances" {
+  count              = var.licensify_documentdb_instance_count
+  identifier         = "licensify-documentdb-${count.index}"
+  cluster_identifier = aws_docdb_cluster.licensify_cluster.id
+  # TODO: make sure this is the right DB instance size
```
theseanything, line 75: Is this copy pasta? Or do we still need to do this?

theseanything (approved):

terraform/deployments/govuk-publishing-infrastructure/licensify_documentdb.tf
samsimpson1: It is copied from govuk-aws, I have no idea if we still need to do this though.

PR #1406, opened 2024-07-26
Increase EBS volume size and change from m6g to r6g instance type

ianhowell-gds:
### What
Increase EBS volume size from 45Gb to 90Gb, change instance type from `m6g.2xlarge.search` to `r6g.2xlarge.search` for the nodes, and update the engine version from `2.11` to `2.13`.

### Why
The Chat Opensearch cluster has run out of disk space, so we need to increase that to get it working again. We are also taking the opportunity to update the nodes to give more RAM, and put the cluster on the latest supported version of Opensearch.


kevindew (approved):
Nice one, looks good to me

PR #1410, opened 2024-08-01
Add variable in Integration for Test Opensearch Cluster Endpoint

ianhowell-gds:
### What
Add variable `test_opensearch_url` in Integration TFC Configuration

### Why
The Test Opensearch Cluster endpoint will be required for the Route53 record to be created, which is needed by the snapshot cronjob to import indices on a daily basis


rtrinque (approved):
Looks good

PR #1411, opened 2024-08-01
Migrate "infra-mirror-bucket" project from govuk-aws

theseanything:
This migrates the related terraform to manage the govuk-mirror buckets from https://github.com/alphagov/govuk-aws.

This changes the IAM User used by GCP to a federated IAM Role that can be assume. This helps eliminate the need for long lived credentials.


theseanything:
Apart of https://github.com/alphagov/govuk-infrastructure/issues/1127

theseanything:
We can ignore the failure in staging - as it'll just create the new resources, instead of importing.

samsimpson1 (approved):
Thanks for sorting this

PR #1412, opened 2024-08-01
Configure Integration Opensearch environment to allow snapshots to be imported by Test env.

ianhowell-gds:
### What
- Create Route53 record in Integration for K8s cronjob to use for importing indices into Test Opensearch Cluster
- Set permission to allow Test Opensearch access to snapshot S3 bucket
- Update python script to register S3 bucket as repository in Test Opensearch Cluster

### Why
This is one part of the changes required to allow the Test Opensearch Cluster to import a snapshot of Prod indices on a daily basis


rtrinque (approved):
Looks good

PR #1417, opened 2024-08-05
Update permission for Integration to read Production Chat Snapshot S3 bucket

ianhowell-gds:
### What
This change is to update permission for Chat Opensearch in Integration to read from the Production snapshot bucket

### Why
There have been a number of times when the snapshot imported from Staging into Integration has been corrupted somehow, resulting in the Integration Chat Opensearch Cluster Health going Red and the Integration Chat Service experiencing an outage. When the Test and Staging environments import snapshots from Production, this issue has not occurred.


rtrinque (approved):
Looks good

PR #1420, opened 2024-08-07
Add new variables for the Chat terraform project

rtrinque:
We add new resources in the form of a Cloudfront instance that sit in fron of the Chat app.


ianhowell-gds (approved):
LGTM

PR #1421, opened 2024-08-07
Add a cloudfront instance in front of the GOV.UK Chat app

rtrinque:
We add a cloudfront instance in front of the GOV.UK Chat This instance has two origins, during normal operation we use the usual origin in Kubernetes. The other origin is defined as a static website hostes in S3, we can switch to this origin in case we want to stop the Chat service in an emergency.


ianhowell-gds (approved):
Looks good

PR #1425, opened 2024-08-12
Remove custom_origin_config where not applicable

rtrinque:
An S3 origin cannot use custom_origin_config


ianhowell-gds (approved):
Looks good

PR #1433, opened 2024-08-22
Set the TFC service account as CDN bigquery dataset's owner

samsimpson1:
GCP doesn't like it when the dataset doesn't have an owner


Nyzl (approved):
looks good

PR #1439, opened 2024-08-28
s3 sync operation requires s3:ListBucket permission

jflm:
In our GHA we are using `s3 sync <local> <remote>` to publish the config in S3. As per https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html this requires the `s3:ListBucket` permission


gclssvglx (approved):
Yup, makes sense that it needs to list what's in there before putting. Nice one

theseanything (approved):

PR #1440, opened 2024-08-29
update GHA policy to include sub-resources

jflm:
Further to previous 'fix', we also need to update the policy to include sub-resources on the S3 bucket (which again makes sense)


gclssvglx (approved):
Yup, makes total sense that this is needed 👍 Thanks @jflm

PR #1443, opened 2024-09-10
Set maxmemory-policy=noeviction parameter for Chat Redis

kevindew:
This configures the Redis instance that GOV.UK Chat uses for Sidekiq to use it's own parameter group and for that group to have `maxmemory-policy=noeviction`. This is to resolve the warning we get from Sidekiq of:

```
WARNING: Your Redis instance will evict Sidekiq data under heavy load.
The 'noeviction' maxmemory policy is recommended (current policy: 'volatile-lru').
See: https://github.com/mperham/sidekiq/wiki/Using-Redis#memory
```

I need to confirm that this terraform a) actually works and b) doesn't mess up any other parameters.


ianhowell-gds (approved):
LGTM

PR #1446, opened 2024-09-12
Remove cluster-enabled from parameters

kevindew:
Continuing my whack-a-mole process of trying to get a Terraform plan that will apply from
https://github.com/alphagov/govuk-infrastructure/pull/1443.

It turns out we don't want to set cluster enabled to yes and the default option is no, so we shouldn't have been trying to change it.

I've been through and compared all the other parameters and they seem to match.


ianhowell-gds (approved):
Just checked the cluster and it does say Cluster mode is Disabled

PR #1448, opened 2024-09-26
Remove special-route-publisher (now retired)

KludgeKML:
Retired as of: https://github.com/alphagov/special-route-publisher/pull/353

https://trello.com/c/9rOxWNnD/396-retire-special-route-publisher


dj-maisy (approved):
_@KludgeKML casts **Yeetus Deletus**. It's super-effective._

💣

PR #1449, opened 2024-09-30
Remove S3 lifecycle configuration from chat opensearch snapshot bucket

ianhowell-gds:
## What

Remove the lifecycle configuration from the Chat Opensearch Snapshot buckets

## Why

We have discovered that all the files in the bucket are required when the restore of a snapshot takes place. When older files have transitioned into Glacier, the result is that the restore fails. As we are restoring indices from Production into Staging & Integration on a daily basis, to prevent the process from causing them to become unavailable this change is to stop any further files being transitioned.

The daily snapshot job is configured to delete previous snapshots, leaving only the latest in place, which should prevent the bucket from building up with unnecessary files that are no longer required.

### DIFF
--- a/terraform/deployments/opensearch/s3.tf
+++ b/terraform/deployments/opensearch/s3.tf
@@ -32,32 +32,6 @@ resource "aws_s3_bucket_server_side_encryption_configuration" "opensearch_snapsh
   }
 }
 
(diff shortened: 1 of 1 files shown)


kevindew (approved):
Looks good to me.

It'd be also nice to have some info in the commit or the PR to state our expectations that something else will delete files and this bucket won't just continuously grow? (as per our chat just now)

PR #1458, opened 2024-10-03
Document how to remove repo from GitHub config

AgaDufrat:
To avoid repos being deleted or un-archived.


theseanything (approved):
thanks!

PR #1462, opened 2024-10-11
add workspace config for GCP projects

Nyzl:
This PR adds a config file for GCP projects and defines the workspace of the GA4 GCP project


terraform/deployments/tfc-configuration/gcp-projects.tf
```diff
@@ -0,0 +1,32 @@
...
+
+  team_access = {
+    "GOV.UK Non-Production" = "write"
+    "GOV.UK Production"     = "write"
+  }
+
+  variable_set_names = [
+    "gcp-credentials-production",
+    "common",
+    "common-integration"
```
samsimpson1, line 30: You may not actually need the common var sets on this. Review them remove them from the list if they aren't necessary

https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/tfc-configuration/variables-common.tf

https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/tfc-configuration/variables-integration.tf

samsimpson1 (changes requested):

terraform/deployments/tfc-configuration/gcp-projects.tf
```diff
@@ -0,0 +1,32 @@
+module "gcp-ga4-analytics-352613" {
```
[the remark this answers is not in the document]
Nyzl: thank you, i wasn't sure if it should match the project exactly, but it is cleaner without

terraform/deployments/tfc-configuration/gcp-projects.tf
```diff
@@ -0,0 +1,32 @@
...
+
+  project_name = "govuk-infrastructure"
+  vcs_repo = {
+    identifier     = "alphagov/govuk-infrastructure"
+    branch         = "main"
+    oauth_token_id = data.tfe_oauth_client.github.oauth_token_id
+  }
+
+  team_access = {
+    "GOV.UK Non-Production" = "write"
```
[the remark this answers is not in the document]
Nyzl: I consciously made that decision because I'm not in GOV.UK Production but wanted to be able to initiate runs. Adding Non-Production is a lazy workaround but t feels like adding me to Production is too much and creating a new team for me is also too much, what do you think?

terraform/deployments/tfc-configuration/gcp-projects.tf
Nyzl: Thanks for flagging, i'll remove the common ones because they aren't used.

terraform/deployments/tfc-configuration/gcp-projects.tf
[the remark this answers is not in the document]
Nyzl: Thinking more about it, I should solve the problem and not do a bad workaround. I will ask for a new team to be created, GCP Data Production or something like that

Nyzl (commented):
thanks for the suggestions

terraform/deployments/tfc-configuration/gcp-projects.tf
```diff
@@ -0,0 +1,30 @@
+module "gcp-ga4-analytics" {
+  source  = "alexbasista/workspacer/tfe"
+  version = "0.10.0"
+
+  organization        = var.organization
+  workspace_name      = "gcp-ga4-analytics"
+  workspace_desc      = "GCP project management for the GA4 production project"
+  workspace_tags      = ["production", "ga4-analytics-352613", "gcp"]
```
samsimpson1, line 8: The tag should be updated to match the workspace name

terraform/deployments/tfc-configuration/gcp-projects.tf
```diff
@@ -0,0 +1,30 @@
...
+  terraform_version   = var.terraform_version
+  execution_mode      = "remote"
+  working_directory   = "/terraform/deployments/ga4-analytics-352613/"
```
samsimpson1, line 11: Could you also remove the numbers from this directory name?

terraform/deployments/tfc-configuration/gcp-projects.tf
```diff
@@ -0,0 +1,30 @@
...
+  global_remote_state = true
+
+  project_name = "govuk-infrastructure"
+  vcs_repo = {
+    identifier     = "alphagov/govuk-infrastructure"
+    branch         = "main"
+    oauth_token_id = data.tfe_oauth_client.github.oauth_token_id
+  }
+
+  team_access = {
```
samsimpson1, line 22: Generally we give the non-prod team read-only access so people who don't have prod access yet can still see all of our infrastructure

PR #1474, opened 2024-10-21
Add staging and production variables for publishing AmazonMQ instances

samsimpson1:
#1127


terraform/deployments/tfc-configuration/govuk-publishing-infrastructure.tf
```diff
@@ -60,7 +60,8 @@ module "govuk-publishing-infrastructure-staging" {
     "aws-credentials-staging",
     "gcp-credentials-staging",
     "common",
-    "common-staging"
+    "common-staging",
+    #"amazonmq-staging"
```
theseanything, line 64: how come its commented out?

theseanything (approved):

PR #1479, opened 2024-10-23
Add serverless memcached for Chat service

ianhowell-gds:
## What

We want to have a memcached that [GOV.UK](http://gov.uk/) Chat can use for caching data across pods. We should expect that this cache is accessed when the app is under heavy load so will need to be able to sized according to chat at high activity.

## Why

We want to mitigate against the risk of a user spamming us with sign-ups. To mitigate against that we want to use [Rack::Attack](https://github.com/rack/rack-attack) (A Ruby Gem) which makes use of the Rails caching layer - however as we don’t currently do Rails caching we don’t have this configured.

We decided to go with the serverless option rather than a cluster as this allows for scaling in and out with high usage and over time should save money as we would need a large resilient cluster to cope with peak loads. Maximum storage and throughput limits have been put in place to cap at a bit above what we believe will be the highest traffic levels.

### DIFF
--- a/terraform/deployments/chat/serverless_memcache.tf
+++ b/terraform/deployments/chat/serverless_memcache.tf
@@ -1,0 +1,40 @@
+locals {
+  chat_memcached_name = "chat-memcached"
+}
+
+resource "aws_security_group" "chat_memcached" {
+  name        = local.chat_memcached_name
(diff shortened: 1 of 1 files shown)


rtrinque (approved):
Looks good

PR #1487, opened 2024-10-29
Add variables required for Chat WAF rules

ianhowell-gds:
## What

New variables for TFC to be consumed by Chat WAF rules

## Why

The Chat service needs to be moved onto it's own load balancer and needs it's own WAF ruleset


rtrinque (approved):
Looks good

PR #1488, opened 2024-10-29
Create new WAF ACL for Chat

ianhowell-gds:
## What 

Create a WAF ACL for the Chat service

## Why

The Chat service is being moved from the shared load balancer `backend` to it's own load balancer, and this is in preparation for that move


rtrinque (approved):
Looks good

PR #1490, opened 2024-10-29
Add logging to Chat WAF ACL

ianhowell-gds:
## What

Add logging configuration for Chat WAF ACL

## Why

To provide visibility of any rules being triggered


rtrinque (approved):
Looks good

PR #1491, opened 2024-10-30
Change Chat WAF ACL `SizeRestrictions_BODY` from `Block` to `Allow`

ianhowell-gds:
## What

Change Chat WAF ACL `SizeRestrictions_BODY` from `Block` to `Allow`

## Why

It is possible this rule could block some of the admin pages being accessed


rtrinque (approved):
Looks good

PR #1493, opened 2024-11-04
Remove alternative CNAME for Chat memcache

ianhowell-gds:
## What

Remove the alternative CNAME for Chat memcache

## Why

Testing has shown that memcache needs to verify the hostname when using TLS, so the original URL needs to be used


rtrinque (approved):
Looks good

PR #1494, opened 2024-11-04
Remove Chat WAF ACL

ianhowell-gds:
## What

Remove the resources for Chat Web ACL

## Why

The Chat service is no longer going to be moved to it's own load balancer, so the ACL will not be used and is not required


rtrinque (approved):
Looks good

PR #1495, opened 2024-11-04
Remove variables used by Chat WAF ACL

ianhowell-gds:
## What

Remove the variables in Terraform Cloud that were being used for Chat WAF ACLs

## Why

The Chat WAF ACLs are no longer going to be used so this is to tidy up


rtrinque (approved):
Looks good

PR #1502, opened 2024-11-08
Search API gateway: Proxy all paths and methods

csutter:
This changes the API Gateway for Search API (v2) to proxy through any HTTP method and path (rather than just `search.json`).

This will allow us to do routing on the application level, and allows for `OPTION` requests to deal with CORS in addition to `GET` requests. The entire API provided by Search API v2 is read-only and non-sensitive, so we are okay with all endpoints being publicly available (subject to CDN, caching, WAF etc configured on the API gateway).


sihugh (approved):

csutter:
Closing in favour of a different approach agreed yesterday 🔐

PR #1526, opened 2024-11-21
Update Slack channel name for Signon token expiry alerts

ChrisBAshton:
Most of these alerts are already coming into this channel, and the channel #govuk-2ndline-tech is no more. Let's just put all Signon token related alerts into Publishing Platform's alerts channel.


JonathanHallam (approved):
thanks :)

PR #1527, opened 2024-11-21
Add govuk_content_block_tools repo

pezholio:
This adds the new [govuk_content_block_tools](https://github.com/alphagov/govuk_content_block_tools) repo, so the various Github settings etc can be automagically applied.

### DIFF
--- a/terraform/deployments/github/repos.yml
+++ b/terraform/deployments/github/repos.yml
@@ -369,6 +369,13 @@ repos:
       additional_contexts:
         - test
 
+  govuk_content_block_tools:
+    homepage_url: "https://rubygems.org/gems/content_block_tools"
+    required_status_checks:
+      standard_contexts: *standard_security_checks
+      additional_contexts:
+        - test
+
   govuk_personalisation:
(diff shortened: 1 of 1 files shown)


terraform/deployments/github/repos.yml
```diff
@@ -369,6 +369,13 @@ repos:
       additional_contexts:
         - test
 
+  govuk_content_block_tools:
+    homepage_url: "https://components.publishing.service.gov.uk"
```
AgaDufrat, line 373:
```suggestion
    homepage_url: "https://rubygems.org/gems/content_block_tools"
```
Is there a reason for the discrepancy between repo name and name of the gem?

AgaDufrat (approved):

terraform/deployments/github/repos.yml
pezholio: Yeah, the main reason is we named the gem before we were aware of the pattern of prefixing our gems with `govuk_` - Renaming gems is non-trivial unfortunately, but changing the repo name at least gets us some of the way there.

PR #1529, opened 2024-11-22
Update workflow to support standard format `CHANGELOG` files

brucebolt:
At the moment, the gem autorelease workflow only supports `CHANGELOG` fles that have version numbers with h1 level headings.

Whilst this format is used in most GOV.UK gems, there are some that have been updated to use the more standard format with a single h1 and then h2 level headings for each version.

This latter format meets accessibility standards by only having a single h1 on the page and appears to be a [community standard](https://keepachangelog.com/en/1.1.0/).

Therefore updating the gem autorelease workflow to support both types of file.

### DIFF
--- a/.github/workflows/autorelease-rubygem.yml
+++ b/.github/workflows/autorelease-rubygem.yml
@@ -73,9 +73,13 @@ jobs:
 
         find . -type f \( -iname "version.rb" -o -iname "*.gemspec" \) -exec sed -i "0,/$curr_ver/s//${next_ver}/" {} \;
 
-        printf "# %s\n\n* Update dependencies\n\n" "$next_ver" > NEW_CHANGELOG.md
-        cat CHANGELOG.md >> NEW_CHANGELOG.md
-        mv NEW_CHANGELOG.md CHANGELOG.md
+        if [[ $(head -n1 CHANGELOG.md) == "# Changelog" ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/autorelease-rubygem.yml
```diff
@@ -73,9 +73,13 @@ jobs:
 
         find . -type f \( -iname "version.rb" -o -iname "*.gemspec" \) -exec sed -i "0,/$curr_ver/s//${next_ver}/" {} \;
 
-        printf "# %s\n\n* Update dependencies\n\n" "$next_ver" > NEW_CHANGELOG.md
-        cat CHANGELOG.md >> NEW_CHANGELOG.md
-        mv NEW_CHANGELOG.md CHANGELOG.md
+        if [[ $(head -n1 CHANGELOG.md) == "# Changelog" ]]; then
+          sed -i "s/# Changelog/# Changelog\n\n## $next_ver\n\n* Update dependencies/" CHANGELOG.md
```
yndajas, line 77: **question:** regex

Does the asterisk need escaping given this is regex?

brucebolt: I don't believe it does if in the replacement part of the regex (after the second `/`).

yndajas: Oh that makes sense!

theseanything:
@brucebolt Does this PR need a further review?

MuriloDalRi (approved):

PR #1533, opened 2024-11-25
Merge `search-v2-infrastructure/terraform/environment` into `terraform/deployments/search-api-v2`

ianhowell-gds:
## What

This is an import of the code in `https://github.com/alphagov/search-v2-infrastructure.git` directory `terraform/environment` into `terraform/deployments/search-api-v2`

## Why

We believe that the separate repo for search-v2-infrastructure is a tech debt as `govuk-infrastructure` is the definitive resource for [GOV.UK](http://gov.uk/) infrastructure as code. We are concerned that the leaving this tech debt carries two risks:
  1. `search-v2-infrastructure` will not be maintained as well as `govuk-infrastructure`
  2. that the longer this debt is left the harder it will be to resolve.


rtrinque (approved):
Looks good

terraform/deployments/search-api-v2/discovery_engine.tf
```diff
@@ -0,0 +1,96 @@
+# Creates and configures an unstructured datastore for Google Discovery Engine ("Vertex AI Search")
+# see https://cloud.google.com/generative-ai-app-builder/docs/reference/rest/
+
+module "govuk_content_discovery_engine" {
+  source = "../modules/google_discovery_engine_restapi"
```
csutter, line 5: This depends on the [`discovery_engine_restapi` module](https://github.com/alphagov/search-v2-infrastructure/tree/main/terraform/modules/google_discovery_engine_restapi) that will also need migrating at the same time 💡

csutter (changes requested):
This looks great so far – thanks for taking this off my plate 🙏 

See my comment about the module that also needs migrating, and ideally I'd like to get the following PRs in place on the original repo _before_ we merge this:
* https://github.com/alphagov/search-v2-infrastructure/pull/314 (to save us from migrating about-to-be-decommissioned resources)
* https://github.com/alphagov/search-v2-infrastructure/pull/315 (I'll need this soon and don't want to risk the infra pipeline being broken)

Also just a thought: How much do we consider it worth keeping Git history here? The history is still available in the old repo (which I'm assuming we'll eventually archive), so we can probably just squash this into a single commit – although if you think it's more useful to have it around I'm happy to leave it as is.

ianhowell-gds:
Closing this PR and will recreate it without the decommissioned resources and will include the modules directory

PR #1535, opened 2024-11-26
Ianhowell gds/merge search api v2

ianhowell-gds:
## What

This is an import of the code in `https://github.com/alphagov/search-v2-infrastructure` directory `terraform/environment` into `terraform/deployments/search-api-v2`

## Why

We believe that the separate repo for `search-v2-infrastructure` is a tech debt as `govuk-infrastructure` is the definitive resource for [GOV.UK](http://gov.uk/) infrastructure as code. We are concerned that the leaving this tech debt carries two risks:
1. `search-v2-infrastructur`e will not be maintained as well as `govuk-infrastructure`
2. The longer this debt is left, the harder it will be to resolve


rtrinque (approved):
Looks good

csutter (approved):
Awesome – can't see anything now that will obviously break 🤞

PR #1538, opened 2024-11-28
Search API v2: Add permissions to API service account

csutter:
These are needed as we move some tasks from running as GCP Cloud Functions to being Rake tasks in the API app itself.


ianhowell-gds (approved):
Looks good

PR #1539, opened 2024-11-29
Add `name` tag for clamav_db in EFS

nimalank7:
Description:
- Make it easier to see in the AWS console


nimalank7:
See https://github.com/alphagov/govuk-infrastructure/pull/1540 which removes `clamav-db-efs` altogether.

PR #1550, opened 2024-12-12
Add AWS EFS CSI Driver

nimalank7:
Description:
- Adds a statically provisioned EFS CSI Driver as part of a series of PRs to move asset-manager from `nfs` volume type. This shouldn't affect the current NFS setup as it only provisions the driver not call the RPCs on it
- https://github.com/alphagov/govuk-infrastructure/pull/1549 added the IAM roles for EKS nodes to access EFS
- See https://github.com/kubernetes-sigs/aws-efs-csi-driver/tree/master/ for configuration details
- As part of https://github.com/alphagov/govuk-helm-charts/issues/1883

### DIFF
--- a/terraform/deployments/cluster-services/aws_efs_csi_driver.tf
+++ b/terraform/deployments/cluster-services/aws_efs_csi_driver.tf
@@ -1,0 +1,24 @@
+resource "helm_release" "efs_csi_driver" {
+  chart      = "aws-efs-csi-driver"
+  name       = "aws-efs-csi-driver"
+  namespace  = "kube-system"
+  repository = "https://kubernetes-sigs.github.io/aws-efs-csi-driver"
+  version    = "3.1.1" # TODO: Dependabot or equivalent so this doesn't get neglected.
+
+  values = [yamlencode({
+    controller = {
+      serviceAccount = {
+        create = true
+        name   = data.tfe_outputs.cluster_infrastructure.nonsensitive_values.aws_efs_csi_driver_controller_service_account_name
+        annotations = {
+          "eks.amazonaws.com/role-arn" = data.tfe_outputs.cluster_infrastructure.nonsensitive_values.aws_efs_csi_driver_iam_role_arn
+        }
+      }
+    }
+    storageClasses = [{
+      name          = "efs-sc"
+      apiVersion    = "storage.k8s.io/v1"
+      reclaimPolicy = "Retain"
+    }]
+  })]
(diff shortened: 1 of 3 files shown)


terraform/deployments/cluster-services/aws_efs_csi_driver.tf
```diff
@@ -0,0 +1,24 @@
...
+  name       = "aws-efs-csi-driver"
+  namespace  = "kube-system"
+  repository = "https://kubernetes-sigs.github.io/aws-efs-csi-driver"
+  version    = "3.1.1" # TODO: Dependabot or equivalent so this doesn't get neglected.
+
+  values = [yamlencode({
+    controller = {
+      serviceAccount = {
+        create = true
+        name   = data.terraform_remote_state.cluster_infrastructure.outputs.aws_efs_csi_driver_controller_service_account_name
```
samsimpson1, line 12:
```suggestion
        name   = data.tfe_outputs.cluster_infrastructure.nonsensitive_outputs.aws_efs_csi_driver_controller_service_account_name
```

terraform/deployments/cluster-services/aws_efs_csi_driver.tf
```diff
@@ -0,0 +1,24 @@
...
+      serviceAccount = {
+        create = true
+        name   = data.tfe_outputs.cluster_infrastructure.nonsensitive_values.aws_efs_csi_driver_controller_service_account_name
+        annotations = {
+          "eks.amazonaws.com/role-arn" = data.tfe_outputs.cluster_infrastructure.nonsensitive_values.aws_efs_csi_driver_iam_role_arn
+        }
+      }
+    }
+    storageClasses = [{
+      name          = "clamav-db-efs-sc"
```
samsimpson1, line 19: Could it be better to use a generic name for the SC in case we want to use this somewhere else?

terraform/deployments/cluster-services/aws_efs_csi_driver.tf
nimalank7: Ah I forgot to change this when I made the branch. All fixed.

terraform/deployments/cluster-services/aws_efs_csi_driver.tf
nimalank7: Changed to `efs-sc`

samsimpson1 (approved):

Issue #1546, opened 2024-12-04
IAM Policy for EBS volume change


nimalank7:
AWS are making a change to IAM policies for EBS volumes in **January 2025**. We currently dynamically provision EBS volumes using the EBS CSI Driver. Verify if we need to do update anything.

See https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2190

PR #1553, opened 2024-12-18
Allow EBS CSI driver role to perform ec2:CreateVolume on snapshots

samsimpson1:
This is required because AWS is going to start checking permissions on snapshots when creating a volume from one

https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/2190

Closes https://github.com/alphagov/govuk-infrastructure/issues/1546


nimalank7 (approved):
LGTM!

PR #1566, opened 2025-01-14
Extend Rubocop workflow to allow setting of Github token

jackbot:
This allows workflows which call this reusable workflow to optionally set a
`BUNDLER_GITHUB_TOKEN` secret.

Setting a value for this token will set a `BUNDLE_GITHUB__COM` environment
variable in the `setup-ruby` step. This env var is an authentication
string which Bundler will use when attempting to install gems from
private repositories on Github.

Note: this is backwards-compatible. Omitting this secret when calling the 
workflow has no effect on installing gems from Rubygems, or public gems 
from Github. It only takes effect when installing gems from a private Github repo.

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -2,6 +2,10 @@ name: Run RuboCop
 
 on:
   workflow_call:
+    secrets:
+      BUNDLER_GITHUB_TOKEN:
+        required: false
+        description: "Token used for Bundler to authenticate when installing gems from private Github repos"
 
 jobs:
   run-rubocop:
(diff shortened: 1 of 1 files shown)


kevindew (approved):
Looks good to me 👍 I think before merging we should try it with a repo that has a github gem. I'll just give it a quick whirl on publisher

kevindew:
Test with publisher was successful: https://github.com/alphagov/publisher/actions/runs/12765805593/job/35580811043 based on this code change: https://github.com/alphagov/publisher/compare/rubocop-token-workflow

PR #1568, opened 2025-01-15
Archive `router-api` repository

nimalank7:
Description:
- No longer needs to be managed as Router now retrieves routes directly from Content Store


theseanything (approved):

AgaDufrat:
Do we need to [remove it from TF state first](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/github/README.md#removing-repositories)?

PR #1585, opened 2025-01-21
Lower Instance & Capacity Preferences in Non-Production

dj-maisy:
## What?
This PR changes the default "worker instance types" for x86/AMD64 workloads to be a preference of one of the following:

* `r7i.large`
* `r7a.large`
* `m7i-flex.xlarge`
* `m6i.xlarge`

The basis here is that all these instances are specced with 16GB of RAM. Our current instances in Integration and Staging are using `m6i.4xlarge` which come with 64GB of RAM and is currently underutilised as most workloads are now running on our `m7g.4xlarge` instances.

Further, we can set the minimum instance counts to 0 for this node group, so that our clusters retain the ability (in theory) to support x86/AMD64 workloads without necessarily needing to pay for that availability 24/7.

There may be some further tweaking necessary if we find the node group trying to scale beyond its set limits, e.g. we may find a scheduled task that favours CPU over RAM and can accommodate this accordingly.

I've also "overridden" the min/desired/max counts on the Production config so we keep that as it is until we're ready to do the switch in Production. Which is probably soon!

### DIFF
--- a/terraform/deployments/cluster-infrastructure/grafana.tf
+++ b/terraform/deployments/cluster-infrastructure/grafana.tf
@@ -74,8 +74,9 @@ resource "aws_iam_policy" "grafana" {
 }
 
 data "aws_rds_engine_version" "postgresql" {
-  engine  = "aurora-postgresql"
-  version = "13"
+  engine       = "aurora-postgresql"
+  version      = "13"
+  default_only = true
 }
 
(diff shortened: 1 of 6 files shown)


jasonBirchall:
The naming of that variable scares me a little. 😓 
Is there a chance we can rename it as part of this PR? Perhaps x86 workers?

samsimpson1 (approved):

PR #1593, opened 2025-01-22
Migrate infra-public-wafs from govuk-aws

theseanything:
This is the initial step to migrating infra-public-wafs from govuk-aws. 

aws_wafv2_web_acl_logging_configuration need to be manually imported, as the terraform provider doesn't support the data objects for retrieving the "ids".


samsimpson1:
I assume you're planning on using `terraform import` commands for the logging config, that will probably be fine but if not you could do something similar to what I had to do for the CSP reporter and create a list of the ID for each env

https://github.com/alphagov/govuk-infrastructure/blob/c3e15189a19c6f31474d2228ac0160fd9b2fd7b7/terraform/deployments/csp-reporter/imports.tf

samsimpson1 (approved):

PR #1619, opened 2025-01-23
Switch to renovate to update terraform dependencies

theseanything:
Dependbot doesn't support dependency grouping into a single PR. This causes a lot of noise as specify the dependencies in lots of files.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,11 +1,18 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": [
-    "config:base"
-  ],
+  "extends": ["config:base"],
   "platformAutomerge": true,
   "pre-commit": {
     "enabled": true
   },
-  "enabledManagers": ["pre-commit"]
+  "enabledManagers": ["pre-commit", "terraform"],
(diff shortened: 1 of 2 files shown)


renovate.json
```diff
@@ -1,11 +1,22 @@
...
   "platformAutomerge": true,
   "pre-commit": {
     "enabled": true
   },
-  "enabledManagers": ["pre-commit"]
+  "enabledManagers": ["pre-commit", "terraform"],
+  "terraform": {
+    "fileMatch": ["^/terraform/.*\\.tf$"]
+  },
+  "packageRules": [
```
samsimpson1, line 12: Is there a reason these two things specifically have their own rules?

nimalank7 (approved):
LGTM!

renovate.json
theseanything: So alexbasista/workspacer/tfe module we specify in basically in each file in the "tac-configuration" project - we don't want to update all of those individually.

Probably doesn't make sense to group the hashicorp/tfe - removed.

samsimpson1 (approved):

PR #1634, opened 2025-01-24
Fix renovate configuration to remove packageRule and increase PRs per hr limit

theseanything:
Remove redundant packageRule for workspaces/tfe as it's the default behaviour. This also bumps up the PR per hr rule, as we want to speed up the creation of all our dependency PRs.


AgaDufrat (approved):
👌

PR #1642, opened 2025-01-24
Add repo to allow deployment from publisher-on-postgres-branch

syed-ali-tw:
Mainstream publisher is migrating to postgres. In order to be able to deploy and test in integration, one of the option is to have a separate app and to be able to do this we want to be able to create a separate repo to allow image build from "publisher-on-postgres" branch have this repo as its target.


dj-maisy (approved):
Approved!

PR #1655, opened 2025-01-28
Add workspace for making sensitive variables

theseanything:


terraform/deployments/tfc-configuration/variables-sensitive.tf
```diff
@@ -0,0 +1,26 @@
+module "variables-sensitive" {
+  source  = "alexbasista/workspacer/tfe"
+  version = "0.12.0"
+
+  organization        = var.organization
+  workspace_name      = "variables-sensitive"
+  workspace_desc      = "This module manages sensitive variables for Terraform Cloud workspaces."
+  workspace_tags      = ["tfc", "variables", "configuration"]
```
samsimpson1, line 8: These will need changing, because terraform uses these to find the workspace when you use the CLI
```suggestion
  workspace_tags      = ["tfc", "variables", "sensitive"]
```

samsimpson1 (approved):

PR #1663, opened 2025-01-29
♻️ Replace kanban board link with GitHub projects

jasonBirchall:
## 👀 Purpose

- Very small change to correct the location of our kanban board.

## ♻️ What's changed

- We now use GitHub projects so the link should reflect that.

## 📝 Notes

-


nicholsj (approved):
Thanks!

PR #1666, opened 2025-01-29
Use modules for sensitive WAF config

theseanything:
This uses a private module to create privates WAF resources and set private variables. These modules are defined in https://github.com/alphagov/terraform-govuk-infrastructure-sensitive (annoying had to use this name convention for TFC) .

As part of the tfc-configuration workspace, a sensitive variable set for WAFs is created for each environments.

Then for each of the govuk-publishing-infrastructure projects, they are configured to use the respective variable-set and to create the private WAF resource. 

(tfc-configuration project needs to be run twice; once to create the variable set, then to assign it to the relevant projects. Hence will error on review of this PR.)

### DIFF
--- a/.github/workflows/ci-terraform.yml
+++ b/.github/workflows/ci-terraform.yml
@@ -55,6 +55,7 @@ jobs:
 
     - name: terraform init
       run: |
+        export TF_TOKEN_app_terraform_io="${{ secrets.TFC_CI_READ_ONLY_TOKEN }}"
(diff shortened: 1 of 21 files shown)


samsimpson1:
Could this be a use case for the private module registry? https://developer.hashicorp.com/terraform/cloud-docs/registry/publish-modules

That might be a bit complicated though

MuriloDalRi (approved):

PR #1669, opened 2025-01-30
Add sanity check to github repositories data source

samsimpson1:
This should hopefully protect us from a complete GitHub Search API outage like we had a few weeks ago

https://github.com/alphagov/govuk-platform-internal/issues/9


dj-maisy (approved):
🚀

PR #1671, opened 2025-01-30
Add Graviton (ARM) hardware support to Production

dj-maisy:
## What?
This adds Graviton hardware support to Production, paving the way for us to make the final switchover.

I am also enabling Metrics Server, because this had no adverse reactions outside of production and also opens up the possibility to use Horizontal Pod Autoscaling in the future.

We'll be using r8g.4xlarge as our workloads are memory-intensive rather than CPU-intensive.


jasonBirchall (approved):
Awesome.

PR #1680, opened 2025-02-03
Search API v2: Add new Search Admin serving configs

csutter:
This removes a temporary serving config for Search Admin, and sets up three new ones:
- `govuk_default`
- `govuk_preview`
- `govuk_raw`

Eventually, their lifecycle management should be fully in Search Admin, but the Ruby client doesn't support creating them yet, so we're doing it in TF instead.

This also removes the temporary serving config from the Search Admin AWS secret, and the outputs.

### DIFF
--- a/terraform/deployments/search-api-v2/modules/google_discovery_engine_restapi/main.tf
+++ b/terraform/deployments/search-api-v2/modules/google_discovery_engine_restapi/main.tf
@@ -67,19 +67,52 @@ resource "restapi_object" "discovery_engine_serving_config" {
   })
 }
 
-# Future serving config managed by Search Admin
-resource "restapi_object" "discovery_engine_serving_config_site_search" {
-  path      = "/engines/${var.engine_id}/servingConfigs/site_search"
-  object_id = "site_search"
+# Future serving configs managed by Search Admin
+# NOTE: These are created in TF for now, but should migrate to being fully managed by Search Admin
+# as soon as that is supported by the Discovery Engine Ruby client, at which point they should be
+# removed from state and deleted from this config.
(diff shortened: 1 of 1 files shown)


terraform/deployments/search-api-v2/modules/google_discovery_engine_restapi/main.tf
```diff
@@ -67,19 +67,52 @@ resource "restapi_object" "discovery_engine_serving_config" {
   })
 }
 
-# Future serving config managed by Search Admin
-resource "restapi_object" "discovery_engine_serving_config_site_search" {
```
csutter, line 71: Note that this will cause the `site_search` serving config to be removed.

We don't actually use this anywhere (the real serving config in live use is `default_search` above), so this is fine.

terraform/deployments/search-api-v2/discovery_engine.tf
```diff
@@ -86,7 +85,6 @@ resource "aws_secretsmanager_secret_version" "discovery_engine_configuration_sea
   secret_string = jsonencode({
     "GOOGLE_CLOUD_CREDENTIALS"        = base64decode(google_service_account_key.search_admin.private_key)
     "DISCOVERY_ENGINE_ENGINE"         = google_discovery_engine_search_engine.govuk.name
-    "DISCOVERY_ENGINE_SERVING_CONFIG" = local.discovery_engine_site_search_serving_config_path
```
csutter, line 89: We no longer need this - Search Admin can figure out serving config names from the engine name defined above.

ianhowell-gds (approved):
Looks good to me

PR #1686, opened 2025-02-04
Search API v2: Export default collection name to app

csutter:
This introduces a `DISCOVERY_ENGINE_DEFAULT_COLLECTION_NAME` key to the AWS Secrets Manager secret for both Search API v2 and Search Admin.

This is the _parent_ of both the engine and datastore resources, but unfortunately the Terraform resources don't offer an appropriate attribute, so we have to construct the string ourselves.


ianhowell-gds (approved):
LGTM

PR #1692, opened 2025-02-06
Add bedrock access permission for K8s cluster

ianhowell-gds:
## What

Add Bedrock IAM permissions for K8s cluster

## Why 

The Chat service is trialling Bedrock, so the K8s pods need access via a service account and IAM role


rtrinque (approved):
Looks good

PR #1693, opened 2025-02-06
Fix chat deployment in TFC

ianhowell-gds:
## What

Update value of variables `cloudfront_create` from "1" to "true"

## Why

Bumping Terraform to 1.10 appears to have caused the following error:
```
Error: Invalid value for input variable
on /home/tfc-agent/.tfc-agent/component/terraform/runs/run-zMME98VooKS9PwTX/terraform.tfvars line 9:
cloudfront_create = 1
The given value is not suitable for var.cloudfront_create declared at variables.tf:35,1-29: bool required.
```


rtrinque (approved):
Looks good

PR #1700, opened 2025-02-11
Add govuk-graphql to repos.yml

richardTowers:
This is an early prototype, so disabling branch protection rules for now.


AgaDufrat (approved):
TF plan looks good to me

PR #1711, opened 2025-02-14
Only run terraform CI against changed deployments

samsimpson1:
The amount of time it takes to run the Terraform linting was annoying me, so I've updated the workflow to only run on TF deployments that have been modified.

The linting workflow was taking between 2 and 3 minutes before, and takes around 30-40 seconds now.

Examples:
* [when a change passes linting](https://github.com/alphagov/govuk-infrastructure/actions/runs/13329890591/job/37231273003)
* [when a change fails linting](https://github.com/alphagov/govuk-infrastructure/actions/runs/13329916518/job/37231352673)
* [when no changes are made](https://github.com/alphagov/govuk-infrastructure/actions/runs/13329986321/job/37231578091)

### DIFF
--- a/.github/workflows/ci-terraform.yml
+++ b/.github/workflows/ci-terraform.yml
@@ -12,23 +12,36 @@ jobs:
     - uses: actions/checkout@v4
       with:
         show-progress: false
+    
+    - name: Get changed TF deployments
+      id: changed-deployments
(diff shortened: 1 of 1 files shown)


AgaDufrat (approved):
LGTM. Thanks for the examples

PR #1729, opened 2025-02-26
Add report CI errors github action

kentsanggds:
Add a custom github action to make it easier to send CI errors to a team channel with some notes on how to use the action.

Update - action renamed to `report-run-failure` to make it more generic as this action can be used to report any github action run failures

### DIFF
--- a/.github/actions/report-ci-errors/README.md
+++ b/.github/actions/report-ci-errors/README.md
@@ -4,7 +4,11 @@ You will need to get the slack webhook url from the secrets manager first and ad
 
 ## Usage example:
 
-Note: Typically you will want to report CI errors as part of a deployment workflow. The code below demonstrates where to insert the lines and what lines to copy across.
+Notes: 
+  - To minimise noise the slack message is only sent out on the main branch.
+  - Typically you will want to report CI errors as part of a deployment workflow. 
+
+The code below indicates where to insert the lines and what lines to copy across.
 
 ```
 name: Always failing job
--- a/.github/actions/report-ci-errors/README.md
+++ b/.github/actions/report-ci-errors/README.md
@@ -1,0 +1,26 @@
+## Pre-requisites
+
+You will need to get the slack webhook url from the secrets manager first and add it as SLACK_WEBHOOK_URL as a repository secret. 
+
+## Usage example:
+
+name: Always failing job
+on:
+  workflow_dispatch:
+
+jobs:
+  failed-run:
+    name: Failed run
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Always failing
+        run: exit 1
+
+      # copy lines below to report CI errors to your team channel
+      - if: ${{ failure() }}
+        uses: alphagov/govuk-infrastrtucture/.github/actions/report-ci-errors@main
+        with:
+          slack_webhook_url: ${{ secrets.SLACK_WEBHOOK_URL }}
+          github: ${{ toJson(github) }}
+          channel: your-team-slack-channel
+          message: an optional message
--- a/.github/actions/report-ci-errors/README.md
+++ b/.github/actions/report-ci-errors/README.md
@@ -4,6 +4,9 @@ You will need to get the slack webhook url from the secrets manager first and ad
 
 ## Usage example:
 
+Note: Typically you will want to report CI errors as part of a deployment workflow. The code below demonstrates where to insert the lines and what lines to copy across.
+
+```
 name: Always failing job
 on:
   workflow_dispatch:
@@ -24,3 +27,4 @@ jobs:
           github: ${{ toJson(github) }}
           channel: your-team-slack-channel
           message: an optional message
+```
--- a/.github/actions/report-ci-errors/action.yaml
+++ b/.github/actions/report-ci-errors/action.yaml
@@ -1,0 +1,52 @@
+name: "Report CI errors to slack channel"
+description: "report CI errors to a nominated slack channel"
+inputs:
+  slack_webhook_url:
+    description: "Slack webhook URL"
+    required: true
+  channel:
+    description: "Name of the channel to send error notifications to"
+    required: true
+  github:
+    description: "Github Action context as json"
+    required: true
+  message:
+    description: "Optional message to show in channel"
+    default: ""
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        # run checks as github actions required is not being enforced - https://github.com/actions/runner/issues/1070
+        [[ "${{ inputs.slack_webhook_url }}" ]] || { echo "slack_webhook_url input is empty" ; exit 1; }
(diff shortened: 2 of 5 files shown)


.github/actions/report-ci-errors/README.md
```diff
@@ -0,0 +1,26 @@
...
+on:
+  workflow_dispatch:
+
+jobs:
+  failed-run:
+    name: Failed run
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Always failing
+        run: exit 1
```
AgaDufrat, line 17: I'd use a real use case example here like `name: Deploy`. This is how the workflow can be tested so might be a bit confusing/unhelpful.

.github/actions/report-ci-errors/README.md
```diff
@@ -0,0 +1,26 @@
+## Pre-requisites
+
+You will need to get the slack webhook url from the secrets manager first and add it as SLACK_WEBHOOK_URL as a repository secret. 
+
+## Usage example:
```
AgaDufrat, line 5: The formatting is off here
Screenshot 2025-02-27 at 12 45 49

.github/actions/report-ci-errors/README.md
kentsanggds: The code here is just for showing where the code is inserted and used rather than being a demonstration of a use case

kentsanggds: I've added a note to state where the action might be typically used

.github/actions/report-ci-errors/README.md
```diff
@@ -0,0 +1,26 @@
+## Pre-requisites
+
+You will need to get the slack webhook url from the secrets manager first and add it as SLACK_WEBHOOK_URL as a repository secret.
```
AgaDufrat, line 3: We want to move away from ClickOpsing GitHub config so ideally it should be added in the Terraform config. Similar to organisations secrets https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/github/main.tf#L224-L253

As per our chat on Slack: https://gds.slack.com/archives/C079ULB1UEP/p1740663568473939

.github/actions/report-ci-errors/action.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - name: Send slack
+      uses: slackapi/slack-github-action@v2.0.0
+      with:
+        errors: true
+        webhook: ${{ inputs.slack_webhook_url }}
+        webhook-type: incoming-webhook
+        payload: |
+          {
+            "channel": "${{ inputs.channel }}",
+            "text": "The <https://github.com/${{ fromJson(inputs.github).repository }}> CI test failed.",
```
AgaDufrat, line 33: The original request was for CI failures but it can be used to report any GitHub action run failure so I think we could make this message more generic.
```suggestion
            "text": "The <https://github.com/${{ fromJson(inputs.github).repository }}>'s GitHub workflow run failed.",
```

It would also be good if it included workflow name as repos tend to have a few. The acceptance criteria also mentions `branch` name but we decided it should be scoped to failures on `main` branch other wise it will be too noisy.

AgaDufrat (commented):
[The test](https://gds.slack.com/archives/C08D696LAEA/p1740667524061079) looks good.

I left a few comments. I’d give it more general name and message. We should try to scope it to main branch.

.github/actions/report-ci-errors/action.yaml
```diff
@@ -0,0 +1,52 @@
...
+inputs:
+  slack_webhook_url:
+    description: "Slack webhook URL"
+    required: true
+  channel:
+    description: "Name of the channel to send error notifications to"
+    required: true
+  github:
+    description: "Github Action context as json"
+    required: true
```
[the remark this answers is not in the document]
kentsanggds: I'll have a check as I was passing it through to another action but as it's reusing an action maybe its not needed...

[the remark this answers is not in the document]
kentsanggds: I think I ran into this issue before, it might be different because its a custom action and not a workflow. 

I tried to show whats available in the github context to make sure that the correct context was picked up and got an error https://github.com/alphagov/ckanext-datagovuk/actions/runs/13572015348/job/37939186912, this code change is described here https://github.com/alphagov/govuk-ci-github-action/releases/tag/test-7

.github/actions/report-ci-errors/action.yaml
kentsanggds: ok sure, that makes sense

.github/actions/report-ci-errors/action.yaml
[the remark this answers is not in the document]
AgaDufrat: I've also seen it being used in actions 🤔 
https://github.com/alphagov/govuk-infrastructure/blob//021b1a55271496c6c55dee943b8f3e5119c83f64/.github/actions/precompile-rails-assets/action.yml#L12

It would be nice to simplify the config, but it's not blocking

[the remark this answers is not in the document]
kentsanggds: actually, I checked the error message and it looks like the context is available, so I'll remove that github input

.github/actions/report-run-failure/README.md
```diff
@@ -0,0 +1,30 @@
+## Notes on using this action
+
+  - GOVUK_SLACK_WEBHOOK_URL should be available as a secret in your repository, if it is not please post a message on #govuk-platform-engineering
```
AgaDufrat, line 3: I would make this note more informative and remove the "please ask..." bit. It implies that we are expecting this not to work properly which shouldn't be the case. Recent user research confirmed product developers know how to reach our team. This repo is owned by us and the single source of truth of for repos ownership are the docs.

```suggestion
  - This action relies on GOVUK_SLACK_WEBHOOK_URL actions secret being configured for the repository. It’s added to all GOV.UK repositories as an organisation secret in the [GOV.UK GitHub Infrastructure configuration
](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/github/main.tf)
```

.github/actions/report-run-failure/README.md
```diff
@@ -0,0 +1,30 @@
...
+  - To minimise noise the slack message is only sent out on the main branch.
```
AgaDufrat, line 4:
```suggestion
  - To minimise noise, the slack message is only sent out for failures on the main branch.
```

AgaDufrat (approved):
I’m approving as I’m on leave until 11 March. Thanks for addressing all the comments. I suggested some further minor rewording.

The only thing outstanding before we can merge is to add the secret to all `govuk` repos in https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/github/main.tf.

.github/actions/report-run-failure/README.md
kentsanggds: ok, I'll pick up the wording

PR #1761, opened 2025-03-05
Search API v2: Remove Search Admin serving configs

csutter:
We added these when we were planning to move controls and serving configs management into Search Admin imminently.

Given that this work has hit a major blocker (the Google client SDK lagging behind the REST API), we are pausing this work for now.

With that in mind, these serving configs aren't used and are just in the way, so let's remove them until we pick this work up again.


ianhowell-gds (approved):
Makes sense

PR #1762, opened 2025-03-05
Search API v2: Try fixing automated eval function

csutter:
This pins NumPy to the previous major version which seems like it could be the cause of the function refusing to deploy of late.

Upstream seems to be at 2.x now, which probably isn't supported by the older version of Pandas we're running for this.


ianhowell-gds (approved):
Looks good

PR #1774, opened 2025-03-05
Search API v2: Modularise serving configs

csutter:
This refactors serving configs into a module to make creating several of them less painful, and to fix some long standing tech debt around how we use the REST API client.

It removes the `search_raw` serving config for now (which is only used for internal testing) and imports the existing `default_search` serving config using the new module.

⚠️ **Need to remove the `default_search` serving config from state to make sure the old code doesn't try to delete it**


ianhowell-gds (approved):
Looks good

PR #1776, opened 2025-03-05
Search API v2: Fix incorrect serving config argument

csutter:
This kind of controls is called synonyms (plural) rather than synonym (singular).


ianhowell-gds (approved):
Makes sense to me

PR #1778, opened 2025-03-06
Search API v2: Fix serving config static properties

csutter:
The serving config is refusing to update as `solutionType` is an immutable property so can't be specified in the update mask (even though we're not updating it anyway).

This splits out static and dynamic properties so we only update the dynamic ones that actually can be updated.


ianhowell-gds (approved):
Looks good

PR #1786, opened 2025-03-06
Search API v2: Properly build control updateMask

csutter:
This allows updating nested values such as `synonymAction.synonyms`, and fixes the unpermitted top-level `synonymAction` being included in the updateMask.


csutter:
Verified it does what it should do:
image
image

hamish-pollock-gds (approved):

Issue #1742, opened 2025-03-03
☸️ Provision an ephemeral Kubernetes cluster


jasonBirchall:
**User Need**

**As a** platform engineer  
**I want** to create a Kubernetes cluster in our new AWS sandbox account  
**so that** we have an ephemeral environment for deploying GOV.UK-related components independently of our main integration/staging/production pipelines

**Functional Requirements (What):**

- [ ] Define cluster specifications (node types, networking, etc.) in IaC  
- [ ] Provision the cluster via Terraform or equivalent tooling in the new AWS account  
- [ ] Confirm basic cluster connectivity (kubectl and/or Argo CD once installed)

**Non-Functional Requirements (How):**

- [ ] Ensure cluster naming, tagging, and logging align with ephemeral environment standards  
- [ ] Maintain reproducibility (all steps in code, no manual changes)

**Acceptance Criteria:**

- [ ] The Kubernetes cluster is successfully created and reachable  
- [ ] Networking, IAM roles, and any default security configurations pass initial checks  
- [ ] Documentation or IaC references are stored in a shared repo with clear instructions

**Notes:**

- This user story is linked to the epic: [#1725](https://github.com/alphagov/govuk-infrastructure/issues/1725)  
- The cluster will eventually host Argo CD, Prometheus, Grafana, and DGU itself

kentsanggds:
As we have decided to use the existing GOVUK vpc we need to increase the IP range so that it's possible to spin up many more clusters before hitting a limit.

samsimpson1:
`ephemeral.govuk.digital` has been delegated manually

samsimpson1:
Grafana DB wants to use an RDS subnet, which we aren't creating in our ephemeral VPC. It needs modifying so it creates the RDS instance in the same subnets as the cluster is using

samsimpson1:
Modified grafana DB to use cluster subnets in ephemeral envs. We should look at just using SQLite for grafana in these envs to save having to create an RDS instance

samsimpson1:
We get an error when destroying the Grafana RDS instance: `RDS Cluster final_snapshot_identifier is required when skip_final_snapshot is false`

samsimpson1:
Resources created by things running in the cluster (e.g. ELBs, Route53 records) aren't being cleaned up properly

samsimpson1:
* Terraform has been updated to support multiple EKS clusters existing in a single AWS account
[...]

PR #1797, opened 2025-03-07
Move DGU namespace role binding to DGU infra module

samsimpson1:
This fixes a dependency on the DGU namespace in cluster-services, which is applied before the module which creates the namespace


dj-maisy (approved):
👍

PR #1798, opened 2025-03-07
Search API v2: New serving configs and controls

csutter:
This adds two new serving configs:
- `raw_search` to recreate the previous raw config
- `variant_search` to back the upcoming AB test

It also creates a couple of new controls for use by the `variant_search` serving config.


hamish-pollock-gds (approved):
Looks good

PR #1801, opened 2025-03-07
Add dex and external-secrets helm releases

samsimpson1:
These used to be managed by ArgoCD


dj-maisy (approved):
🚀

PR #1805, opened 2025-03-10
Add app.kubernetes.io/part-of label to argocd dex secret

samsimpson1:
This label is required for argocd to pick up the secret


dj-maisy (approved):
👍

PR #1828, opened 2025-03-13
Properly override default maintenance window

brucebolt:
A value was added in the variables for Link Checker API in https://github.com/alphagov/govuk-infrastructure/pull/1807, but wasn't being used anywhere.


brucebolt:
Closing in favour of https://github.com/alphagov/govuk-infrastructure/pull/1829.

PR #1829, opened 2025-03-13
Fix RDS apply issue

samsimpson1:
This fixes issues introduced with this PR: https://github.com/alphagov/govuk-infrastructure/pull/1807

* setting `maintenance_window` on an RDS instance actually does something now
* removed type constraint from databases variable, so optional values are allowed


nimalank7 (approved):
LGTM!

PR #1848, opened 2025-03-18
Add retry and wait_to_retry support for Search API queue

jackbot:
This adds some new configuration to allow messages on the
`search_api_to_be_indexed` queue to be retried by the consumer.

For context, this is to support an effort to remove Sidekiq from Search
API. Currently we consume messages then enqueue them directly on Sidekiq
for processing. This is sometimes causing issues where Redis crashes if
lots of jobs are being processed at once. The aim is to remove Sidekiq
and process all jobs in the consumer directly.

However in doing this, we'll lose the automatic retry functionality that
Sidekiq provides. So instead we'll implement our own retry mechanism
using RabbitMQ only.

The idea is as follows:

* A consumer picks up a job from the `search_api_to_be_indexed` queue
for processing
* If an exception is raised, it's caught and the message is rejected
* The `search_api_to_be_indexed` queue is now configured to have a new
dead letter exchange of `search_api_to_be_indexed_retry_dlx` where
messages have a specific TTL of 30 seconds
* We have a new `search_api_to_be_indexed_wait_to_retry` queue bound to
that dlx
* Nothing is listening on the `search_api_to_be_indexed_wait_to_retry`
queue, so messages are never processed. Instead, when their TTL
expires, they're routes to another dlx,
`search_api_to_be_indexed_discarded_dlx`. This has a binding back to
the original `search_api_to_be_indexed` queue which applies to all
messages.
* Now messages are back on the `search_api_to_be_indexed` queue they're
picked up by the consumer again. The consumer can then decide whether
it wants to try to process the message again (e.g. by inspecting the
retry count)

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/amazonmq_schema.json.tpl
+++ b/terraform/deployments/govuk-publishing-infrastructure/amazonmq_schema.json.tpl
@@ -147,6 +147,31 @@
     },
     {
       "vhost": "publishing",
+      "name": "search_api_to_be_indexed_retry",
+      "pattern": "search_api_to_be_indexed",
+      "apply-to": "queues",
+      "definition": {
+        "dead-letter-exchange": "search_api_to_be_indexed_retry_dlx",
+        "ha-mode": "all",
+        "ha-sync-mode": "automatic"
+      },
+      "priority": 0
+    },
+    {
+      "vhost": "publishing",
+      "name": "search_api_to_be_indexed_wait_to_retry_discarded",
+      "pattern": "search_api_to_be_indexed_wait_to_retry",
+      "apply-to": "queues",
+      "definition": {
+        "dead-letter-exchange": "search_api_to_be_indexed_discarded_dlx",
+        "message-ttl": ${/* value in milliseconds */ "60000"}
+        "ha-mode": "all",
+        "ha-sync-mode": "automatic"
+      },
+      "priority": 0
+    },
+    {
+      "vhost": "publishing",
       "name": "ha-all",
       "pattern": ".*",
       "apply-to": "all",
@@ -181,6 +206,13 @@
     },
     {
       "name": "search_api_to_be_indexed",
+      "vhost": "publishing",
+      "durable": true,
+      "auto_delete": false,
+      "arguments": {}
+    },
+    {
+      "name": "search_api_to_be_indexed_wait_to_retry",
       "vhost": "publishing",
       "durable": true,
       "auto_delete": false,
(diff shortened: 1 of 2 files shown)


terraform/deployments/govuk-publishing-infrastructure/amazonmq_schema.json.tpl
```diff
@@ -145,6 +145,31 @@
...
+      "priority": 0
+    },
+    {
+      "vhost": "publishing",
+      "name": "search_api_to_be_indexed_wait_to_retry_discarded",
+      "pattern": "search_api_to_be_indexed_wait_to_retry",
+      "apply-to": "queues",
+      "definition": {
+        "dead-letter-exchange": "search_api_to_be_indexed_discarded_dlx",
+        "message-ttl": ${/* value in milliseconds */ "30000"}
```
jackbot, line 167: I decided not to configure this as a variable. Not sure if the GOV.UK Chat stuff used a variable just for testing the implementation but all the value are the same, so it doesn't look like it needs to be different across environments.

terraform/deployments/govuk-publishing-infrastructure/variables.tf
```diff
@@ -101,7 +101,7 @@ variable "amazonmq_host_instance_type" {
 variable "amazonmq_govuk_chat_retry_message_ttl" {
   type        = number
   default     = 300000
```
jackbot, line 103: I noticed that the variable across the board (default, then the value specified for each environment) is set to `300000` which is 5 mins. [The comment and configuration](https://github.com/alphagov/govuk-chat/blob/main/lib/tasks/message_queue.rake/#L15-L18) in the Chat repo for dev machines sets it to 30 seconds - which is `30000`. 

Not sure if this is a typo here; it feels like 5 mins is quite a long time to wait to retry.

kevindew: Yeah good shout. I did some digging and it does look like we specified 5 mins originally as I can see it in the original card: https://trello.com/c/MTNzUp0w/908-can-we-set-up-a-delay-for-rabbitmq-retries?search_id=625f021b-11a9-4f73-a2ee-26d0760cb822. 

I'm guessing I went for 30s in the repo as 5 mins was such a long wait locally. It does feel a bit arbitrary though.

terraform/deployments/govuk-publishing-infrastructure/amazonmq_schema.json.tpl
kevindew: Sounds reasonable. 

I wonder if, given some of the confusion with chat numbers we set this to be 60s so it is a clearly distinct number from GOV.UK Chat (I also imagine 60s can span the time out a bit)

kevindew (commented):
Looks great, just a suggestion on the time.

I imagine we might need to try this out in integration first to check it all works ok.

terraform/deployments/govuk-publishing-infrastructure/amazonmq_schema.json.tpl
jackbot: Sounds good, it's set to 60s now.

kevindew (approved):
Great, looks good to me

PR #1858, opened 2025-03-19
Update AmazonMQ instance type for integration publishing

jackbot:
The current instance for the publishing broker on integration is
`mq.m5.large`, but the configuration here shows `mq.t3.micro`.

Trying to apply a TF plan that touches the AMQ instance fails with
the following:

```
Error: updating MQ Broker host instance type: operation error mq:
UpdateBroker, https response error StatusCode: 400
BadRequestException: Downgrading from current instance type [mq.m5.large] to requested
instance type [mq.t3.micro] is not supported for RabbitMQ brokers.
```

I think maybe this was just a typo from when things were moved into this
repo from `govuk-aws`.


kevindew (approved):
Looks good, I agree this seems like it should fix it.

PR #1864, opened 2025-03-19
Add missing commas in AMQ schema

jackbot:


kevindew (approved):
Nice one

kevindew:
Merging despite failing CI as we think running with the invalid JSON is at risk of breaking RabbitMQ and that by fixing this it might help lead to unblocking the problems with the plan.

PR #1876, opened 2025-03-21
Search API v2: Remove spring statement synonym

csutter:
It turns out we can't use `queryTerms` for synonyms controls, so removing this as it causes an error on apply.


hamish-pollock-gds (approved):
😢

Issue #1746, opened 2025-03-03
♻️ Update DGU Infra and Helm chart for ephemeral environment


jasonBirchall:
**User Need**

**As a** platform engineer  
**I want** to modify the DGU Helm chart in a dedicated branch  
**so that** it can be deployed automatically to the ephemeral cluster with minimal manual intervention

**Functional Requirements (What):**

- [x] Update the [Helm chart](https://github.com/alphagov/govuk-dgu-charts) for the ephemeral DGU deployment  
- [x] Ensure all environment-specific values (e.g. resource names, secrets references) are parameterised  

**Non-Functional Requirements (How):**

- [x] Ensure secret values are not hard-coded—only placeholders or references to external secrets management  
- [x] Maintain compatibility with the existing deployment structure (i.e. do not break integration/staging/prod charts)

**Acceptance Criteria:**

- [x] A branch with updated Helm chart values/configuration for ephemeral DGU deployment is available  
- [x] Documentation is updated to reflect ephemeral-specific values or references that differ from the main GOV.UK environment  

**Notes:**

- Linked to epic [#1725](https://github.com/alphagov/govuk-infrastructure/issues/1725)  
- Coordinate with the “Prevent naming collisions” and “Identify and reference non-app secrets” stories to ensure consistency

kentsanggds:
Seems like the postgres database that we are trying to spin up is not available as arm64 - https://github.com/alphagov/ckanext-datagovuk/pkgs/container/postgis.

I tried to set the arch to amd64, the arch for the build image, but still no go.

kentsanggds:
Had to set the UID for ckan-postgres to 999 to get it to run with the correct access permissions. https://github.com/alphagov/govuk-dgu-charts/pull/504

kentsanggds:
Haven't built the image for postgis, trying out the postgres 16 image for arm64 instead and it appears to be working but need to test spatial harvests against it.

kentsanggds:
In order to get the cluster up I had to manually set the environment variable in argo for ckan and also update the bucket reference to just use govuk-ephemeral-aws-logging rather than something like govuk-eph-abc123-ephemeral-aws-logging

kentsanggds:
I tried to refactor the changes and hit some problems, it seems like its not easy to pass on the environment var from argo-bootstrap to the app and I've had to make postgres not persistent because there are permission errors writing to a directory with pre-existing folders created before.

kentsanggds:
[...]

PR #1887, opened 2025-03-24
Import DGU organogram S3 settings into Terraform

nimalank7:
Description:
- Enabled ACLs and S3 block public access settings were clickopsed and not managed by Terraform
- As part of https://github.com/alphagov/govuk-infrastructure/issues/1746


kentsanggds (approved):
LGTM

PR #1904, opened 2025-03-27
Tighten up pattern regexes in AMQ policies

jackbot:
RabbitMQ policies are defined by applying them to a specific pattern. We
generally use them for applying to queues, and so the pattern matches
against the queue name.

Using the queue name as the pattern has the potential for matching
unintended queues. This has bitten us in the past and is a bit tricky to
figure out.

So this commit tightens up all patterns where we're using the string as
a direct match against the queue name, and ensures that the policy will
only match queues with exactly those names.


rtrinque (approved):
Looks good

PR #1905, opened 2025-03-27
Add AMQ configuration for search-api retry queues

jackbot:
Search API has 3 queues it listens on[1], which all need the same retry
configuration set up:

- `search_api_to_be_indexed`
- `search_api_bulk_reindex`
- `search_api_govuk_index`

We already have the config in place for the `search_api_to_be_indexed`
queue, this commit adds all the necessary bits for the remaining two.

[1] https://github.com/alphagov/search-api/blob/8b02d8ed26d6e918a4bc43e7d3585d8def43a655/lib/tasks/message_queue.rake


rtrinque (approved):
Looks good

Issue #1867, opened 2025-03-19
👯 Separate Logging Resources from VPC Module for Ephemeral Kubernetes Cluster Creation


jasonBirchall:
**User Need**

**As a** Platform Engineer
**I want to** extract logging resources (CloudWatch Logs, VPC Flow Logs) into a dedicated Terraform module (vpc-logging) separate from the shared VPC module
**So that I can** provision ephemeral Kubernetes clusters with their own isolated VPC without impacting production, staging, or integration environments.

**Functional Requirements (What):**
- [ ] Create a new Terraform module (terraform/deployments/vpc-logging) containing:

   - AWS CloudWatch logging configurations (aws_logging.tf)
   - VPC flow logs (flow_log.tf)
   - Google logging bucket definitions (google_logging_bucket.tf)

- [ ] Remove logging-related resources from the current VPC module (terraform/deployments/vpc/):

   - aws_logging.tf
   - flow_log.tf
   - google_logging_bucket.tf

- [ ] Configure Terraform Cloud to manage logging infrastructure separately via a dedicated workspace (vpc-logging).

- [ ] Safely perform Terraform state moves (terraform state mv) to transition logging resources from the VPC module state to the new logging module without downtime.

**Non-Functional Requirements (How):**

- [ ] Perform the migration using Terraform state commands (terraform state mv) explicitly to ensure resources are not destroyed or recreated unintentionally.
- [ ] Update documentation clearly describing the new separate management approach for logging resources.
- [ ] Validate all Terraform plans (terraform plan) thoroughly prior to applying any changes, ensuring zero unintended resource deletions.

**Acceptance Criteria:**

- [ ] Logging resources (CloudWatch, Flow logs, logging buckets) are successfully moved to the new module (terraform/deployments/vpc-logging) without downtime or disruption.
- [ ] Production, staging, and integration environments remain fully operational, unaffected by these changes.
- [ ] Terraform plans executed after migration result in no resource destruction.
- [ ] Ephemeral Kubernetes clusters can now provision isolated, dynamically-calculated VPCs without inheriting or conflicting with shared logging resources.

**Notes:**

- It's crucial to verify Terraform state carefully before and after migration to prevent unintended resource impact.

- Clear documentation of the new logging module will help your team effectively manage logging resources moving forward.

PR #1907, opened 2025-03-28
Remove logging resources from vpc module, and skeleton for new logging module

samsimpson1:
Logging resources will be added to the new logging module after the TFC workspaces for it have been created

https://github.com/alphagov/govuk-infrastructure/issues/1867


terraform/deployments/logging/main.tf
```diff
@@ -0,0 +1,40 @@
+terraform {
+  cloud {
+    organization = "govuk"
+    workspaces {
+      tags = ["logging", "aws"]
+    }
+  }
+  required_version = "~> 1.10"
+  required_providers {
```
dj-maisy, line 9: We don't also need to add a provider for google, do we?

dj-maisy (approved):
One comment, all good otherwise. 👍

terraform/deployments/logging/main.tf
samsimpson1: The logging S3 bucket is mirrored to a GCP bucket, so we do need it

PR #1921, opened 2025-04-01
Set argoNamespace when installing ephemeral bootstrap chart

samsimpson1:
This being missing is causing some charts to be installed to the wrong namespace


nimalank7 (approved):
LGTM!

PR #1923, opened 2025-04-01
Add content-block-editor to repos.yml

pezholio:
This is a protoype JS library to demonstrate how we could highlight content blocks within publishing apps.

### DIFF
--- a/terraform/deployments/github/repos.yml
+++ b/terraform/deployments/github/repos.yml
@@ -87,6 +87,13 @@ repos:
         - Lint SCSS / Run Stylelint
         - Test Ruby / Run RSpec
 
+  content-block-editor:
+    required_status_checks:
+      standard_contexts: *standard_security_checks
+      additional_contexts:
+        - Playwright / Run Tests
+        - Vitest / Run Tests
+
   content-data-admin:
     homepage_url: "https://docs.publishing.service.gov.uk/apps/content-data-admin.html"
     required_status_checks:


terraform/deployments/github/repos.yml
```diff
@@ -87,6 +87,12 @@ repos:
         - Lint SCSS / Run Stylelint
         - Test Ruby / Run RSpec
 
+  content-block-editor:
+    required_status_checks:
+      standard_contexts: *standard_security_checks
+      additional_contexts:
+        - CI
```
AgaDufrat, line 94: This is not going to work since it's a workflow name, while it should be job name. 

I suggest adding job names in https://github.com/alphagov/whitehall/tree/main/.github/workflows to use them here to configure branch protection rules (alternatively you can get ids). For workflows that use reusable workflows, the pattern is <initial_workflow.jobs.job.[name/id]> / <reused-workflow.jobs.job.[name/id]>.

There's a lot of examples in other repos.

pezholio: Sorted now - see https://github.com/alphagov/content-block-editor/blob/4e0d7e1e8a5b01f5c641a16dfe63107082fd9c7a/.github/workflows/ci.yml#L30 / https://github.com/alphagov/content-block-editor/blob/4e0d7e1e8a5b01f5c641a16dfe63107082fd9c7a/.github/workflows/playwright.yml#L7 and https://github.com/alphagov/content-block-editor/blob/4e0d7e1e8a5b01f5c641a16dfe63107082fd9c7a/.github/workflows/ci.yml#L37 / https://github.com/alphagov/content-block-editor/blob/4e0d7e1e8a5b01f5c641a16dfe63107082fd9c7a/.github/workflows/vitest.yml#L7 👍

AgaDufrat: Thanks! This should do the trick.

Issue #1891, opened 2025-03-25
Alert when a cluster's node limit is reached


nicholsj:
**User Need**

**As a** platform engineer
**I want** to know when a cluster's node limit is reached, and how to increase that limit
**so that** I can investigate usage and potentially increase the limit

**Functional Requirements (What):**

- [x] Implement or extend an alerting mechanism so that a Slack message is posted to the Platform Engineering team when the node limit of a cluster is reached
- [x] Include relevant details in the alert message
- [ ] Add instructions for checking and increasing node limits to [this page](https://docs.publishing.service.gov.uk/kubernetes/manage-app/scale-app/)
- [ ] Link from the alert message to the instructions for increasing max node limit

**Acceptance Criteria:**

- [ ] When the node limit of a cluster is reached, a Slack message is sent to #govuk-platform-support
- [ ] The message contains enough information to enable investigation, including a link to an appropriate runbook
- [ ] Instructions on how to check and increase node limits are included on [this page](https://docs.publishing.service.gov.uk/kubernetes/manage-app/scale-app/) 

**Notes:**

This is an action arising from a review of [this incident](https://docs.google.com/document/d/1GsLZTyuFBaAyt-VNFk2guV2XIzCoui1KNfe2_gN0lX4/edit), where a cluster reaching its maximum allowed number of nodes caused issues. 

A member of Platform Engineering needed to investigate it and increase the limit.

samsimpson1:
A message will now be sent to #govuk-platform-support if a cluster is about to hit its node limit. It won't link to any relevant documentation, since this is using a basic SNS email subscription which doesn't have any sort of customisability.

PR #1925, opened 2025-04-01
Add Slack alert for when an EKS cluster is about to hit its node limit

samsimpson1:
This adds a CloudWatch alarm which sends an email to the #govuk-platform-support Slack channel when any of the node groups in an EKS cluster are about to hit their node limit.

The email address is taken from a manually managed secret in Secrets Manager, since we don't really want to make it public so it doesn't get spammed.

At the moment, it alerts when a node group size is 1 less than the maximum to allow for action before no more nodes are available.

The change has been tested in integration.

https://github.com/alphagov/govuk-infrastructure/issues/1891

### DIFF
--- a/terraform/deployments/cluster-infrastructure/alarms.tf
+++ b/terraform/deployments/cluster-infrastructure/alarms.tf
@@ -1,0 +1,48 @@
+data "aws_secretsmanager_secret" "slack_email" {
+  count = startswith(var.govuk_environment, "eph-") ? 0 : 1
+  name  = "govuk/slack/platform-support-email"
+}
+
+data "aws_secretsmanager_secret_version" "slack_email" {
+  count     = startswith(var.govuk_environment, "eph-") ? 0 : 1
+  secret_id = data.aws_secretsmanager_secret.slack_email[0].id
+}
+
+resource "aws_sns_topic" "slack_channel" {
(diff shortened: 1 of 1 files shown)


terraform/deployments/cluster-infrastructure/alarms.tf
```diff
@@ -0,0 +1,48 @@
...
+
+  alarm_actions = [aws_sns_topic.slack_channel.arn]
+
+  metric_query {
+    id          = "count"
+    return_data = true
+    metric {
+      metric_name = "GroupDesiredCapacity"
+      namespace   = "AWS/AutoScaling"
+      period      = 60
```
sihugh, line 40: I think the last node exhaustion took quite a while, so a 2 hour reaction period feels OK? Do we want an additional quicker one for when we're actually _at_ the maximum just in case we blow through it? (Maybe a discussion) for another time

sihugh (approved):
Looks great, just a question about the responsiveness, but not a blocker

terraform/deployments/cluster-infrastructure/alarms.tf
samsimpson1: The period is in seconds, so it'd alert if the node group is at 1 less than the max for 2 minutes

Issue #1915, opened 2025-03-31
😷 Extract root DNS zone configuration from VPC module into reusable root-dns module


jasonBirchall:
**User Need**

**As a** platform engineer  
**I want** to extract the root DNS zone configuration from the shared VPC module into a dedicated Terraform module  
**so that** I can allow multiple ephemeral Kubernetes clusters to provision and manage their own DNS zones independently and concurrently

**Functional Requirements (What):**

- [ ] Create a new Terraform module (`terraform/deployments/root-dns`) to contain:
  - `aws_route53_zone.internal_zone`
  - `aws_route53_zone.external_zone`
- [ ] Refactor the shared VPC module to remove the `root_dns_zones.tf` resource definitions
- [ ] Update any existing references to use the new `root-dns` module

**Non-Functional Requirements (How):**

- [ ] Ensure the module is usable by multiple ephemeral environments concurrently without naming conflicts
- [ ] Confirm the module can be called independently from the VPC module, and supports dynamic linking of VPC IDs from ephemeral clusters
- [ ] Maintain backwards compatibility for any current environments until full migration is complete

**Acceptance Criteria:**

- [ ] `root_dns_zones.tf` has been successfully removed from the shared VPC module and added to a new dedicated `root-dns` module
- [ ] All module variables are clearly documented, and the `terraform plan` output shows no unintended deletions or conflicts
- [ ] The new module supports concurrent usage by multiple ephemeral environments with isolated Route53 zones

**Notes:**

- This ticket builds on the previous work in #1867, which extracted logging resources from the shared VPC module to enable safe provisioning of ephemeral Kubernetes clusters.
- This work unblocks the completion of #1745  (ephemeral cluster creation validation) by allowing ephemeral clusters to provision and manage DNS zones independently of the shared VPC.
- The original DNS configuration currently resides in `root_dns_zones.tf` within the VPC module:
  ```hcl
  resource "aws_route53_zone" "internal_zone" {
    name = "${var.govuk_environment}.govuk-internal.digital."
    vpc {
      vpc_id = aws_vpc.vpc.id
    }
  }

  resource "aws_route53_zone" "external_zone" {
    name = "${var.govuk_environment}.govuk.digital."
  }

PR #1971, opened 2025-04-08
Fix correct root dns workspace name for ephemeral

nimalank7:
Description:
- Workspace name is `root-dns-ephemeral`
- As part of https://github.com/alphagov/govuk-infrastructure/issues/1915


AgaDufrat (approved):

AgaDufrat:
Are you sure this is needed? I'd expect it to work ok as it was because `root-dns-ephemeral` workspace has `govuk_environment` variable set to "ephemeral"

Issue #1744, opened 2025-03-03
🚀 Install Prometheus and Grafana for monitoring ephemeral cluster


jasonBirchall:
**User Need**

**As a** platform engineer  
**I want** to deploy Prometheus and Grafana on an ephemeral Kubernetes cluster  
**so that** we can monitor resources, gather metrics, and diagnose issues

**Functional Requirements (What):**

- [x] Deploy Prometheus using Helm charts or equivalent method  
- [x] Deploy Grafana and configure data sources to point to Prometheus  
- [x] Set up basic dashboards for cluster health and application metrics

**Non-Functional Requirements (How):**

- [x] Respect ephemeral environment constraints (clean up any manually persisted data on teardown)  
- [x] Adhere to GOV.UK operational standards (logging, naming, security)  
- [x] Make dashboards accessible to the platform team with minimal friction

**Acceptance Criteria:**

- [x] Prometheus successfully scrapes cluster metrics and any pre-configured app metrics  
- [x] Grafana is reachable via ingress, and at least one default dashboard is operational  
- [x] Documentation exists for how to install, configure, and verify both Prometheus and Grafana

**Notes:**

- This user story is linked to the epic: [#1725](https://github.com/alphagov/govuk-infrastructure/issues/1725)  
- Any service or volume naming should follow the conventions established in the “Prevent naming collisions” story

dj-maisy:
Status update:

* Prometheus is working.
* Grafana exists and is responding to HTTPS, but authentication (OAuth) is not yet configured.

jasonBirchall:
We've had to put this back in TODO due to the mandated work on AWS profiles.

jasonBirchall:
Please check in with @samsimpson1 and @dj-maisy to get an idea of what's remaining.

kentsanggds:
Pairing with Nim on this, we are trying to understand how Grafana works at the moment. It is not going through the dex login, it is instead presenting its own login page, so there is probably a mis-configuration in the charts.

kentsanggds:
https://grafana.com/docs/grafana/latest/setup-grafana/configure-security/configure-authentication/generic-oauth/#configuration-options 

The `grafana.ini` file doesn't appear to have the client id or secret stored but it is listed as a requirement so this might be causing an issue.

kentsanggds:
Next steps to complete this card:

- set up the altertmanager config, as theres nothing there in the ephemeral environment
[...]

PR #1982, opened 2025-04-10
Add moved blocks for grafana RDS database

samsimpson1:
TF tries to destroy and recreate these resources without these moved blocks

https://github.com/alphagov/govuk-infrastructure/issues/1744


nimalank7 (approved):
LGTM!

Issue #1833, opened 2025-03-14
Create `govuk-developer` Role with PowerUser-like Permissions


jasonBirchall:
**User Need**

**As a** Platform Engineer  
**I want** to create a `govuk-developer` AWS IAM role granting baseline access for day-to-day activities
**so that** developers can perform daily tasks without requiring full administrative privileges

**Functional Requirements (What):**

- [x] Define the `govuk-developer` IAM role in code.
- [ ] ~~Include permissions mimicking [PowerUserAccess](https://docs.aws.amazon.com/aws-managed-policy/latest/reference/PowerUserAccess.html)~~.
- [x] Identify one or two volunteer developers and have them assume the role for two weeks.  

**Non-Functional Requirements (How):**

- [x] Review and merge changes through the standard pull request workflow.  

**Acceptance Criteria:**

- [x] The `govuk-developer` role is created in AWS and is usable by one or two developers.  
- [x] Selected testers confirm they can perform their usual tasks (Kubernetes auth, RDS tweaks, etc.) without disruption.  
- [x] Any required policy adjustments from the test period are identified and documented.  
- [ ] TF Auto-apply is re-enabled after the Permissions changes are tested and completed.

**Notes:**
- The new role must exclude direct IAM/Organisations/Account actions while maintaining enough permissions for day-to-day tasks.  
- After the two-week trial, we will gather feedback and decide on broader rollout. 
- connects to https://github.com/alphagov/govuk-infrastructure/issues/1757

dj-maisy:
Runlog:

* We've had a think about permissions access - most developers really should be fine with `ReadOnly` but we're not 100% sure.
* We'd like to set up our own CloudTrail so we can audit current usage of AWS permissions and identify common actions that `ReadOnly` won't satisfy.

dj-maisy:
Where is this?

* We had to deal with deprecation issues with our existing govuk-user-reviewer Terraform
* We think we've fixed those issues and are ready to test the new roles (and all relevant changes) in govuk-test first.

nicholsj:
@dj-maisy there's an action from yesterday's incident review to ensure that these AWS role changes will still allow people to raise a ticket (as per https://docs.publishing.service.gov.uk/manual/how-to-escalate-to-AWS-support.html)

dj-maisy:
[...]

PR #2002, opened 2025-04-14
Manage developer role for datagovuk namespace in datagovuk-infrastructure instead of cluster-services

samsimpson1:
This resolves a dependency where the role needs the namespace to exist, but it is created in datagovuk-infrastructure which is applied after cluster-services

https://github.com/alphagov/govuk-infrastructure/issues/1833


dj-maisy (dismissed):
🚀

dj-maisy (approved):

PR #2027, opened 2025-04-17
Increase threshold for content-data-api db

nicholsj:
This increases the threshold so it warns when there is only 30% free (from 50%), to reduce noisy unhelpful alarms. A short term fix while we explore something better.

### DIFF
--- a/terraform/deployments/tfc-configuration/variables-production.tf
+++ b/terraform/deployments/tfc-configuration/variables-production.tf
@@ -307,7 +307,7 @@ module "variable-set-rds-production" {
         allocated_storage            = 1024
         instance_class               = "db.m6g.large"
         performance_insights_enabled = false
-        freestoragespace_threshold   = 536870912000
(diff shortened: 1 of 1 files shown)


terraform/deployments/tfc-configuration/variables-production.tf
```diff
@@ -307,7 +307,7 @@ module "variable-set-rds-production" {
         allocated_storage            = 1024
         instance_class               = "db.m6g.large"
         performance_insights_enabled = false
-        freestoragespace_threshold   = 536870912000
+        freestoragespace_threshold   = 716800000000
```
samsimpson1, line 310: It alerts on how much free space is left, so this would alert if there is less than ~700GB remaining (30% usage)

nicholsj: Oops thanks. Fixed

samsimpson1 (approved):

Issue #2028, opened 2025-04-17
Alert based on a percentage of disk space used instead of an absolute value for RDS instances


samsimpson1:
**As an** unwilling database admin
**I want** to alert on the percentage of disk space used on RDS instances instead of an absolute value
**so that** it is easier to understand the alert threshold and alerts are more useful

**Functional Requirements (What):**

- [x] Update the `rds` Terraform module to use a percentage for the `freespace_threshold` value
- [x] Provide a sensible default for this value

**Acceptance Criteria:**

- [x] I can set the free space alert threshold as a percentage instead of as an absolute number of bytes
- [x] I don't have to set the threshold in most cases (default is good enough)

**Notes:**

samsimpson1:
`freespace_threshold` has been replaced with `storage_alarm_threshold_percentage`, and the default has been set to 10%. All instances are using the default for now, and tuning is possible in the future if necessary.

PR #2029, opened 2025-04-17
Replace freestoragespace_threshold RDS value with storage_alarm_threshold_percentage

samsimpson1:
This also sets the threshold to 10% by default for all RDS instances

https://github.com/alphagov/govuk-infrastructure/issues/2028


jasonBirchall (approved):
Niiice

Issue #1972, opened 2025-04-08
Ensure the ephemeral cluster Terraform can be destroyed cleanly


samsimpson1:
**As a** platform engineer
**I want** to ensure the ephemeral Terraform config can be destroyed cleanly
**so that** ephemeral clusters can be destroyed without manual intervention

**Functional Requirements (What):**

- [ ] Ensure any Terraform resources that depend on others are destroyed before their dependencies are
- [ ] Ensure deletions are cascading (such as ArgoCD applications) so that child resources are also destroyed correctly

**Non-Functional Requirements (How):**

- [ ] Add `depends_on` to resources that are being destroyed out of order
- [ ] Split our Terraform modules into more 'stages' if necessary
- [ ] Make sure ArgoCD is destroying all resources when an `Application` resource is deleted on K8s
- [ ] Make sure these deletions are complete before things like the `aws-ebs-csi-driver` and `aws-load-balancer-controller` are uninstalled

**Acceptance Criteria:**

- [ ] Running `terraform apply -destroy` on an ephemeral module completes without error
- [ ] No resources are left lingering after the destroy has completed

**Notes:**

Related: https://github.com/alphagov/govuk-infrastructure/issues/1969

Doing this may require a rethink of how the ephemeral TF module works. It may require a script to manage Terraform apply runs to make sure everything is done in the right order.

samsimpson1:
* We now have a script, `shutdown.sh` which handles shutting down ephemeral clusters cleanly
* The README in the `ephemeral` module has been updated to reflect the new shutdown process

PR #2041, opened 2025-04-22
Add shutdown script for ephemeral environments

samsimpson1:
This is a script to automate ephemeral cluster shutdown.

It has three steps:
* Deleting ArgoCD applications that were installed via Helm (Terraform)
* Uninstalling all Helm charts
* Running Terraform destroys on workspaces in the correct order

https://github.com/alphagov/govuk-infrastructure/issues/1972


terraform/deployments/cluster-services/main.tf
```diff
@@ -10,7 +10,7 @@ terraform {
   cloud {
     organization = "govuk"
     workspaces {
-      tags = ["cluster-services", "eks", "aws"]
+      tags = ["cluster-services", "aws"]
```
samsimpson1, line 13: Not sure how this change made it in, but it fixes a small issue where the Terraform CLI doesn't work for this workspace in ephemeral environments (tags aren't set correctly)

PR #2058, opened 2025-04-28
Add Content Modelling E2E repo

pezholio:
This adds the new https://github.com/alphagov/content-modelling-e2e repo, which contains a more extensive suite of e2e tests for content modelling, which can be run on demand.


terraform/deployments/github/repos.yml
```diff
@@ -110,6 +110,13 @@ repos:
       additional_contexts:
         - Test Ruby / Run RSpec
 
+  content-modelling-e2e:
+    required_status_checks:
+      standard_contexts: *standard_security_checks
+      additional_contexts:
+        - Lint / Eslint
+        - Prettier / Check
+
```
nimalank7, line 119: Do you want to run the e2e tests as part of this repo?

nimalank7 (changes requested):

terraform/deployments/github/repos.yml
pezholio: I did consider it, but I think we want to keep them only running on demand.

nimalank7: Sure but how do you prevent someone merging in a bad test?

pezholio: Ah, that's fair. Have added 👍

nimalank7 (approved):
LGTM!

PR #2066, opened 2025-04-29
Search API v2: Add user events purge permission

csutter:
This adds the purge permission for user events to the Search API v2 GCP service user to allow it to purge events at the end of the retention period.


ianhowell-gds (approved):
Looks ok

PR #2067, opened 2025-04-29
Search API v2: Allow Search Admin to manage denylist

csutter:
This adds the required permissions to the role used by Search Admin's service account for it to be able to manage the autocomplete denylist.


ianhowell-gds (approved):
LGTM

Issue #1765, opened 2025-03-05
🛑 Restrict Containers Running With High Privileges


jasonBirchall:
Connects to https://github.com/alphagov/govuk-platform-internal/issues/18

jasonBirchall:
We're fairly certain was already done, but just so we can record it.

dj-maisy:
Action:

* https://github.com/alphagov/govuk-infrastructure/pull/2079

dj-maisy:
See connected ticket - all containers now checked and configured to run without elevation, or signed-off as accepted to require elevation.

PR #2079, opened 2025-04-30
Harden security context for filebeat DaemonSet

dj-maisy:
## What?
This adds security context configuration to the Filebeat Daemonsets running on our clusters. This should allow Filebeat to run with the necessary security scope without continuing to need to run as user 0 (root).

We will want to test this on Integration first to make sure that filebeat can continue to operate normally.

We will also want to replace this with the eck-beat chart in the future, as the Chart we are currently using has been deprecated.

This also adds the `DAC_READ_SEARCH` Capability which would allow filebeat to override read and execute permissions, which it kind of needs to in order to be able to correctly read logs. This is not as permissive as giving it `DAC_OVERRIDE` or root, which would be even more permissive.

### Related Resources
Link to the relevant chart:

* https://github.com/elastic/helm-charts/tree/main/filebeat

### DIFF
--- a/terraform/deployments/cluster-services/logging.tf
+++ b/terraform/deployments/cluster-services/logging.tf
@@ -26,6 +26,9 @@ resource "helm_release" "filebeat" {
           add  = ["DAC_READ_SEARCH"]
           drop = ["ALL"]
         }
+        seccompProfile = {
+          type = "RuntimeDefault"
+        }
       }
       secretMounts = []
       tolerations = [{ # TODO: remove once we no longer run amd64 (Intel) nodes.
--- a/terraform/deployments/cluster-services/logging.tf
+++ b/terraform/deployments/cluster-services/logging.tf
@@ -20,8 +20,9 @@ resource "helm_release" "filebeat" {
         allowPrivilegeEscalation = false
         privileged               = false
         runAsUser                = 1000
-        runAsNonRoot             = true
         runAsGroup               = 1000
+        fsGroup                  = 1000
+        runAsNonRoot             = true
         capabilities = {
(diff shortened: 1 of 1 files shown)


terraform/deployments/cluster-services/logging.tf
```diff
@@ -16,6 +16,17 @@ resource "helm_release" "filebeat" {
...
+      securityContext = {
+        allowPrivilegeEscalation = false
+        privileged               = false
+        runAsUser                = 1000
+        runAsNonRoot             = true
+        runAsGroup               = 1000
+        capabilities = {
+          add  = ["DAC_READ_SEARCH"]
+          drop = ["ALL"]
+        }
```
nimalank7, line 28: You'll need `seccompProfile` as well

nimalank7 (changes requested):

terraform/deployments/cluster-services/logging.tf
```diff
@@ -16,6 +16,20 @@ resource "helm_release" "filebeat" {
   timeout          = var.helm_timeout_seconds
   values = [yamlencode({
     daemonset = {
+      securityContext = {
+        allowPrivilegeEscalation = false
+        privileged               = false
+        runAsUser                = 1000
+        runAsNonRoot             = true
+        runAsGroup               = 1000
```
nimalank7, line 24: Probably add `fsGroup: 1000` to ensure any directories and files are owned by the same user you're running as

nimalank7 (approved):
LGTM!

jasonBirchall:
connects to https://github.com/alphagov/govuk-infrastructure/issues/1765

Issue #1745, opened 2025-03-03
🧪 Validate ephemeral cluster functionality


jasonBirchall:
**As a** platform engineer
**I want** to create an ephemeral Kubernetes cluster in the AWS test account independently  
**so that** I can validate the reproducibility and reliability of the infrastructure code from the previous work

## **Functional Requirements (What)**

- [x] Use the existing Infrastructure as Code (IaC) (Terraform or equivalent) to provision the Kubernetes cluster.  
- [x] Access ArgoCD, Grafana, Prometheus and DGU to confirm the cluster is functional.

## **Non-Functional Requirements (How)**

- [x] Ensure all steps follow the documented procedure from the previous ticket (no hidden manual steps).  
- [x] Confirm that the cluster naming, tagging, and logging adhere to ephemeral environment standards.

## **Acceptance Criteria**

- [x] **Cluster Creation:** I can run the provided IaC scripts and end up with a fully running Kubernetes cluster in the AWS sandbox.  
- [x] **Connectivity & Basic Checks:** I can use `kubectl` and see the cluster’s nodes, pods, etc. A basic “Hello World” deployment is successful.  
- [x] **Documentation Validation:** The instructions for creating and tearing down the cluster are clear, complete, and match my actual experience.  
- [x] **Ephemeral Requirements:** Tagging, naming, and logging comply with ephemeral conventions—no long-lived resources left over.

## **Notes**

- This directly builds on the work from ticket [[#1742](https://github.com/alphagov/govuk-infrastructure/issues/1742)](https://github.com/alphagov/govuk-infrastructure/issues/1742).  
- The outcome should demonstrate that any engineer can replicate the ephemeral cluster setup without special permissions or undocumented steps.  
- Consider capturing logs or screenshots of key stages to confirm reproducibility.

jasonBirchall:
Running into a small problem. https://gds.slack.com/archives/C079ULB1UEP/p1742382539484119

jasonBirchall:
## ⚠️ **Ticket Update: Currently Blocked**

We are currently blocked on progressing this ticket until we complete the work defined in the related ticket (separation of logging resources from the shared VPC module).

**Reasoning:**  
The current VPC module includes logging configurations used across all environments (production, staging, integration). Attempting to provision an ephemeral Kubernetes cluster with its own isolated VPC directly from this module risks unintended changes or conflicts with shared resources, particularly logging (CloudWatch logs, flow logs).

[...]

PR #2083, opened 2025-05-01
Create VPCs for ephemeral clusters as part of cluster start up

samsimpson1:
We were using a single VPC for all ephemeral clusters, which was causing issues when trying to run multiple clusters at once due to IP ranges clashing

This was meant to be done once [this issue](https://github.com/alphagov/govuk-infrastructure/issues/1867) was complete, but was missed.

https://github.com/alphagov/govuk-infrastructure/issues/1745


samsimpson1:
The failing tests are not related to the changes here, they're from workspaces that are due to be destroyed and are not managing any resources.

dj-maisy (approved):

PR #2085, opened 2025-05-01
Harden Dex Pods

dj-maisy:
## What?
This now attempts to harden the Dex Pods in the cluster to not use root and to prevent privilege escalation. It also shouldn't need any capabilities (unless we find things don't work).

### Related

* https://github.com/alphagov/govuk-platform-internal/issues/18


samsimpson1 (approved):

dj-maisy:
Logs from Deployment Attempt:
```
error: cannot create temp file: open /tmp/dex.config.yaml-2385404921: read-only file system
```

PR #2092, opened 2025-05-06
Document our policy on using non-GOV.UK domains

ChrisBAshton:
This is [copied over from govuk-aws](https://github.com/alphagov/govuk-aws/blob/main/docs/architecture/decisions/0039-non-govuk-domain-policy.md) (and therefore optimistically marked as "Status: Accepted"), minus the 'Consequences' section which is no longer relevant.

This was one of the few ADRs that was worth carrying over from govuk-aws, so that it can be automatically pulled into govuk-developer-docs and therefore have a canonical URL we can direct people towards to document the dangers of using non-GOV.UK URLs for hosting GOV.UK-like content.

### DIFF
--- a/docs/architecture/decisions/0012-non-govuk-domain-policy.md
+++ b/docs/architecture/decisions/0012-non-govuk-domain-policy.md
@@ -1,0 +1,38 @@
+# 12. Non-GOV.UK domain policy
+
+Date: 2025-05-06
+
+## Status
+
+Accepted
+
+## Definitions
+
+This proposal uses the [RFC2119](https://www.ietf.org/rfc/rfc2119.txt) standard for definitions of MUST, MUST NOT and MAY.
+
+Definitions:
+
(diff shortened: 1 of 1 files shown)


docs/architecture/decisions/0012-non-govuk-domain-policy.md
```diff
@@ -0,0 +1,38 @@
+# 12. Non-GOV.UK domain policy
+
+Date: 2025-05-06
```
nicholsj, line 3: Shouldn't this be the original date of 2022-10-10 ?

nicholsj (approved):
LGTM - one comment but not blocking

docs/architecture/decisions/0012-non-govuk-domain-policy.md
ChrisBAshton: I changed it as an opportunity to re-review the doc with today's date. Thought it might throw things off a bit if these ADRs aren't in date order. I'm happy leaving it as today's date if you are 👍

PR #2135, opened 2025-05-12
Upsize and Upgrade the Staging Pub API RDS.

dj-maisy:
## What?
This changes the Publishing API RDS Instance in Staging to use a larger and newer Instance size, as it appears to have outgrown its current instance size.

### Changes

* From `db.m6g.large` to `db.m7g.2xlarge`


nicholsj (approved):
LGTM thanks

richardTowers (approved):
Not sure if the instance has "apply immediately" set, or if it will wait for the maintenance period, but it doesn't actually matter really as (a) it's staging and (b) the instance is broken in the mornings anyway.

PR #2141, opened 2025-05-14
Add select-with-search-component repository

minhngocd:
New repository to host the select with search UI component that will be shared across a few publishing apps

https://trello.com/c/13PXF8wR/3687-design-system-move-selectwithsearch-component-somewhere-shared

### DIFF
--- a/terraform/deployments/github/import.tf
+++ b/terraform/deployments/github/import.tf
@@ -1,4 +1,4 @@
 import {
-  to = github_repository.govuk_repos["govuk-content-publishing-guidance"]
-  id = "govuk-content-publishing-guidance"
+  to = github_repository.govuk_repos["select-with-search-component"]
+  id = "select-with-search-component"
 }
--- a/terraform/deployments/github/repos.yml
+++ b/terraform/deployments/github/repos.yml
@@ -630,6 +630,13 @@ repos:
 
   search-v2-evaluator:
 
+  select-with-search-component:
+    homepage_url: "https://docs.publishing.service.gov.uk/repos/select-with-search-component.html"
+    required_status_checks:
+      standard_contexts: *standard_security_checks
+      additional_contexts:
+        - test
+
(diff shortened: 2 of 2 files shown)


terraform/deployments/github/repos.yml
```diff
@@ -630,6 +630,13 @@ repos:
 
   search-v2-evaluator:
 
+  select-with-search-component:
```
AgaDufrat, line 633: Terraform will try to create this repo but raise errors because it already exists. 

We need to import it into terraform state. Example code change: https://github.com/alphagov/govuk-infrastructure/commit/e5056f2193fd3e03954ee3d7bb325fc18f30c79f
You can update the current block (change repo name), as it's only used once.

terraform/deployments/github/repos.yml
```diff
@@ -630,6 +630,13 @@ repos:
...
+    homepage_url: "https://github.com/alphagov/select-with-search-component/"
```
AgaDufrat, line 634: Nitpicking

```suggestion
    homepage_url: "https://docs.publishing.service.gov.uk/repos/select-with-search-component.html"
```

minhngocd:
Thank you for the review @AgaDufrat 🙏 
I've updated the PR - is this what you mean?

terraform/deployments/github/import.tf
```diff
@@ -1,4 +1,4 @@
 import {
-  to = github_repository.govuk_repos["govuk-content-publishing-guidance"]
-  id = "govuk-content-publishing-guidance"
-}
+  to = github_branch_protection.govuk_repos["select-with-search-component"]
```
AgaDufrat, line 2: We first need to import existing GitHub repository and then add GitHub branch protection for it.

```suggestion
  to = github_repository.govuk_repos["select-with-search-component"]
```

AgaDufrat (approved):

PR #2179, opened 2025-05-19
Added default database environment override and additional workflows with tags and schedules

hamish-pollock-gds:
* Added `default_database` override to be environment specific
* Added addition workflows, tags and schedules for `search-intraday`, `search-daily` and `search-monthly`


rtrinque (approved):
Looks good

Issue #2172, opened 2025-05-19
Review and treat: IAM Access Analyzer - external access


nicholsj:
See https://github.com/alphagov/govuk-platform-internal/issues/45

jasonBirchall:
@samsimpson1, can we mark this ticket resolved?

samsimpson1:
We now have access analysers in each region in the integration/staging/production AWS accounts, and the Trusted Advisor alert will disappear the next time AWS checks it.

nicholsj:
Closes https://github.com/alphagov/govuk-platform-internal/issues/45

PR #2193, opened 2025-05-21
Create an AWS Access Analyzer in each non-ephemeral account

samsimpson1:
https://github.com/alphagov/govuk-infrastructure/issues/2172


nimalank7:
Any particular reason why we're not doing this in the ephemeral environment?

samsimpson1:
Its a box ticking exercise and we don't really want one analyser per cluster

nimalank7 (approved):
LGTM!

PR #2208, opened 2025-05-29
Update Bedrock permissions for EKS access

ianhowell-gds:
## What

Update Bedrock permissions to allow access from EKS services to all models

## Why

This is to enable the devs to test different Bedrock models in order to compare results


rtrinque (approved):
Looks good

PR #2221, opened 2025-06-02
Create AWS IAM Access Analyzers in each AWS region

samsimpson1:
One per region is required to satisfy the Trusted Advisor check

https://github.com/alphagov/govuk-infrastructure/issues/2172

### DIFF
--- a/terraform/deployments/vpc/iam_access_analyzers.tf
+++ b/terraform/deployments/vpc/iam_access_analyzers.tf
@@ -1,0 +1,102 @@
+resource "aws_accessanalyzer_analyzer" "ap-northeast-1" {
+  analyzer_name = "govuk-ap-northeast-1"
+  provider      = aws.ap-northeast-1
+  count         = local.is_ephemeral ? 0 : 1
+}
+
+resource "aws_accessanalyzer_analyzer" "ap-northeast-2" {
+  analyzer_name = "govuk-ap-northeast-2"
+  provider      = aws.ap-northeast-2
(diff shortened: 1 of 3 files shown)


terraform/deployments/vpc/iam_access_analyzers.tf
```diff
@@ -0,0 +1,260 @@
+# Creates IAM Access Analyzers in each AWS region.
+# This file has loads of providers because Terraform requires one provider per region.
+
+provider "aws" {
```
AP-Hunt: Should we move all the provider blocks to `main.tf` where the other provider lives?

samsimpson1: Normally I'd agree they should live in `main.tf`, but here they're only used for this one purpose, so I think it makes more sense to keep them with the only resources that use them

jasonBirchall: ... I kind of agree with @AP-Hunt. It makes sense that they live centrally.

terraform/deployments/vpc/iam_access_analyzers.tf
```diff
@@ -0,0 +1,260 @@
...
+  region = "ap-northeast-1"
+  alias  = "aa-ap-northeast-1"
```
jasonBirchall: The `aa-..` part of this. Is there any logic to `aa`?

samsimpson1: The prefixes have been removed

PR #2222, opened 2025-06-02
Search API v2: Simplify datastore schema management

csutter:
Now that we have [removed the schema compliance tests][1] from Search
API v2, we no longer need the datastore schema as a full JSON Schema
(which VAIS never truly supported anyway).

This moves the datastore schema into HCL Terraform configuration, which
is more idiomatic in a Terraform repository, gets nicely autoformatted,
and allows us to use comments.

It also removes some of the additional fields that were previously
included in the schema for validation purposes but aren't actually
supported by VAIS or in some cases actively causing issues (such as
`required`, `description`, etc).

[1]: https://github.com/alphagov/search-api-v2/pull/455


terraform/deployments/search-api-v2/datastore_schema.tf
leenagupte, line 1: Could you update the commit to say why the file is being moved. I'm pretty sure it's related to the other PR that removes the schema tests, but it would be good to confirm that and link to that PR.

terraform/deployments/search-api-v2/datastore_schema.tf
```diff
@@ -0,0 +1,203 @@
...
+        content_id = {
+          type        = "string"
+          retrievable = true
+        }
+        # Unique identifier for search quality evaluation (identical to content_id)
+        #
+        # Note: This is only used to compare results for evaluation purposes, and is not a URL
+        # despite the name. We decided against using the alternative 'uri' field name to avoid
+        # confusion with our existing 'url' field and 'uri' key property (see
+        # https://cloud.google.com/generative-ai-app-builder/docs/evaluate-search-quality)
```
leenagupte, line 20: Why have the descriptions been added as comments rather than in a description field as they were previously? Is the description one of the fields not supported by VAIS?

terraform/deployments/search-api-v2/datastore_schema.tf
```diff
@@ -0,0 +1,203 @@
...
+        # stays here.
+        additional_searchable_text = {
+          type       = "string"
+          searchable = true
+        }
+        # URI reference either as a relative path for GOV.UK content, or as an absolute URL for
+        # external content (used to link to the content from search results)
+        link = {
+          type        = "string"
+          retrievable = true
```
leenagupte, line 51: `links` also had a `format` field. I assume this is one of ones not support by VAIS?

terraform/deployments/search-api-v2/datastore_schema.tf
```diff
@@ -0,0 +1,203 @@
...
+        }
+        # A list of GOV.UK taxon IDs that this content object is tagged to (for filtering)
+        part_of_taxonomy_tree = {
+          type = "array"
+          items = {
+            type               = "string"
+            keyPropertyMapping = "category"
+          }
+        }
+        # Whether the content is historic (for boosting; boosts don't support boolean)
```
leenagupte, line 93: Is this comment still true? Is there not a new boolean field now?

terraform/deployments/search-api-v2/datastore_schema.tf
```diff
@@ -0,0 +1,203 @@
...
+        is_historic = {
```
leenagupte, line 94: I'm assuming here that `multipleOf`, `minimum` and `maximum` are also not supported by VAIS.

terraform/deployments/search-api-v2/datastore_schema.tf
csutter: Correct – `description` is not supported by VAIS, and even though it's currently silently ignored (unlike some other fields such as `required` which now seem to do _something_ under the hood), I've removed them and turned them into comments in case they start causing trouble in the future.

The idea is that we only have fields **explicitly and actively supported by VAIS** in here now.

The nice thing about this being HCL/Terraform now is that we can actually have comments (Terraform's JSON parser didn't support the unofficial JSON-with-comments format so we couldn't with the previous setup) ✨

terraform/deployments/search-api-v2/datastore_schema.tf
csutter: There is, but:
a) we cannot change existing fields in the schema, and
b) while there is a boolean type now, it [still isn't supported for boosting purposes](https://cloud.google.com/generative-ai-app-builder/docs/filter-search-metadata?authuser=1#filter-expression-syntax) 🤪

terraform/deployments/search-api-v2/datastore_schema.tf
csutter: Ah this is just a structural choice to avoid cluttering up `discovery_engine.tf` with the massive schema. That said, we could just also leave it in there if you think that makes more sense?

leenagupte: I supposed I guess it was being moved because it's no longer being used in the search-api-V2 tests. But I guess, the fields could have just been removed from the `json` file and everything else left as is?

leenagupte: Except for adding comments? That's also good information to add to the commit message for the "why".

csutter: It makes more sense for it to be Terraform code as that allows for comments and autoformatting, and is more idiomatic in a Terraform repository rather than loading it from a separate file - but yes, I'll update the commit! ✏️

Issue #2171, opened 2025-05-19
Review and treat: IAM Access Key Rotation (Trusted Advisor finding)


nicholsj:
See https://github.com/alphagov/govuk-platform-internal/issues/33

nicholsj:
Isn't completion of this dependent on https://github.com/alphagov/govuk-infrastructure/issues/222?

PR #2223, opened 2025-06-02
Create IAM role for content-data-admin

samsimpson1:
This app is currently using long-lived IAM user and access keys. This role will replace the long-lived user.

https://github.com/alphagov/govuk-infrastructure/issues/2171


AgaDufrat (approved):
LGTM

PR #2259, opened 2025-06-06
Try to import the management of some Legacy SGs

dj-maisy:
## What?
Along with the work of trying to get rid of the use of terraform_remote_state and straggling TF State Data, this aims to import/adopt some legacy SGs that are still in use by some of our current infrastructure.

### The Plan
The plan is to get the existing SGs into state before then substituting the old terraform_remote_state calls in the govuk-publishing-infra Deployment. Hopefully this will minimise on errors.


terraform/deployments/tfc-configuration/security.tf
```diff
@@ -0,0 +1,98 @@
...
+
+  project_name = "govuk-infrastructure"
+  vcs_repo = {
+    identifier     = "alphagov/govuk-infrastructure"
+    branch         = "main"
+    oauth_token_id = data.tfe_oauth_client.github.oauth_token_id
+  }
+
+  team_access = {
+    "GOV.UK Non-Production (r/o)" = "write"
```
samsimpson1: Non-prod shouldn't have write access to this workspace

terraform/deployments/tfc-configuration/security.tf
```diff
@@ -0,0 +1,98 @@
...
```
samsimpson1: Or this one

PR #2277, opened 2025-06-10
✨ Add ADR-0014: Replace Terraform Cloud backend with S3 + DynamoDB

jasonBirchall:
Introduces a pending architecture decision to migrate GOV.UK's Terraform state from Terraform Cloud to an S3 + DynamoDB backend. This change targets >99% cost reduction and removes proprietary lock-in. Includes migration plan, access controls, and consequences. I deliberately don't mention how much TFC currently costs, but the 99% is valid.

ADR number: 14
Status: Pending


nicholsj (approved):
Looks good, thanks

PR #2289, opened 2025-06-12
Use Security "Common" TFvars in Sec Deployment

dj-maisy:
### What?
This "consumes" the new Security "Common" TF Vars that are now available in the latest Sensitive TF Module, published as version 0.0.23.


AP-Hunt (approved):
The new output usage matches when I've seen from the relevant PR creating the output

PR #2290, opened 2025-06-12
Create a role in Kubernetes for readonly access

AP-Hunt:
The readonly role has access to all things, but cannot modify anything.

### DIFF
--- a/terraform/deployments/cluster-services/eks_access.tf
+++ b/terraform/deployments/cluster-services/eks_access.tf
@@ -2,6 +2,7 @@
 data "aws_iam_roles" "cluster-admin" { name_regex = "(\\..*-fulladmin$|\\..*-platformengineer$)" }
 data "aws_iam_roles" "developer" { name_regex = "\\..*-developer$" }
 data "aws_iam_roles" "licensing" { name_regex = "\\..*-licensinguser$" }
+data "aws_iam_roles" "readonly" { name_regex = "\\..*-readonly$" }
 
 locals {
   developer_namespaces = ["apps", "datagovuk", "licensify"]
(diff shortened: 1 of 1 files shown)


terraform/deployments/cluster-services/eks_access.tf
```diff
@@ -87,6 +98,22 @@ resource "aws_eks_access_policy_association" "licensing" {
...
+resource "aws_eks_access_policy_association" "readonly" {
+  for_each = data.aws_iam_roles.readonly.arns
+
+  cluster_name  = local.cluster_name
+  policy_arn    = "arn:aws:eks::aws:cluster-access-policy/AmazonEKSViewPolicy"
+  principal_arn = each.value
+
+  access_scope {
+    type = "cluster"
+  }
```
dj-maisy: We probably don't want to give ReadOnly developers cluster-wide read access. Only what we have in `local.developer_namespaces`.

dj-maisy (approved):
Just the EKS policy association's access_scope that I think needs looking at - the rest looks good though!

terraform/deployments/cluster-services/eks_access.tf
AP-Hunt: :+1: Sounds sensible

terraform/deployments/cluster-services/eks_access.tf
```diff
@@ -87,6 +98,23 @@ resource "aws_eks_access_policy_association" "licensing" {
...
+  policy_arn    = "arn:aws:eks::aws:cluster-access-policy/AmazonEKSViewPolicy"
+  principal_arn = each.value
+
+  access_scope {
+    type       = "namespace"
+    namespaces = local.developer_namespaces
+  }
+
+  depends_on = [
+    aws_eks_access_entry.developer
```
dj-maisy:
```suggestion
    aws_eks_access_entry.readonly
```

dj-maisy (changes requested):
Got one more for ya.

terraform/deployments/cluster-services/eks_access.tf
AP-Hunt: That'll teach me to be a copyin' and a pastin'

Issue #2248, opened 2025-06-05
♻️ Migrate `url_rewrite` Lambda to govuk-infrastructure and upgrade runtime to Node 22


jasonBirchall:
**User Need**

**As a** GOV.UK platform engineer  
**I want** the Terraform definition for the `url_rewrite` Lambda moved to the maintained *govuk-infrastructure* repository and upgraded to a supported runtime  
**so that** we eliminate deprecated-runtime alerts and manage the function in actively maintained code

---

### What’s Needed

- [x] Copy the existing `url_rewrite` Lambda block from  
      `govuk-aws/terraform/projects/infra-mirror-bucket/main.tf`  
      into `govuk-infrastructure/terraform/deployments/cloudfront/main.tf`.
- [x] Update `runtime` from **`nodejs10.x` → `nodejs22.x`** (or later supported version).
- [ ] Import the live Lambda into state for **staging** and **production** workspaces to avoid re-creation.
- [ ] Apply the change in staging → production.

---

### Acceptance Criteria

- [ ] Everything in **What’s Needed** is delivered.
- [x] No Lambda functions in any environment use deprecated runtimes (verified in Trusted Advisor).
- [ ] Terraform state in all three workspaces references the function in its new location without replacement.
- [ ] Any changes during delivery are reflected in this story.

---

### User Comms Plan (if applicable)

Post a brief summary in `#govuk-platform-engineering` once production is applied:
* Confirms the Lambda is now managed in *govuk-infrastructure* on Node 22.
* Includes link to updated documentation and the PR.

---

### Assumptions (optional)

- The existing Lambda code runs unchanged under Node 22.
- CloudFront distribution and triggers remain valid after migration.

---

### Notes

- Original block: <https://github.com/alphagov/govuk-aws/blob/01ef070b8d3cb8a0ba18ddcf491b374b29f674f2/terraform/projects/infra-mirror-bucket/main.tf>
- Target file: `https://github.com/alphagov/govuk-infrastructure/terraform/deployments/cloudfront/main.tf`
- Connects to https://github.com/alphagov/govuk-platform-internal/issues/54
---

jasonBirchall:
@MuriloDalRi, can you explain why this is blocked? It'll mean we can stop explaining each sprint 😸

MuriloDalRi:
I'm just looking at govuk-docker issues at the moment.

MuriloDalRi:
[...]

PR #2292, opened 2025-06-12
Import url_rewrite lambda and upgrade its Node runtime

MuriloDalRi:
This lambda still exists but had not been ported over from the old Terraform.

https://github.com/alphagov/govuk-infrastructure/issues/2248

### DIFF
--- a/terraform/deployments/cloudfront/main.tf
+++ b/terraform/deployments/cloudfront/main.tf
@@ -37,6 +37,11 @@ provider "aws" {
   region = "us-east-1"
 }
 
+provider "aws" {
+  alias  = "eu_west_2"
+  region = "eu-west-2"
+}
+
 provider "archive" {}
 
 resource "aws_cloudfront_cache_policy" "no-cookies" {
@@ -194,10 +199,9 @@ resource "aws_wafv2_web_acl" "cdn_poc_govuk" {
 }
 
 resource "aws_cloudfront_distribution" "www_distribution" {
-  count = var.cloudfront_create ? 1 : 0
-
   aliases    = var.cloudfront_www_distribution_aliases
   web_acl_id = aws_wafv2_web_acl.cdn_poc_govuk.arn
+
   origin {
     domain_name = var.origin_www_domain
     origin_id   = var.origin_www_id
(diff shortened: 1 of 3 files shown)


terraform/deployments/cloudfront/main.tf
```diff
@@ -194,7 +199,7 @@ resource "aws_wafv2_web_acl" "cdn_poc_govuk" {
 }
 
 resource "aws_cloudfront_distribution" "www_distribution" {
-  count = var.cloudfront_create ? 1 : 0
+  count = 1
```
samsimpson1: If the count is always 1, the parameter can be removed

terraform/deployments/cloudfront/main.tf
```diff
@@ -343,3 +348,53 @@ resource "aws_cloudfront_distribution" "assets_distribution" {
...
+  policy_arn = "arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"
+}
+
+data "archive_file" "url_rewrite" {
+  type        = "zip"
+  source_file = "${path.module}/index.js"
+  output_path = "${path.module}/CloudfrontUrlRewrite.zip"
+}
+
+resource "aws_lambda_function" "url_rewrite" {
```
samsimpson1, line 396: Where is this function actually used? I don't see a reference to it in the cloudfront terraform

MuriloDalRi: I've added a `lambda_function_association` that was missing. Does that look like it might work now? @samsimpson1

MuriloDalRi: I asked Roch and he reckons the lambda is still needed, but it's out of scope to test CF as part of this issue. This is just to import the lambda and upgrade the node runtime.

terraform/deployments/cloudfront/main.tf
```diff
@@ -343,3 +352,53 @@ resource "aws_cloudfront_distribution" "assets_distribution" {
     minimum_protocol_version = "TLSv1.2_2021"
   }
 }
+
+import {
+  to = aws_lambda_function.url_rewrite
```
AP-Hunt: This is causing the plan to fail in staging, because it looks like the function doesn't already exist there? If that's right, and intentional, we should probably add a `for_each = var.govuk_environment == "production"? [1] : []` type conditional to this.


iirc you can't use count here, but using a single element array and an empt array are usable substitutes. We're going to remove the import statement in a subsequent commit anyway.

AP-Hunt (changes requested):

Issue #2236, opened 2025-06-03
♻️ Fix AWS assume role permissions for Release app and developers


jasonBirchall:
**User Need**

**As a** developer working on Release
**I want** to assume the correct role for accessing Kubernetes in each environment
**so that** I can develop and test against real deployment data

---

### What’s Needed

- [ ] Developers can assume the `release-assumed` role for each target EKS environment
- [ ] The Release app can assume the same role at runtime
- [ ] Permissions verified using an automated or manual call to the Kubernetes API
- [ ] Ensure the change is resilient to IAM updates or role rotation

---

### Acceptance Criteria

- [ ] Everything in **What’s Needed** is delivered.
- [ ] Any changes during delivery are reflected in this story.
- [ ] Impacted users have been informed appropriately (e.g. change logs, Slack, docs).

---

### User Comms Plan (if applicable)

Notify #govuk-platform-engineering in Slack when access is restored.

---

### Assumptions (optional)

- The same IAM role can be used by both humans and apps (or well-scoped alternatives exist).
- Role trust policies are configured correctly in AWS.

---

### Risks & Mitigation (optional)

- Risk: Permissions regress again in future.
  Mitigation: Document permissions and add monitoring/check to detect access loss early.

---

### Notes

Relevant branch: https://github.com/alphagov/release/tree/use-k8s-info
Connects to https://github.com/alphagov/govuk-infrastructure/issues/1980

jasonBirchall:
[backlog refinement] @samsimpson1 will create a comment on how the current IAM structure works.

samsimpson1:
These are the roles that currently exist:

* `release-assumer` - This role is attached to the Release app via a service account. It is used to assume the `release-assumed` roles in each AWS account
* `release-assumed` - This role exists in each AWS account and has access to the k8s API. It has permission to list/get various k8s resources such as pods, applications, etc

The flow for Release works like this:

1. Release makes request to assume a `release-assumed` role in an AWS account to STS
2. STS responds with temporary credentials
3. Release uses those credentials to generate a K8s access token
4. Access token is used to make requests to the K8s cluster API in that account

[Terraform for release-assumer](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/release/assumer.tf)

[Terraform for release-assumed](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/release/assumed.tf)

[...]

PR #2295, opened 2025-06-13
Allow assumer in non-prod envs to assume other non-prod env accounts

kentsanggds:
Only allow non-production assumer role to assume accounts in other non-produstion environments.

This will limit the security risk if the permissions for the `release-assumed` role were to change in the future. 

https://github.com/alphagov/govuk-infrastructure/issues/2236

### DIFF
--- a/terraform/deployments/release/assumer.tf
+++ b/terraform/deployments/release/assumer.tf
@@ -2,13 +2,14 @@ data "aws_caller_identity" "current" {}
 
 locals {
   # production can assume everywhere,
-  # other accounts can assume into themselves
-  account_ids = var.govuk_environment == "production" ? [
-    "172025368201", # production
+  # other accounts can assume into non-prod accounts
+  _account_ids = [
     "696911096973", # staging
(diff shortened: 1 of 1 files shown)


terraform/deployments/release/assumer.tf
```diff
@@ -2,13 +2,14 @@ data "aws_caller_identity" "current" {}
...
-    "172025368201", # production
+  # other accounts can assume into non-prod accounts
+  _account_ids = [
     "696911096973", # staging
     "210287912431", # integration
     "430354129336"  # test
-  ] : [data.aws_caller_identity.current.account_id]
+  ]
+  account_ids = (var.govuk_environment == "production" ?
+  "${concat(local._account_ids, ["172025368201"])}" : local._account_ids)
```
samsimpson1:
```suggestion
  concat(local._account_ids, ["172025368201"]) : local._account_ids)
```

samsimpson1 (approved):

PR #2304, opened 2025-06-16
Add secret for discovery engine default location

leenagupte:
[Trello](https://trello.com/c/PUn43yi3)

This value will be used to populate a
DISCOVERY_ENGINE_DEFAULT_LOCATION_NAME env var that will be used to in the automated judgment list tasks.

See:
https://github.com/alphagov/search-api-v2/blob/fc7a362b80145c9a16c4d91c6f343eef382c6d7f/config/application.rb#L23C54-L23C113

### DIFF
--- a/terraform/deployments/search-api-v2/discovery_engine.tf
+++ b/terraform/deployments/search-api-v2/discovery_engine.tf
@@ -1,7 +1,8 @@
 # TODO: These IDs/paths are semi-hardcoded here as there aren't first party resources/data sources
 # available for them yet.
 locals {
-  discovery_engine_default_collection_name = "projects/${var.gcp_project_id}/locations/${var.discovery_engine_location}/collections/default_collection"
+  discovery_engine_default_location_name   = "projects/${var.gcp_project_id}/locations/${var.discovery_engine_location}"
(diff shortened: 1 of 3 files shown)


terraform/deployments/search-api-v2/discovery_engine.tf
```diff
@@ -2,6 +2,7 @@
 # available for them yet.
 locals {
   discovery_engine_default_collection_name = "projects/${var.gcp_project_id}/locations/${var.discovery_engine_location}/collections/default_collection"
+  discovery_engine_default_location_name   = "projects/${var.gcp_project_id}/locations/global"
```
csutter: Might be worth flipping these around and using the location as the first part of the collection?

leenagupte: Oh yeah! Good shout.

leenagupte: So:

```
locals {
  discovery_engine_default_location_name   = "projects/${var.gcp_project_id}/locations/${var.discovery_engine_location}"
  discovery_engine_default_collection_name = "${discovery_engine_default_location_name}/collections/default_collection"
}
```

csutter (approved):
🚢

PR #2312, opened 2025-06-18
♻️ Change the k8s serviceaccount name

jasonBirchall:
Mirror has had a little environment refactor. This change reflects a change in serviceaccount that needs to be kept in sync with the role that's created for it.


AgaDufrat (approved):
Checked that it's matching the value in helm charts.

PR #2313, opened 2025-06-18
ADR-0015 maintain a readonly role

AP-Hunt:
Introduces an ADR which documents why we've chosen to maintain a readonly role. It also introduces a new guardrail test to ensure the ADR numbering on the file matches the title.


dj-maisy (approved):
Looks good to me - and the additional workflow is a nice touch too.

samsimpson1 (approved):

Issue #2226, opened 2025-06-03
Configure Fastly to use an IAM role for shipping logs to S3


samsimpson1:
**As a** security-minded infrastructure maintainer
**I want** Fastly to use short-lived credentials for accessing S3 buckets in our AWS accounts instead of long-lived access keys
**so that** I no longer have to manage long-lived credentials and can benefit from the security provided by temporary credentials

---

### What’s Needed

- [x] Replace the `govuk-production-fastly-logs-writer` and `govuk-analytics-logs-fastly` IAM users with IAM roles with the equivalent permissions
- [x] Configure the new roles to allow Fastly to assume them
- [x] Configure Fastly to use the new roles instead of the existing IAM users
- [x] Disable/remove the old IAM user accounts and access keys

---

### Acceptance Criteria

- [x] IAM users are no longer used for Fastly to authenticate with AWS on our behalf
- [x] Short-lived credentials are used by Fastly instead
- [x] Any new infrastructure is managed by Terraform
- [ ] Documentation is updated to reflect the new way of authentication for Fastly's S3 access

---

### Notes

Relevant Fastly documentation:

https://www.fastly.com/documentation/guides/integrations/streaming-logs/creating-an-aws-iam-role-for-fastly-logging/

https://www.fastly.com/documentation/guides/integrations/logging-endpoints/log-streaming-amazon-s3/

Related issue: https://github.com/alphagov/govuk-platform-internal/issues/33

---

dj-maisy:
### Update (2025-06-12):

* Created Sensitive Module variable for Fastly Account ID
* Now trying to figure out whether the new role should live in `govuk-infrastructure` or `govuk-fastly`. Could use a second opinion on this.

dj-maisy:
### Update (2025-06-17):
* https://github.com/alphagov/govuk-fastly/pull/168
* https://github.com/alphagov/govuk-infrastructure/pull/2310

dj-maisy:
### Update (2025-05-19):

* https://github.com/alphagov/govuk-fastly/pull/170
* https://github.com/alphagov/govuk-fastly-secrets/pull/60
* https://github.com/alphagov/govuk-infrastructure/pull/2320
* https://github.com/alphagov/govuk-fastly-secrets/pull/61

New roles applied and available in all environments, old credentials disabled. Final task is to delete/destroy old IAM users.

dj-maisy:
All old IAM users are now deleted. Fastly has now adopted the new IAM Roles.

PR #2329, opened 2025-06-20
Document IAM Roles work as an ADR

dj-maisy:
## What?
This sets up an ADR around our "Decision" to use IAM Roles (and by extension Pod Identities) to deprecate the use of AWS Access Keys wherever possible.

Occasionally, there may be a situation where an external service or app does not support using Roles, however, this policy will minimise the number of IAM users and Access Keys that we would have to manage and rotate.

### Related

* Closes https://github.com/alphagov/govuk-infrastructure/issues/2226


jasonBirchall (approved):
Awesome ❤️

PR #2341, opened 2025-06-24
Change publishing infrastructure subdomain DNS handling to match govuk-dns-tf

AP-Hunt:
We'll be importing the DNS records from govuk-dns-tf and it has some logic associated with the subdomain names. We make the logic here match [the logic there](https://github.com/alphagov/govuk-dns-tf/blob/main/modules/zone/zone.tf#L80) so that we can import the records with no issues.


jfharden:
2 of the plans aren't clean (unrelated changes which look like they could be important, but are hard to grok because TF Clouds plan output is really hard to read)

AP-Hunt:
Looks like the change in the plan is Terraform doing a bad diff of ordering within some JSON

Issue #2326, opened 2025-06-19
✨ Create Runbooks for Upgrading Postgres


jasonBirchall:
User Need

As an Application Developer
I need a clear, step-by-step runbook for upgrading my service’s PostgreSQL 13 cluster to PostgreSQL 14
so that I can perform the change confidently, with minimal downtime and without SRE hand-holding.

---

Functional Requirements (What):

1. Author a canonical **Upgrade Runbook** covering:
   • Pre-requisites (extension upgrades, backups, feature freezes, primary-key checks).  
   • RDS Blue/Green workflow (creation, verification, switchover, rollback).  
   • Post-upgrade validation (health checks, smoke tests, metrics to watch).  

2. Provide **Environment-specific checklists** (integration → staging → production) with dates/placeholders.

3. Package example Terraform snippets for:
   • Enabling logical replication/back-ups.  
   • Importing the green DB into state post-promotion.  

4. Publish a **Quick-Reference Cheat Sheet** (one-page) linking to the full runbook, common psql commands, and slack escalation points.

5. **Socialise** the runbook:
   • Walk-through demo in platform demos and invite key stakeholders like @richardTowers.
   • Message in `#govuk-developers` and `#govuk-platform-engineering`.  
   • Add to GOV.UK Developer Docs.

---

Acceptance Criteria:

- [x] A runbook lives in Google Docs or somewhere central
- [ ] Environment-checklist template documented and shared.
- [ ]  At least two product teams have walked through the runbook in integration and signed off that it’s clear.

User Communication Plan:

* Slack announcement in `#govuk-developers`, `#govuk-platform-engineering`, and `#govuk-platform-support` with link to the runbook.
* Five-minute show-and-tell in next sprint demo.

Risks and Mitigation:

* **Risk:** Steps become stale as AWS updates RDS UI/CLI flags  
  **Mitigation:** Add “Validate against latest AWS docs” checkpoint before each prod upgrade; assign a doc maintainer.

* **Risk:** Developers skip pre-requisite primary-key checks, causing logical-replication failures  
  **Mitigation:** Runbook includes mandatory SQL snippet and automated CI job to block merge if violations exist.

Notes:

* PostgreSQL 13 community EOL: **13 Nov 2025**; AWS deprecation expected **28 Feb 2026**.
* Document builds on the detailed options analysis and test results captured in “PostgreSQL 13 → 14 upgrade proposal”.
[...]

PR #2344, opened 2025-06-25
Turn on logical replication for ckan integration

nimalank7:
Description:
- Set up green parameter group for Postgres14 upgrade
- Turn on logical replication to enable blue-green deployments
- https://github.com/alphagov/govuk-infrastructure/issues/2326

### DIFF
--- a/terraform/deployments/tfc-configuration/variables-integration.tf
+++ b/terraform/deployments/tfc-configuration/variables-integration.tf
@@ -205,7 +205,10 @@ module "variable-set-rds-integration" {
           log_statement              = { value = "all" }
           deadlock_timeout           = { value = 2500 }
           log_lock_waits             = { value = 1 }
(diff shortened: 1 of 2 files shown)


terraform/deployments/tfc-configuration/variables-integration.tf
```diff
@@ -205,7 +205,10 @@ module "variable-set-rds-integration" {
           log_statement              = { value = "all" }
           deadlock_timeout           = { value = 2500 }
           log_lock_waits             = { value = 1 }
+          rds.logical_replication    = { value = 1, apply_method = "pending-reboot" }
```
AP-Hunt: I'm tentative about these getting applied _before_ Terraform causes the database to get rebooted, but I'm going to trust Terraform to do the right thing.

AP-Hunt (approved):

PR #2356, opened 2025-06-25
Import ckan postgres14 integration to Terraform

nimalank7:
Description:
- Import ckan integration postgresql14 into Terraform
- https://github.com/alphagov/govuk-infrastructure/issues/2326


AP-Hunt (changes requested):
We should delete the commented out bits, not leave the commented out

PR #2385, opened 2025-07-04
Provision new Postgres DBs for Publisher

Tetrino:
We are migrating Publisher to run on Postgres 14 from Mongo. The integration database was provisioned in January but now we are ready to look towards delivery we need these available in staging and production.

The actual settings are based on service-manuals-publisher.

### DIFF
--- a/terraform/deployments/tfc-configuration/variables-production.tf
+++ b/terraform/deployments/tfc-configuration/variables-production.tf
@@ -464,6 +464,23 @@ module "variable-set-rds-production" {
         has_read_replica             = true
       }
 
+      publisher = {
+        engine         = "postgres"
+        engine_version = "17"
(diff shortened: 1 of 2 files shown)


terraform/deployments/tfc-configuration/variables-production.tf
```diff
@@ -464,6 +464,23 @@ module "variable-set-rds-production" {
         has_read_replica             = true
       }
 
+      publisher = {
+        engine         = "postgres"
+        engine_version = "14"
```
richardTowers: 🤔  I wonder if we should go straight to 17, if it's a new database... Otherwise we're just storing ourselves up an upgrade task.

richardTowers (approved):

PR #2387, opened 2025-07-04
♻️ Change the IAM role name for reports prototype

jasonBirchall:
Before this commit it is set to `govuk-reports-govuk` (the second govuk is the cluster name) , which doesn't make a lot of sense. This change adds the word prototype and removes the second govuk instance

### DIFF
--- a/terraform/deployments/govuk-reports/iam.tf
+++ b/terraform/deployments/govuk-reports/iam.tf
@@ -1,7 +1,7 @@
 # IAM role for govuk-reports application with access to Cost Explorer, RDS, and tagging APIs
 
 locals {
-  govuk_reports_service_account_name = "govuk-reports"
+  govuk_reports_service_account_name = "govuk-reports-prototype"
 }
 
 # IAM policy document for govuk-reports permissions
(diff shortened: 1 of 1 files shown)


terraform/deployments/govuk-reports/iam.tf
```diff
@@ -1,7 +1,7 @@
 # IAM role for govuk-reports application with access to Cost Explorer, RDS, and tagging APIs
 
 locals {
-  govuk_reports_service_account_name = "govuk-reports"
+  govuk_reports_service_account_name = "govuk-reports-prototype"
```
AP-Hunt, line 4: This should probably still include the cluster id/name, otherwise we'll get conflicts if we try to deploy it in an ephemeral cluster (which share an AWS account)

AP-Hunt (changes requested):

terraform/deployments/govuk-reports/iam.tf
jasonBirchall: so keep `govuk-reports-govuk`?

jasonBirchall: it's a good point on ephemeral clusters.

jasonBirchall: maybe remove the first govuk. Just so it doesn't make my skin crawl :p

AP-Hunt: It'll say `govuk-reports-prototype-govuk` in most envs, yeah. But in Terraform it'll read in a less skin itching way at least

PR #2423, opened 2025-07-15
Set up Cloudwatch logging for Bedrock

jackbot:
https://trello.com/c/V3Dnjlfl/2579

We want to track Bedrock usage (specifically token usage) for GOV.UK
Chat, so know how close to our quota we are.

To do this, we need to set up Bedrock to enable logging to CW. From
there we'll pull the data out and display it in a Grafana dashboard.

This sets up the following:

* Adds a log group to Cloudwatch for Bedrock logs
* Attaches an IAM policy to the existing Bedrock role which allows
  creating log streams and sending log events
* Turns on logging in Bedrock

See [AWS documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/model-invocation-logging.html)

### DIFF
--- a/terraform/deployments/chat/bedrock_iam.tf
+++ b/terraform/deployments/chat/bedrock_iam.tf
@@ -36,6 +36,19 @@ data "aws_iam_policy_document" "bedrock_access" {
   }
 }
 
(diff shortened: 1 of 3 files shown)


ianhowell-gds (approved):
LGTM

Issue #2335, opened 2025-06-20
✨ Pilot the Upgrade Runbook on data.gov.uk (DGU)


jasonBirchall:
User Need

As a Platform Team
We need to trial the new PostgreSQL 13 → 14 upgrade runbook on a real service (data.gov.uk)
so that we can validate the instructions, timings and rollback steps before wider rollout.

⚠️ Constraint: @nimalank7 must not participate in hands-on work, as he authored the runbook. _Maybe advice only_ 🙂

Functional Requirements (What):

1. Select the DGU database clusters in integration, staging and production.
2. Apply the runbook exactly as written, recording:
   • Actual durations for each phase (blue/green creation, storage initialisation, switchover).  
   • Any deviations / ambiguities discovered.  
   • Metrics and logs captured before, during and after.  
3. Exercise the rollback drill in integration **even if the upgrade succeeds**, to test documentation.
4. Raise PRs to refine the runbook or Terraform examples for any gaps found.
5. Produce a short “lessons-learnt” write-up and attach it to the runbook ticket.

Acceptance Criteria:

* DGU’s integration and staging clusters are running PostgreSQL 14.18 with no data loss.
* Production upgrade is executed in a planned window and downtime ≤ 2 minutes.
* Rollback drill in integration proves restorable to PostgreSQL 13 within 15 minutes.
* Runbook updated (PR merged) to reflect all clarifications; peer-review shows no outstanding comments.
* “Lessons-learnt” document published and linked from the runbook.

User Communication Plan:

* Pre-upgrade Slack heads-up in `#govuk-platform-support` & `#govuk-datagovuk`.
* Add an item to the next sprint demo summarising outcome and runbook changes.

Risks and Mitigation:

* **Risk:** DGU extensions (CKAN/PostGIS) lag behind minimum versions.  
  **Mitigation:** Verify extension versions a week in advance; schedule upgrade work if needed.

* **Risk:** Runbook ambiguity causes pause mid-flight.  
  **Mitigation:** Ensure pair has direct Slack bridge to @nimalank7  for clarification; document any blockers immediately.

Notes:

* Previous ticket: “Developer-Facing Upgrade Runbook” (https://github.com/alphagov/govuk-infrastructure/issues/2326).

jasonBirchall:
~[backlog refinement] Blocked by #2326~
Now unblocked.

nimalank7:
[...]

PR #2466, opened 2025-07-21
Prepare ckan for postgres upgrade in production

MuriloDalRi:
https://github.com/alphagov/govuk-infrastructure/issues/2335

### DIFF
--- a/terraform/deployments/rds/ckan_production_postgres_14_upgrade.tf
+++ b/terraform/deployments/rds/ckan_production_postgres_14_upgrade.tf
@@ -1,0 +1,25 @@
+resource "aws_db_parameter_group" "ckan_postgresql14_green_params" {
+
+  name_prefix = "${var.govuk_environment}-ckan-postgres-"
+  family      = "postgres14"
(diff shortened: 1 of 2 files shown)


terraform/deployments/rds/ckan_production_postgres_14_upgrade.tf
```diff
@@ -0,0 +1,25 @@
+resource "aws_db_parameter_group" "postgresql14_green_params" {
```
samsimpson1: Is this parameter group going to stick around? If so, it should probably have something in the name to distinguish it (since its only for CKAN but the name is generic)

nimalank7: Only temporarily but yeah the resource name should be `ckan_postgresql14_green_params`

MuriloDalRi: I've updated it now to `ckan_postgresql14_green_params`

PR #2494, opened 2025-07-29
Rational Repo Configuration

dj-maisy:
## What?
This changes our Github Terraform so that we are using our explicit YAML list of known Repositories to configure the necessary settings for our Repos to work correctly.

Also:

* There is an output so this can be used by the ECR deployment.
* There is a schema check too so we can validate that the YAML file makes sense and isn't missing anything required.

### DIFF
--- a/.github/workflows/github-repos-validation.yml
+++ b/.github/workflows/github-repos-validation.yml
@@ -1,0 +1,18 @@
+name: Validate Github Repos YAML
+on:
+  push:
+    paths: ['terraform/deployments/github/**']
+jobs:
+  github_repos_validation:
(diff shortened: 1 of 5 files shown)


dj-maisy:
Repo ID `521697368` relates to `govuk-replatform-test-app`, which IMO doesn't need those permissions any longer as we no longer deploy this. So that should be fine.

Have asked the relevant teams Re their repos:

* `govuk_web_banners` is fine to be imported and managed.
* `govuk_chat_private` shall be removed from this PR for now and we'll leave it up to the Chat team to deal with.

PR #2499, opened 2025-07-30
Update cluster-secrets helm chart version

jfharden:
Update to the new version of cluster-secrets which includes the logit opensearch secrets needed for the lubernetes events shipper


jfharden:
Ah I didn't know this would happen automatically, so this is superseded by https://github.com/alphagov/govuk-infrastructure/pull/2500/

PR #2512, opened 2025-08-01
Fix url_rewrite lambda in staging

MuriloDalRi:


AP-Hunt (approved):
I'm trusting that our deployer role has sufficient permission to do whatever API actions are needed for versioned Lambda functions

PR #2524, opened 2025-08-05
Add govuk_chat_private as a repo

kevindew:
This adds the repo checks. The regular standard security checks CodeQL doesn't work for a private repo and the dependency-review one seems to require additional set-up beyond what a GitHub admin can do. So both are disabled.

### DIFF
--- a/terraform/deployments/github/repos.yml
+++ b/terraform/deployments/github/repos.yml
@@ -429,6 +429,19 @@ repos:
       additional_contexts:
         - test
 
+  govuk_chat_private:
+    homepage_url: "https://docs.publishing.service.gov.uk/repos/govuk_chat_private.html"
+    visibility: private
+    # This utility is a gem that isn't pushed to Rubygems, but we need to set
+    # this to true for the appropriate CI token to be set
+    publishes_gem: true
+    required_status_checks:
+      # standard security checks are disabled as not configured for a private repo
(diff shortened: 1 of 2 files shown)


terraform/deployments/github/repos.yml
```diff
@@ -383,6 +383,15 @@ repos:
       additional_contexts:
         - test
 
+  govuk_chat_private:
+    homepage_url: "https://docs.publishing.service.gov.uk/repos/govuk_chat_private.html"
```
dj-maisy, line 433:
```suggestion
    homepage_url: "https://docs.publishing.service.gov.uk/repos/govuk_chat_private.html"
    visibility: private
```

dj-maisy: I would add ~these additional lines~ the `visibility: private` proeprty ~(as we are replacing the use of GOV.UK Topics)~

dj-maisy (changes requested):
You may also wish to edit `main.tf` and add:

```
import {
  to = github_repository.govuk_repos["govuk_chat_private"]
  id = "govuk_chat_private"
}

import {
  to = github_branch_protection.govuk_repos["govuk_chat_private"]
  id = "govuk_chat_private"
}
```

Otherwise the Terraform will likely complain that you're trying to add a Repo that already exists.

PR #2549, opened 2025-08-11
✨ Add default tagging options to chat terraform

jasonBirchall:
The default options are described in further details by the aws way definition: https://gds-way.digital.cabinet-office.gov.uk/manuals/aws-tagging.html#tagging-aws-resources

### DIFF
--- a/terraform/deployments/chat/main.tf
+++ b/terraform/deployments/chat/main.tf
@@ -16,6 +16,16 @@ terraform {
 
 provider "aws" {
   region = var.aws_region
+  default_tags {
+    tags = {
+      product              = "govuk"
+      system               = "govuk-chat"
+      environment          = var.govuk_environment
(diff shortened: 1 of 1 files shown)


ianhowell-gds:
Would it make sense to add the same default tags to Chat Opensearch resources at the same time?

- `https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/opensearch/main.tf#L18`

jasonBirchall:
@ianhowell-gds, absolutely. This is where I'm heading. One file at a time :) 

So, the change you suggest will happen, just in a later PR.

ianhowell-gds (approved):
Awesome, thank you Jason 🙂

jasonBirchall:
@jfharden are you happy with this? There will be loads more.

PR #2599, opened 2025-08-15
Improves link text throughout our documentation

AP-Hunt:
The GOV.UK style guide explicitly says to avoid links like "here" or "page". This is a follow up to #2582


terraform/deployments/opensearch/README.md
```diff
@@ -27,5 +28,5 @@ pip install boto3 requests requests-aws4auth
 python register-snapshot-repository.py [test|integration|staging|production]
 ```
 
-[here]: https://docs.aws.amazon.com/opensearch-service/latest/developerguide/managedomains-snapshots.html
-[page]: https://docs.publishing.service.gov.uk/manual/manage-opensearch-on-aws.html
+[here]: 
+[page]:
```
jasonBirchall: Should we remove these as well?

jasonBirchall (approved):

PR #2616, opened 2025-08-19
Add check for if workspaces exist in ephemeral cluster shutdown script

samsimpson1:
This will stop the script from getting into an infinite loop if a TFC workspace doesn't exist. This sometimes happens if a given workspace was never created. For example, sometimes the rds module is not created to save time.

### DIFF
--- a/terraform/deployments/ephemeral/shutdown.sh
+++ b/terraform/deployments/ephemeral/shutdown.sh
@@ -98,6 +98,13 @@ function tfc_get_workspace_id {
     | jq -r ".data[0].id"
 }
 
+# check if a workspace exists
+function tfc_workspace_exists {
+  WORKSPACE_NAME="${1}"
+  WORKSPACE_ID="$(tfc_get_workspace_id "${WORKSPACE_NAME}")"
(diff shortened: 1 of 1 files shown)


terraform/deployments/ephemeral/shutdown.sh
```diff
@@ -190,6 +204,14 @@ function retry {
   exit 1
 }
 
+echo "AWS IAM Role ARN: $(aws sts get-caller-identity --query Arn --output text)"
+echo "EKS Cluster Name: ${CLUSTER_ID}"
+echo ""
+echo "This script is destructive."
+echo "It will delete all Application resources in the cluster and trigger Terraform destroy runs."
+echo "Continue with cluster shutdown? [Enter]"
+read
```
jaskaransarkaria, line 213: Will this continue if any key is pressed but the suggested output from the echo above is looking for the Enter key?

samsimpson1: `read` will block until the user hits enter, so the choice is to either exit the script or hit enter

PR #2618, opened 2025-08-19
Always install the current LTS version of NodeJS in GitHub Actions

AP-Hunt:
Our GitHub Actions workflows install NodeJS in two reusable workflows:
- `setup-node` (n.b. this is NOT the standard GHA action, but our wrapper around it)
- `publish-npm-package` (this DOES use the standard GHA actions/setup-node action)

The latter already installed the latest NodeJS LTS version using the supported `lts/*` syntax. This commit makes our `actions/setup-node` reusable workflow do the same thing.

This should mean we don't need to change the version in future. We should continue to get the LTS versions as they're released.


MuriloDalRi (approved):

MuriloDalRi:
Nice!

Issue #2543, opened 2025-08-08
✨ Create query & docs for OOM Killed pods


jasonBirchall:
**User Need**

**As a** GOV.UK developer  
**I want** a ready-made Grafana dashboard (plus clear instructions) that surfaces OOM killed trends  
**so that** I can diagnose pod issues and cluster health quickly without crafting ad-hoc queries.

---

### What’s Needed

- [ ] Build a **Grafana dashboard** using the `container_oom_events_total` metric, containing saved visualisations:
  - **OOMKilled & Evicted pods by application**
- [ ] **Developer documentation**:
  - Steps to locate and load the dashboard.
  - Brief explanation of each widget and typical filters (cluster, namespace, time range).
  - How to clone/extend the dashboard.
- [ ] Announce availability in **#govuk-platform-engineering** and **#govuk-developers** Slack channels (include screenshot + doc link).

---

### Acceptance Criteria

- [ ] Dashboard is visible to the *Developer*  in staging & production.
- [ ] Documentation is merged, searchable, and linked from the dashboard description.
- [ ] Everything in **What’s Needed** is delivered.
- [ ] Any changes during delivery are reflected in this story.
- [ ] Impacted users have been informed via Slack and the developer change log.

---

### User Comms Plan

---

### Assumptions

---

### Risks & Mitigation

---

### Notes

---

samsimpson1:
Context for this issue: this is just to get an idea of what pods are OOMing in our clusters, since we don't really have any way of knowing that right now.

samsimpson1:
I have created a Grafana dashboard that uses a Countach Logs Insights query to fetch a list of pods that have been OOM killed, however the query this dashboard uses is very expensive if it is ran over multiple days ($0.15 for 7 days of logs).

I have instead decided to just document the CloudWatch query that can be used directly via the AWS console, along with a statement about how much the query can cost to execute. This should keep costs low.

samsimpson1:
A CloudWatch Logs Insights saved query has been created in all environments. It is managed by terraform, and will list OOM events for pods in a given EKS cluster.

This saved query has been documented on the ['Get information about your app'](https://docs.publishing.service.gov.uk/kubernetes/manage-app/get-app-info/#view-out-of-memory-events) page, along with a note about how the query is expensive to run.

PR #2621, opened 2025-08-20
Add Killed Pods (Out of Memory) CloudWatch Logs Insights query

samsimpson1:
https://github.com/alphagov/govuk-infrastructure/issues/2543

### DIFF
--- a/terraform/deployments/cluster-infrastructure/cloudwatch_logs_queries.tf
+++ b/terraform/deployments/cluster-infrastructure/cloudwatch_logs_queries.tf
@@ -1,0 +1,13 @@
+locals {
+  # only include cluster name if its an ephemeral cluster
+  cloudwatch_query_suffix = var.cluster_name == "govuk" ? "" : " (${var.cluster_name})"
+}
+
+# CloudWatch Logs Insights query to find OOM-killed pods
+resource "aws_cloudwatch_query_definition" "oom_killed_pods" {
+  name = "Pods Killed (Out of Memory)${local.cloudwatch_query_suffix}"
+
+  log_group_names = [module.eks.cloudwatch_log_group_name]
+
+  query_string = file("cloudwatch_queries/oom_killed_pods.txt")
+}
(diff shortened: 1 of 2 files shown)


terraform/deployments/cluster-infrastructure/cloudwatch_queries/oom_killed_pods.txt
```diff
@@ -0,0 +1,6 @@
+fields @timestamp, responseObject.metadata.namespace, responseObject.metadata.name
```
jfharden, line 1: The oomkilled events only occur in 1 of the 5 log types, you can significantly reduce the cost of this query by restricting the search to specific log stream prefixes.

I can't remember which one it is, but if you find one event and look at it's log stream name, you should be able to filter for that prefix (something like kube-security-audit) or something

terraform/deployments/cluster-infrastructure/cloudwatch_logs_queries.tf
```diff
@@ -0,0 +1,13 @@
+locals {
+  # only include cluster name if its an ephemeral cluster
+  cloudwatch_query_suffix = var.cluster_name == "govuk" ? "" : " (${var.cluster_name})"
```
jfharden, line 3: ❤️

terraform/deployments/cluster-infrastructure/cloudwatch_queries/oom_killed_pods.txt
samsimpson1: Is that how the querying works? I was under the impression that it would search the entire log group no matter what filters you add

jfharden: I believe logstream is one of the things it's partitioned by, so I think it means it doesn't need to scan the streams which don't match the pattern. I'm not 100% sure, but I think this is right

samsimpson1: I just tried filtering based on `@logStream` and it still scanned the same amount of logs, so I'm going to merge this as-is.

Issue #2522, opened 2025-08-05
Fix url_rewrite lambda in staging


MuriloDalRi:
The terrafrom fails when applying in [cloudfront-staging](https://app.terraform.io/app/govuk/workspaces/cloudfront-staging/runs/run-FCos32Wzb5jMrAQB) because of url_rewrite lambda.

jasonBirchall:
@MuriloDalRi are you happy I've captured the error?

nimalank7:
https://github.com/alphagov/govuk-infrastructure/pull/2611

I recreated the Lambda in the right AWS region but I had to temporarily remove the Lambda association as it was causing an issue with calling `https://d2uru0rbwu9sz0.cloudfront.net/__canary__` which is our pingdom alerts

kentsanggds:
Paired with @jfharden on this ticket and we made a number of discoveries.

It appears that the `url_rewrite` lambda will not serve pages correctly when implemented as it will always append a `.html` to the request uri passed in. 

The rewrite code was extracted and tested on a javascript compiler -

```
    var requestUrl = "/__canary__";

    var regex_ext = "^([^#\\?]+)\\.(atom|chm|css|csv|diff|doc|docx|dot|dxf|eps|gif|gml|html|ico|ics|jpeg|jpg|JPG|js|json|kml|odp|ods|odt|pdf|PDF|png|ppt|pptx|ps|rdf|rtf|sch|txt|wsdl|xls|xlsm|xlsx|xlt|xml|xsd|xslt|zip)([\\?#]+.*)?$";
    var regex_html = "^([^#\\?]+)([\\?#]+.*)?$";
    var regex_root = "^\/(.+)"

    // Replace multiple slashes
    var redirectUrl = requestUrl.replace(/([^:])\/\/+/g,'$1/');
    console.log(redirectUrl);
    // Requests without document extension, rewrite adding .html
    if ((redirectUrl.match(regex_ext) == null) &&  (redirectUrl.match(regex_root) != null)){
            var match_html = redirectUrl.match(regex_html);
            redirectUrl =  [match_html[1], '.html', match_html[2]].join('');
    }

console.log(redirectUrl);

```

The output from this was `__canary__.html`, originally it was `.html` but this was because the `regex_html` var was missing a group capture block which is fixed in this code extract. 

[...]

PR #2622, opened 2025-08-20
Send `www-govuk` CloudFront logs to CloudWatch instead of S3 using CloudFront V2 logging

nimalank7:
Description:
- Our current setup uses the legacy CloudFront logging but it is currently not working as we haven't configured the [S3 ACL](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/standard-logging-legacy-s3.html#AccessLogsBucketAndFileOwnership) on `govuk-production-aws-logging`
- Switch to [V2 standard logging](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/standard-logging.html) and deliver logs to CloudWatch
- https://github.com/alphagov/govuk-infrastructure/issues/2522

### DIFF
--- a/terraform/deployments/cloudfront/logging.tf
+++ b/terraform/deployments/cloudfront/logging.tf
@@ -1,0 +1,67 @@
+resource "aws_iam_role" "cloudfront_cloudwatch" {
+  name               = "${var.govuk_environment}-cloudfront-cloudwatch-role"
+  assume_role_policy = data.aws_iam_policy_document.cloudfront_cloudwatch_assume_role.json
+}
+
+data "aws_region" "current" {
+  region = var.aws_region
+}
+
+data "aws_caller_identity" "current" {}
+
+data "aws_iam_policy_document" "cloudfront_cloudwatch_assume_role" {
+  statement {
+    principals {
+      type        = "Service"
+      identifiers = ["cloudfront.amazonaws.com"]
+    }
+    actions = ["sts:AssumeRole"]
+  }
+}
+
+data "aws_iam_policy_document" "cloudfront_cloudwatch" {
+  statement {
+    actions = [
+      "logs:CreateLogStream",
+      "logs:PutLogEvents"
+    ]
+    effect = "Allow"
+    resources = [
+      "arn:aws:logs:${data.aws_region.current.region}:${data.aws_caller_identity.current.account_id}:log-group:${aws_cloudwatch_log_group.www_distribution_cloudfront_log_group.name}:*"
+    ]
+  }
+}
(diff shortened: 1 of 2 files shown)


terraform/deployments/cloudfront/logging.tf
```diff
@@ -0,0 +1,67 @@
+resource "aws_iam_role" "cloudfront_cloudwatch" {
+  name               = "cloudfront-cloudwatch-role"
```
AP-Hunt: Can this have the environment name in it so we're able to deploy multiple of them in one account, please?

terraform/deployments/cloudfront/logging.tf
```diff
@@ -0,0 +1,67 @@
...
+    ]
+    effect = "Allow"
+    resources = [
+      "arn:aws:logs:${data.aws_region.current.region}:${data.aws_caller_identity.current.account_id}:log-group:${aws_cloudwatch_log_group.www_distribution_cloudfront_log_group.name}:*"
+    ]
+  }
+}
+
+resource "aws_iam_policy" "cloudfront_cloudwatch" {
+  name   = "cloudfront-cloudwatch-policy"
```
AP-Hunt: Similarly here

terraform/deployments/cloudfront/logging.tf
```diff
@@ -0,0 +1,67 @@
...
+  policy = data.aws_iam_policy_document.cloudfront_cloudwatch.json
+}
+
+resource "aws_iam_role_policy_attachment" "cloudfront_cloudwatch" {
+  role       = aws_iam_role.cloudfront_cloudwatch.name
+  policy_arn = aws_iam_policy.cloudfront_cloudwatch.arn
+}
+
+resource "aws_cloudwatch_log_group" "www_distribution_cloudfront_log_group" {
+  name              = "/aws/cloudfront/www"
```
AP-Hunt: These probably ought to include the env name too

terraform/deployments/cloudfront/logging.tf
nimalank7: Done.

terraform/deployments/cloudfront/logging.tf
nimalank7: Done.

terraform/deployments/cloudfront/logging.tf
nimalank7: Appended an environment name to the end.

AP-Hunt (approved):

PR #2623, opened 2025-08-20
Add govuk-chat-gradio-prototype to repos.yml

davidgisbey:
We've recently added this prototype to conduct rapid experiments for GOV.UK Chat using Gradio.

### DIFF
--- a/terraform/deployments/github/repos.yml
+++ b/terraform/deployments/github/repos.yml
@@ -298,6 +298,17 @@ repos:
         - Format
         - Type checks
 
+  govuk-chat-gradio-prototype:
+    visibility: private
+    homepage_url: "https://docs.publishing.service.gov.uk/repos/govuk-chat-gradio-prototype.html"
+    required_status_checks:
+      # standard security checks are disabled as not configured for a private repo
+      additional_contexts:
+        - Tests
+        - Lint
+        - Format
+        - Type checks
+
   govuk-content-api-docs:
     homepage_url: "https://content-api.publishing.service.gov.uk"
     need_production_access_to_merge: false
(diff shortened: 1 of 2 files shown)


terraform/deployments/github/repos.yml
```diff
@@ -298,6 +298,16 @@ repos:
         - Format
         - Type checks
 
+  govuk-chat-gradio-prototype:
+    homepage_url: "https://docs.publishing.service.gov.uk/apps/govuk-chat-gradio-prototype.html"
```
kevindew: This 404s you want: https://docs.publishing.service.gov.uk/repos/govuk-chat-gradio-prototype.html as it's not an app

terraform/deployments/github/repos.yml
```diff
@@ -298,6 +298,16 @@ repos:
...
+    required_status_checks:
+      standard_contexts: *standard_security_checks
```
kevindew: We can't have these because of the private repo.

kevindew (commented):
Looks like this needs a couple of changes - we'll need to get someone from platform-engineering to approve it.

terraform/deployments/github/repos.yml
davidgisbey: oh ofc sorry i did this before the seal PR. Will remove them.

davidgisbey:
Sorry about that. I've made those changes. I'll request a review.

AgaDufrat (approved):
Since this repo already exists, terraform apply will fail when it tries to create it. To avoid failure, we need to import the repo into the terraform state. Here is an example commit: https://github.com/alphagov/govuk-infrastructure/commit/c6774a7d42ca2eb9b0987a51cde8b57e13e0577f 
Since this only runs once you can re-use the existing block.

I will document this.

kevindew:
Thanks @davidgisbey shall we just put in a comment about skipping the standard checks so no-one thinks they just got missed?

AgaDufrat:
You may also need to set `visibility: private`.

davidgisbey:
Thanks both. @AgaDufrat i've pushed that second commit up so i'll merge if you're happy for me to.

PR #2639, opened 2025-08-21
Fix brakeman alerting

ChrisBAshton:
We had a [Brakeman security scan alert](https://github.com/alphagov/specialist-publisher/security/code-scanning/30) for Specialist Publisher, which turned out to be a false positive. We [added it to the ignore list](https://github.com/alphagov/specialist-publisher/pull/3340). But the alert didn't auto-resolve.

This appears to be because GitHub doesn't honour 'suppressions' in the generated SARIF file. So instead, we add a pre-processing step whereby any detected issue that is also 'suppressed' is purged from the SARIF file prior to uploading.

Tested by pointing a [branch of Specialist Publisher](https://github.com/alphagov/specialist-publisher/pull/3348) to the new workflow. We can see that the alert is removed on the 'branch' version.

Before:

> Screenshot 2025-08-21 at 13 22 50

After: 

> Screenshot 2025-08-21 at 13 22 46

### DIFF
--- a/.github/workflows/brakeman.yml
+++ b/.github/workflows/brakeman.yml
@@ -32,11 +32,17 @@ jobs:
             --no-exit-on-warn --no-exit-on-error \
             -o brakeman.json -o brakeman.sarif
 
+      - name: Strip suppressed results from SARIF
+        run: |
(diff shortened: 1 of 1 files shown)


AgaDufrat (approved):
This sparks joy ✨ Thanks!

PR #2642, opened 2025-08-21
Add "archived" option to Github Repo YAML.

dj-maisy:
## What?
This change should start allowing us to archive Repositories while also keeping them managed within our YAML file so they remain in IaC.

### Example
This is what happens if you try to archive a Repo which has PRs open (in case someone skips a step in the instructions):
```
╷
│ Error: Resource precondition failed
│ 
│   on main.tf line 211, in resource "github_repository" "govuk_repos":
│  211:       condition =  !try(each.value.archived, false) || length(data.github_repository_pull_requests.govuk_repos_prs[each.key].results) == 0
│     ├────────────────
│     │ data.github_repository_pull_requests.govuk_repos_prs is object with 134 attributes
│     │ each.key is "govuk-replatform-test-app"
│     │ each.value.archived is true
│ 
│ You cannot archive a Repo with open PRs. Review and close the PRs first.
```

A similar failure should also happen if GitHub pages are still configured.

### DIFF
--- a/terraform/deployments/github/main.tf
+++ b/terraform/deployments/github/main.tf
@@ -208,8 +208,16 @@ resource "github_repository" "govuk_repos" {
     ]
 
     precondition {
-      condition =  !try(each.value.archived, false) || length(data.github_repository_pull_requests.govuk_repos_prs[each.key].results) == 0
+      condition     = !try(each.value.archived, false) || length(data.github_repository_pull_requests.govuk_repos_prs[each.key].results) == 0
       error_message = "You cannot archive a Repo with open PRs. Review and close the PRs first."
+    }
+
+    precondition {
+      condition = (
+        !try(each.value.archived, false) ||
+        try(length(data.github_repository.govuk[format("alphagov/%s", each.key)].pages), 0) == 0
+      )
+      error_message = "You cannot archive a Repo with an active GitHub Pages Configuration. Remove this first."
     }
   }
 }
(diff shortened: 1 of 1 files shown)


terraform/deployments/github/main.tf
```diff
@@ -164,6 +164,12 @@ resource "github_team_repository" "ithc_repos" {
   permission = try(each.value.teams["govuk_ithc"], "pull")
 }
 
+data "github_repository_pull_requests" "govuk_repos_prs" {
```
jfharden, line 167: You can probably for_each this based on only repos that need to be archived, that should massively reduce the queries made to github?

dj-maisy:
The Precondition check still works with the changes:
```
╷
│ Error: Resource precondition failed
│ 
│   on main.tf line 215, in resource "github_repository" "govuk_repos":
│  215:       condition = (
│  216:         !try(each.value.archived, false) ||
│  217:         (
│  218:           contains(keys(data.github_repository_pull_requests.govuk_repos_prs), each.key)
│  219:           ? length(try(data.github_repository_pull_requests.govuk_repos_prs[each.key].results, [])) == 0
│  220:           : true
│  221:         )
│  222:       )
│     ├────────────────
│     │ data.github_repository_pull_requests.govuk_repos_prs is object with 1 attribute "govuk-replatform-test-app"
│     │ each.key is "govuk-replatform-test-app"
│     │ each.value.archived is true
│ 
│ You cannot archive a Repo with open PRs. Review and close the PRs first.
```

PR #2650, opened 2025-08-22
[SCH-1418] Separate production and integration GCP access

leenagupte:
At the moment the "govuk-gcp-access" google groups grants admin access to all three search api v2 gcp environments. This isn't ideal as on GOV.UK we have rules about when someone should be granted [production access].

A new google group has been created, `govuk-gcp-access-integration` that only gives members access to the Search API V2 integration project.

The existing `govuk-gcp-access` group has been made a member of `govuk-gcp-access-integration` so members of that group will still have access to integration.

These changes should mean that we can give new starters access to integration only whilst leaving the access for people with production access unchanged.

[production access]: https://docs.publishing.service.gov.uk/manual/rules-for-getting-production-access.html#when-you-get-production-admin-access


jfharden (approved):
🚀

PR #2651, opened 2025-08-22
Document adding new repos to GOV.UK GitHub Infrastructure config

AgaDufrat:
Those aspects are frequently overlooked.

### DIFF
--- a/terraform/deployments/github/README.md
+++ b/terraform/deployments/github/README.md
@@ -33,6 +33,64 @@ protecting the `main` branch.
   ```shell
   GITHUB_TOKEN=<token> gds aws govuk-production-poweruser -- terraform apply
   ```
+## Adding existing repositories
+
+To manage an existing repository using the GOV.UK GitHub Infrastructure configuration, it needs to be imported into terraform state. Otherwise terraform apply will fail attempting to create a repository that already exists. 
+
+To import the resource, use an import block:
+```
+import {
+  to = github_repository.govuk_repos["govuk-existing-repo-name"]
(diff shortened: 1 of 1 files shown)


terraform/deployments/github/README.md
```diff
@@ -33,6 +33,64 @@ protecting the `main` branch.
...
+```
+import {
+  to = github_repository.govuk_repos["govuk-existing-repo-name"]
+  id = "govuk-existing-repo-name"
+}
+```
+
+[Example commit of importing an existing repository](https://github.com/alphagov/govuk-infrastructure/commit/c6774a7d42ca2eb9b0987a51cde8b57e13e0577f). Note that the code only has to run once so it's ok to remove old entries.
+
+### Private reposotories
```
nicholsj:
```suggestion
### Private repositories 
```

terraform/deployments/github/README.md
```diff
@@ -33,6 +33,64 @@ protecting the `main` branch.
   ```shell
   GITHUB_TOKEN=<token> gds aws govuk-production-poweruser -- terraform apply
   ```
+## Adding existing repositories
+
+To manage an existing repository using the GOV.UK GitHub Infrastructure configuration, it needs to be imported into terraform state. Otherwise terraform apply will fail because it will attempt to create a repository that already exists.
```
nicholsj:
```suggestion
To manage an existing repository using the GOV.UK GitHub Infrastructure configuration, it needs to be imported into terraform state. Otherwise terraform apply will fail attempting to create a repository that already exists. 
```

PR #2665, opened 2025-08-27
Add `govuk-content-block-manager` repo

pezholio:
We're carrying out a spike to move the Content Block Manager code from Whitehall to its own application, so need a repo. The repo already exists, so we need to import it.


richardTowers (approved):

AgaDufrat (approved):

AgaDufrat:
You'll need to tag it to `govuk` topic before running terraform. I just noticed that the docs were recently updated and are now incorrect 😞

PR #2679, opened 2025-08-28
Update 'Report GitHub workflow run failure to Slack' README

AgaDufrat:
- Mention the default Slack message – It's not clear without looking into the code what message will be posted in Slack. This should ensure that the information is not duplicated in the additional message.
- Add usage example for workflow with multiple jobs - the `secrets` context is not available within the context of `jobs`


dj-maisy (approved):

AgaDufrat:
Removed `needs:` because it is not required

PR #2682, opened 2025-08-29
Retire markdown-toolbar-element (WHIT-2020)

ChrisBAshton:
https://gov-uk.atlassian.net/browse/WHIT-2020


terraform/deployments/github/repos.yml
```diff
@@ -1,4 +1,4 @@
-access_levels:
+alphagov/markdown-toolbar-element access_levels:
```
jfharden: I don't think this change is intentional right?

ChrisBAshton: Argh, GitHub UI fail. Good catch.

jfharden (approved):

PR #2683, opened 2025-08-29
Search API V2 allow rake tasks to use bucket

nacnudus:
Rake tasks run with the identity of a service account, which needs permission to create objects in a bucket. Required for https://github.com/alphagov/search-api-v2/pull/536

### DIFF
--- a/terraform/deployments/search-api-v2/automated_evaluation.tf
+++ b/terraform/deployments/search-api-v2/automated_evaluation.tf
@@ -83,7 +83,18 @@ resource "google_storage_bucket" "vais_evaluation_output" {
   location = var.gcp_region
 }
 
-#
+resource "google_storage_bucket_iam_member" "vais_evaluation_output_api_service_account_bucketviewer" {
+  bucket = google_storage_bucket.vais_evaluation_output.name
+  role   = "roles/storage.bucketViewer"
+  member = google_service_account.api.member
+}
+
+resource "google_storage_bucket_iam_member" "vais_evaluation_output_api_service_account_objectcreator" {
+  bucket = google_storage_bucket.vais_evaluation_output.name
+  role   = "roles/storage.objectCreator"
+  member = google_service_account.api.member
+}
+
 resource "google_storage_bucket_object" "qrels_seed_file" {
   name   = "ts=1970-01-01T00:00:00/qc=0/rc=0/judgement_list=sample/qrels.csv"
   bucket = google_storage_bucket.automated_evaluation_output.name


terraform/deployments/search-api-v2/automated_evaluation.tf
```diff
@@ -83,7 +83,18 @@ resource "google_storage_bucket" "vais_evaluation_output" {
   location = var.gcp_region
 }
 
-#
+resource "google_storage_bucket_iam_member" "vais_evaluation_output_api_service_account_bucketviewer" {
+  bucket = google_storage_bucket.vais_evaluation_output.name
+  role   = "roles/storage.bucketViewer"
+  member = google_service_account.api.member
```
hannako, line 89: I'm not familiar with GCP resources, so apologies for stupid questions! 

Your change seems sensible, but am not sure I understand why the `google_service_account.automated_evaluation_pipeline.member` is added to an IAM binding at the bottom of this file, but this `google_storage_bucket_iam_member` isn't. Is this fine? Will it work without adding a project id etc?

nacnudus: I think this is the section you're referring to. I've annotated each resource. The gist is that the existing code is giving rather broad permissions to the service account, whereas the new code is only giving it permissions in relation to one particular bucket.

```terraform
# This creates the service account
resource "google_service_account" "automated_evaluation_pipeline" {
  account_id   = "automated-evaluation-pipeline"
  display_name = "automated-evaluation-pipeline"
  description  = "Pipeline to trigger automated evaluation function, with permisssions to read google storage/ read and write to BQ / query vertex"
}

# This defines a set of permissions that can be granted to an account as a "role"
resource "google_project_iam_custom_role" "automated_evaluation_pipeline" {
  role_id     = "automated_evaluation_pipeline"
  title       = "automated_evaluation_pipeline"
  description = ""
  permissions = [
    "discoveryengine.servingConfigs.search",
    "storage.objects.get",
    "storage.objects.list",
    "storage.objects.create",
    "storage.objects.update",
    "bigquery.tables.update",
    "bigquery.tables.updateData",
    "bigquery.jobs.create",
    "bigquery.datasets.get",
    "bigquery.tables.get",
    "bigquery.tables.getData",
    "cloudfunctions.functions.invoke",
    "run.jobs.run",
    "run.routes.invoke",
    "cloudfunctions.functions.get"
  ]
}

# This grants the role to the service account at the "project" level. This means
# that the service account has permissions such as `storage.objects.update`
# on every bucket in this project, which is rather broad.
resource "google_project_iam_binding" "automated_evaluation_pipeline" {
  role    = google_project_iam_custom_role.automated_evaluation_pipeline.id
  project = var.gcp_project_id
  members = [
  google_service_account.automated_evaluation_pipeline.member]
}
```

hannako: Ah OK. That makes sense, thank you.

Next question!

Do your config changes mean that the Service Account gets these permissions (as it's defined as a member) but  we need to explicitly assign SearchAPIv2 a `google_storage_bucket_iam_member` for SearchAPIv2 to get these permissions?

nacnudus: I think the answer is no. The thought process (which may well be flawed):

* The error message was `search-api-v2@search-api-v2-integration.iam.gserviceaccount.com does not have storage.buckets.get access to the Google Cloud Storage bucket`.
* That service account is created by
https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/search-api-v2/service_accounts.tf#L1-L7
    ```terraform
    # Creates and configures service accounts, IAM roles, role bindings, and keys for `search-api-v2` to
    # be able to access the Discovery Engine API.
    resource "google_service_account" "api" {
      account_id   = "search-api-v2"
      display_name = "search-api-v2 (Rails API app and document sync worker)"
      description  = "Service account to provide access to the search-api-v2 Rails app and document sync worker"
    }
    ```
* Presumably the rails API app and tasks run as that service account. It's akin to the app being logged into that account.

hannako: Ah OK! Thanks for explaining all that 👍

hannako (approved):
Thanks for answering all my questions - makes much more sense now!

samsimpson1 (approved):

PR #2688, opened 2025-09-01
(CM-416) Rename Content Block Manager repo

pezholio:
We’ve now changed this to remove the prefix (which is only relevant to gems), as having to type `govuk-` each time would get tedious, and it’s not consistent with any of the other apps.

### DIFF
--- a/terraform/deployments/github/import.tf
+++ b/terraform/deployments/github/import.tf
@@ -1,4 +1,0 @@
-import {
-  to = github_repository.govuk_repos["govuk-content-block-manager"]
-  id = "govuk-content-block-manager"
-}
--- a/terraform/deployments/github/main.tf
+++ b/terraform/deployments/github/main.tf
@@ -174,6 +174,11 @@ data "github_repository_pull_requests" "govuk_repos_prs" {
   state           = "open"
 }
 
+moved {
+  from = github_repository.govuk_repos["govuk-content-block-manager"]
+  to   = github_repository.govuk_repos["content-block-manager"]
+}
+
 resource "github_repository" "govuk_repos" {
   for_each = local.repositories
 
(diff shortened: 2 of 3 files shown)


MuriloDalRi (approved):

pezholio:
Oh yikes. Good catch.

jfharden:
> This will cause terraform to delete the repo. Will write more in a sec but want to stop this happening accidentally

You can see in the terraform plan for this, it will delete the govuk-content-block-manager repo and create a new one called content-block-manager

Screenshot 2025-09-01 at 11 11 53

If you are ok with the entire repo, all it's code, history, PRs, issues, etc being deleted I can approve it, but I want to make sure this is intentional, and we don't accidentally lose the codebase.

pezholio:
Ah, no, I definitely don't want this 😅 I've renamed the repo this side (https://github.com/alphagov/content-block-manager), but will Github still recognise the app as it has had that name in the past? (much how the redirects work)

jfharden:
Given the yikes I assume this isn't intended :) 

We can add a [terraform `moved` block](https://developer.hashicorp.com/terraform/language/moved) into the [terraform definition](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/github/main.tf#L177) to tell it the resource is changing name, something like:

```tf
moved {
  from = github_repository.govuk_repos["govuk-content-block-manager"]
  to      = github_repository.govuk_repos["content-block-manager"]
}
```

Then a new plan should hopefully look much less destructive

jfharden:
Ah the new plan looks generally better, however it turns out we still have an import to bring it in with the old name, if you just delete the file https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/github/import.tf it should be fine then

edit: so many typos 🫠

pezholio:
@jfharden Should be good now 👍

Issue #2644, opened 2025-08-22
Create a new TF root for EKS user access


samsimpson1:
**User Need**

**As an** EKS cluster administrator
**I want** EKS user access management to be done in a separate Terraform root
**so that** I can automatically apply changes to user access easily

---

### What’s Needed

List anything the solution must do or be (behaviour, performance, security, UX, etc.).

- [x] Create a new terraform root in govuk-infrastructure for EKS user management (`cluster-access` perhaps?)
- [x] Move all EKS access code from `cluster-services` (EKS access entries, K8s roles, etc)
- [x] Configure this workspace to not require manual approval for applies on `main`
- [x] Configure the user management TF to automatically trigger runs of these new TF workspaces

---

### Acceptance Criteria

- [x] Changes are sufficiently documented. Existing docs are updated, new docs are created as required
- [x] EKS access entries are automatically created/destroyed when the user management TF is modified
- [x] EKS access management TF can be ran without interference from other TF code 
- [x] TF run order documentation has been updated

PR #2699, opened 2025-09-01
cluster-access depends on cluster-services, not vice-versa

jfharden:
cluster-access requires the namespaces to have already been created, so it cannot be created prior to cluster-services (which creates the namespaces), otherwise you get errors as in [this ephemeral cluster creation run](https://app.terraform.io/app/govuk/workspaces/cluster-access-eph-jfharden-2025-09-01/runs/run-ZDxa6sKJr6sbyLcy)

```
Error: namespaces "datagovuk" not found
with kubernetes_role.developer["datagovuk"]
on eks_access.tf line 204, in resource "kubernetes_role" "developer":
resource "kubernetes_role" "developer" {
```

After this change, spinning up an ephemeral cluster works again

### DIFF
--- a/terraform/deployments/ephemeral/tfe.tf
+++ b/terraform/deployments/ephemeral/tfe.tf
@@ -80,7 +80,7 @@ module "cluster_access" {
     ship_kubernetes_events_to_logit = false
   }
 
-  depends_on = [module.cluster_infrastructure, tfe_project.project]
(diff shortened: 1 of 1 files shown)


samsimpson1 (approved):

samsimpson1:
Related to https://github.com/alphagov/govuk-infrastructure/issues/2644

PR #2702, opened 2025-09-02
Add separate IP based rate limiting for chat domain

rtrinque:
At the moment all traffic that comes into the backend load balancer is subject to an IP based rate limiting at the WAF level. This is a problem for chat as we expect all traffic to come from a single source of traffic -the gov.uk app- therefore to solve this issue and add more flexibility in controlling the traffic for chat we add new rules that introduce a separate rate limiting when traffic is going towards the chat domain


AgaDufrat:
Hi @rtrinque, do you still need a review on this. It's been open for a while.

Issue #2598, opened 2025-08-15
🧪 Create synthetic tests to capturing release pipeline failures


jaskaransarkaria:
**User Need**

**As a** Platform Engineer
**I want** to be aware when applications cannot be deployed through to production
**so that** we can pipeline failures effectively and quickly

---

### What’s Needed

- [ ] establish a testing framework from which we can perform synthetic tests that mimic a user deploying a change to production
- [ ] run this test periodically
- [ ] test the change makes it's way through to production
- [ ] alert engineers if this test fails (probably in slack)

---

### Acceptance Criteria

- [ ] Everything in **What’s Needed** is delivered.
- [ ] Any changes during delivery are reflected in this story.

---

### Risks & Mitigation

- This will help reduce the risk of our deployment pipeline line being broken and us as a platform not being aware of it.

---

### Notes

It's not clear exactly where these tests will run. Although concourse is on the near horizon.

It seems logical that this test will grow into a fuller test suite, capturing user behaviour accross the release pipelines and cluster generally, and that this suite would eventually be ran and viewed in concourse.

kentsanggds:
Forked the replatforming test app https://github.com/alphagov/govuk-synthetic-test-app/tree/add-synthetic-test-cronjob, and have managed to trigger updates on the branch which update the version file and deploys image with tags matching the version.

Will now try to figure out how to test it on the cluster pipeline.

kentsanggds:
Trying a slightly different approach using the app to trigger updates to the version and to push up changes to github and to create tags - https://github.com/alphagov/govuk-synthetic-test-app/tree/update-version, this seems to be better as running the trigger from within the github actions workflow caused problems with trying to tag the correct commit.

kentsanggds:
New card created for the infra work required for the synthetic-test roles - https://github.com/alphagov/govuk-infrastructure/issues/2724

AgaDufrat:
[Stand Up] 1 more issue to be written up

jaskaransarkaria:
Issue for writing the actual test suite code https://github.com/alphagov/govuk-infrastructure/issues/2733

kentsanggds:
Issue for creating the synthetic test runner repo and cronjob - https://github.com/alphagov/govuk-infrastructure/issues/2736

kentsanggds:
[...]

PR #2707, opened 2025-09-02
Add govuk-synthetic-test-app to extra_repo

kentsanggds:
Adding the govuk-synthetic-test-app repo to the extra_repositories block because it isn't being picked up possibly because when it was imported into the github terraform state the `govuk` topic wasn't set.

https://github.com/alphagov/govuk-infrastructure/issues/2598


kentsanggds:
Not needed as discovered that forks don't appear in the search results, fix was to remove the fork relationship

PR #2735, opened 2025-09-05
Allow govuk-fastly-diff-generator to be pulled by ECR

jfharden:
The diff generator needs to be pulled by lambda, so the ecr repo policy needs to allow lambda to pull it!

This solution is a generalised approach which means any repo can be added to the list which can be pulled by lambda.

### DIFF
--- a/terraform/deployments/ecr/main.tf
+++ b/terraform/deployments/ecr/main.tf
@@ -67,6 +67,11 @@ locals {
     "search-api-learn-to-rank",
     "toolbox",
   ]
+
+  // Add repos to this list if you want lambda to have permission to pull the image
+  allow_lambda_pull = [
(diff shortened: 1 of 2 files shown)


terraform/deployments/ecr/main.tf
```diff
@@ -55,7 +55,6 @@ locals {
 
   extra_repositories = [
     "clamav",
-    "govuk-e2e-tests",
```
jfharden, line 58: This is removed because it's already in the regular list. The toset operations done on this have been hiding the duplication.

The fact the TF plan output for this PR shows no changes to govuk-e2e-tests proves this is duplicated

jaskaransarkaria (approved):

MuriloDalRi (approved):

PR #2738, opened 2025-09-08
Add govuk-chat-gradio-user-research repo to terraform

davidgisbey:
## Description

This is a new repo that will be used to build repid prototypes for user research sessions.

I've copied the config from the existing govuk-chat-gradio-prototype repo as we will be using the same CI actions etc.

## Trello card

https://trello.com/c/lYX0DFvQ/2713-create-prototypes-for-research-on-response-latency

### DIFF
--- a/terraform/deployments/github/repos.yml
+++ b/terraform/deployments/github/repos.yml
@@ -315,6 +315,17 @@ repos:
         - Format
         - Type checks
 
+  govuk-chat-gradio-user-research:
+    visibility: private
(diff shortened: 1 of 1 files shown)


AgaDufrat (approved):
LGTM. I think this is the first time someone is creating a new repo with this (rather than adding an existing repo) 🍀 
You may need to add the `govuk` topic to the repo in the UI and then kick off terreform again for some of the config to be applied.

PR #2790, opened 2025-09-11
Add "Service: govuk-chat" tag to Chat resources

ianhowell-gds:
## What

Add `Service: govuk-chat` tag to Chat resources

## Why

So Cost Allocation can be applied specifically to Chat resources

### DIFF
--- a/terraform/deployments/opensearch/main.tf
+++ b/terraform/deployments/opensearch/main.tf
@@ -25,6 +25,7 @@ provider "aws" {
       owner                = "govuk-platform-engineering@digital.cabinet-office.gov.uk"
       repository           = "govuk-infrastructure"
       terraform-deployment = basename(abspath(path.root))
+      Service              = "govuk-chat"
     }
   }
 }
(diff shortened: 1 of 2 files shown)


terraform/deployments/opensearch/main.tf
```diff
@@ -25,6 +25,7 @@ provider "aws" {
       owner                = "govuk-platform-engineering@digital.cabinet-office.gov.uk"
       repository           = "govuk-infrastructure"
       terraform-deployment = basename(abspath(path.root))
+      Service              = "govuk-chat"
```
samsimpson1, line 28: This resource already has a service tag, and the system this resource is a part of is already tagged in accordance with the GDS Way https://gds-way.digital.cabinet-office.gov.uk/manuals/aws-tagging.html#mandatory

ianhowell-gds: Unfortunately, tags are case sensitive, and the Cost Allocation tags actived by Engineering Enablement do not include "service", so "System" seems to be the best option as agreed with Jason Birchall.

samsimpson1: Ah that makes sense. Looks like we're going to have to update our tags to match

PR #2797, opened 2025-09-12
Optionally generate release notes

pezholio:
This adds an input to the Release workflow to automatically generate release notes (see https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes).

While we practice continuous delivery, so one PR usually results in one release, this could still be useful to link releases to PRs. For now, I’ve made it optional with a default of true, but I’m happy to set it to `true` for our project and see how we go.

I did this manually for one of our releases, and this is how it looks - https://github.com/alphagov/content-block-manager/releases/tag/v37

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,6 +2,11 @@ name: Release
 
 on:
   workflow_call:
+    inputs:
+      generate_release_notes:
+        required: false
+        type: boolean
+        default: false
     secrets:
       GH_TOKEN:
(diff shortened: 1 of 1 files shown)


AgaDufrat:
We tend to use the [Release app](https://release.publishing.service.gov.uk/) for this information and we're [planning to rebuild our CI/CD pipeline](https://github.com/alphagov/govuk-infrastructure/milestone/2).

PR #2820, opened 2025-09-16
Import chat rds snapshot

jfharden:
The snapshot worked, but the tf timed out. Import it. We need to apply this only in integration, then remove the import block


jfharden:
We know staging and prod are erroring, and that's ok, will revert as soon as the import is run

Issue #2819, opened 2025-09-16
🔒 Add aws:SecureTransport enforcement to Chat S3 buckets


jasonBirchall:
This issue connects to https://github.com/alphagov/govuk-platform-internal/issues/62

jasonBirchall:
@ianhowell-gds, are you happy for this to be closed?

ianhowell-gds:
Yes, thank you @jasonBirchall - the overnight jobs that need access to the Chat Opensearch Snapshots ran successfully.

PR #2822, opened 2025-09-16
Configure explicit encryption in transit for Chat S3 buckets

ianhowell-gds:
## What

Add explicit encryption in transit in the bucket policy for Chat S3 buckets

## Why

This is to comply with the ITHC security recommendation that we ensure all access to the buckets are encrypted in transit


jasonBirchall:
Connects to https://github.com/alphagov/govuk-infrastructure/issues/2819

Issue #2724, opened 2025-09-04
Create synthetic-test roles to access all environments from Production


kentsanggds:
**User Need**

**As a** platform engineer
**I want** to be allow apps to assume a synthetic test role
**so that** I can run a synthetic test for the CI/CD pipeline on all environments

---

### What’s Needed

List anything the solution must do or be (behaviour, performance, security, UX, etc.).

- [ ] create a synthetic-test assumer role that can access the k8s api
- [ ] create a synthetic-test assumed role 
- [ ] update govuk-helm-charts so that the roles can access things
- [ ] limit scope of permissions to just have GET access to the k8s api
- [ ] allow access to all environments from Production only
- [ ] non-production environments should be able to access the k8s api in their own environment only

---

### Acceptance Criteria

- [ ] Everything in **What’s Needed** is delivered.
- [ ] Any changes during delivery are reflected in this story.
- [ ] Impacted users have been informed appropriately (e.g. change logs, Slack, docs).
- [ ] The synthetic test framework app can assume the role to access the k8s api in each environment in the production environment

---

### Notes

This role will be very similar to previous work done for the release assumer and assumed roles - 

https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/release/assumer.tf
https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/release/assumed.tf
https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/tfc-configuration/release.tf

https://github.com/alphagov/govuk-helm-charts/tree/main/charts/release/templates
https://github.com/alphagov/govuk-helm-charts/pull/3188/files

---

jaskaransarkaria:
The tests currently need permissions to:

- get pods in "apps" ns
- get and patch Application CR in the "cluster-services" ns

jasonBirchall:
- [x] TODO: @jasonBirchall create an EPIC for the synthetic test application work.

kentsanggds:
Finally managed to get some golang working and bringing back a response from the EKS kubernetes cluster using the assumed role. Need to figure out how to deserialise the response so that it can be parsed.

kentsanggds:
Managed to deserialise the response into a struct which can be parsed so just need to create basic tests in ginkgo to test the boundaries of the synthetic test assumed role.

PR #2856, opened 2025-09-22
Add synthetic-test assumer and assumed roles.

kentsanggds:
The synthetic test assumed role would allow the production assumer to assume roles in non-prod accounts. 

Non-prod account assumers can only assume into their own account.

This configuration was largely based off the work on the Release app roles.

https://github.com/alphagov/govuk-infrastructure/issues/2724


terraform/deployments/synthetic-test/assumed.tf
```diff
@@ -0,0 +1,55 @@
...
+
+data "aws_iam_policy_document" "synthetic_test_assumed_assume" {
+  dynamic "statement" {
+    for_each = toset(local.assumed_identifiers)
+    content {
+      actions = ["sts:AssumeRole"]
+      effect  = "Allow"
+      principals {
+        type        = "AWS"
+        identifiers = [statement.key]
```
jfharden: You don't need the dynamic in here, you can just do `idenfitiers = local.assumed_identifiers`

jfharden (approved):

PR #2857, opened 2025-09-23
Remove Legacy NATs and Public Subnets

dj-maisy:
## What?
This PR makes the following changes:

* Removes Pre-Kubernetes "Legacy" NAT Gateways
* Removes Pre-Kubernetes "Legacy" Public Subnets
* Removes any redundant "local" variables, hard-code local "AZs" variable.
* Remove unused TF outputs
* Make the VPC and IG "Name" tags more verbose, for easier identification in AWS Console

### Not in Scope
I have left the "Private Subnets" largely alone as these are used by RDS, ElastiCache and ElasticSearch and could/should be refactored in future into their respective Deployments/Modules.

### TF Plan

[Plan can be reviewed here](https://app.terraform.io/app/govuk/vpc-integration/runs/run-sW7yEBBiFgXVoiZD)

### DIFF
--- a/terraform/deployments/vpc/subnets.tf
+++ b/terraform/deployments/vpc/subnets.tf
@@ -1,9 +1,26 @@
 locals {
-  azs                        = { for name, subnet in var.legacy_public_subnets : subnet.az => name }
-  nat_legacy_private_subnets = { for name, subnet in var.legacy_private_subnets : name => subnet if subnet["nat"] }
+  azs = [
+    "eu-west-1a",
+    "eu-west-1b",
+    "eu-west-1c"
+  ]
 }
 
 # Private subnets
+
+# Preserve EIPs from Legacy Subnets/NATs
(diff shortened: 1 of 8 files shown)


terraform/deployments/vpc/subnets.tf
```diff
@@ -1,10 +1,27 @@
...
+    "eu-west-1a",
+    "eu-west-1b",
+    "eu-west-1c"
+  ]
 }
 
 # Private subnets
 
+# Preserve EIPs from Legacy Subnets/NATs
+resource "aws_eip" "private_subnet_nat" {
```
AP-Hunt, line 12: Will this need an import to bring it under management?

dj-maisy: Nope, as it has only moved ~down~ up a few lines and the resource has not changed (except the tags). The Git diff is bit misleading.

PR #2867, opened 2025-09-24
Set up "Blue" Node Groups in all envs

dj-maisy:
## What?
This is the next step in fixing the NodeGroup config - add the "blue" Node Group to all environments.

I should have added as well that in Production I want to launch the new Node Group with `r8g.2xlarge` size to try and right-size the Nodes for our current workloads.

The following PR will be to gracefully cycle down the old NodeGroup and rename that to "green" once "blue" has spun up. Any subsequent updates or changes to NodeGroups should be a fair bit easier after this is done.

### DIFF
--- a/terraform/deployments/tfc-configuration/variables-production.tf
+++ b/terraform/deployments/tfc-configuration/variables-production.tf
@@ -51,14 +51,15 @@ module "variable-set-production" {
 
     enable_kube_state_metrics = false
 
-    enable_arm_workers = true
-    enable_x86_workers = true
+    enable_arm_workers      = true
+    enable_arm_workers_blue = true
+    enable_x86_workers      = true
 
     publishing_service_domain = "publishing.service.gov.uk"
 
-    arm_workers_instance_types  = ["r8g.4xlarge", "r7g.4xlarge", "m7g.8xlarge", "m6g.8xlarge"]
-    main_workers_instance_types = ["m6i.8xlarge", "m6a.8xlarge"]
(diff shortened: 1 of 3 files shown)


terraform/deployments/tfc-configuration/variables-production.tf
```diff
@@ -51,14 +51,15 @@ module "variable-set-production" {
...
+    enable_arm_workers_blue = true
+    enable_x86_workers      = true
 
     publishing_service_domain = "publishing.service.gov.uk"
 
-    arm_workers_instance_types  = ["r8g.4xlarge", "r7g.4xlarge", "m7g.8xlarge", "m6g.8xlarge"]
-    main_workers_instance_types = ["m6i.8xlarge", "m6a.8xlarge"]
-    x86_workers_instance_types  = ["r7i.large", "r7a.large", "m7i-flex.xlarge", "m6a.xlarge", "m6i.xlarge"]
+    arm_workers_instance_types      = ["r8g.4xlarge", "r7g.4xlarge", "m7g.8xlarge", "m6g.8xlarge"]
+    arm_workers_blue_instance_types = ["r8g.2xlarge"]
```
jfharden, line 61: Don't you want this to be the same types as above?

dj-maisy: Actually no because I was _originally_ attempting to downsize the nodes from `r8g.4xlarge` down to `r8g.2xlarge`. At present we're bottlenecking at 110 Pods well before we reach memory capacity. Then I discovered that making this change without risking downtime was impossible due to an issue with the current version of the EKS Module, unless I changed how we were configuring the Node Groups.

We might end up having a couple more nodes in the pool, but it should be cheaper overall.

PR #2869, opened 2025-09-25
Archived repos cannot have vulnerability reports

jfharden:
Since archived repos cannot have vulnerability reports, and having it set to true causes perpetual changes in the terraform, if a repo is archived then set vulnerability reports to false


MuriloDalRi (approved):

jfharden:
There is still 1 other change but it's unrelated and this is a good standalone change so I'll merge as is

Issue #2736, opened 2025-09-05
Create synthetic test runner repo and cronjob to run it on the cluster


kentsanggds:
**User Need**

**As a** platform engineer
**I want** a cronjob for the CI/CD synthetic test runner on the cluster
**so that** I will be alerted if the pipeline is broken

---

### What’s Needed

- [ ] add the synthetic test runner repo to https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/github/repos.yml
  - this repo should have a github action to create an image and push it up to GHCR
  - also will need a github action to update the image tag in govuk-helm-charts
  
- [ ] create a new synthetic test runner cronjob as a chart here https://github.com/alphagov/govuk-helm-charts/tree/main/charts
  - this cronjob will simply pull the image created in GHCR and run it periodically, probably once a day?
  - image version of the synthetic test runner to pull should be created for each environment - https://github.com/alphagov/govuk-helm-charts/tree/main/charts/app-config/image-tags

---

### Acceptance Criteria

- [ ] a test image is created in GHCR for the synthetic test runner
- [ ] the repo should have an initial test app which will demonstrate that it can access the k8s api endpoint
- [ ] the cronjob runs once a day and can pull the image from GHCR and run it in each environment 
  - running it on Integration and Staging just for testing purposes
  - running on Production will track a deployment through Integration, Staging and Production

---

### User Comms Plan

Not needed until the synthetic test runner is in place

### Notes

There is prior art in assuming roles and connecting to the k8s api endpoint in the Release app.

---

kentsanggds:
Repo is available - https://github.com/alphagov/govuk-synthetic-test-app-canary

jasonBirchall:
[pre-planning] we feel like it's getting closer to completion so will carry it over.

kentsanggds:
Follow a chat with the team, decided to pull the image from `ECR` as it is currently pulling the image from `GHCR`

kentsanggds:
The cronjob PR should be able to still pull a working an image which is able to run the ginkgo tests.
If the ECR pull through cache is still not working there will be another issue created to look into it but for now the ginkgo tests should be able to run as there are some manually pulled through images which will expire after 30 days.

kentsanggds:
Issue created to investigate why the ECR pull through cache is not working - https://github.com/alphagov/govuk-infrastructure/issues/3052

PR #2971, opened 2025-10-14
Add a canary repo to be used by the synthetic test app

kentsanggds:
This will be a very simple app that will work with the Ginkgo tests in the synthetic test app repo to check that the release pipeline is working.

https://github.com/alphagov/govuk-infrastructure/issues/2736

### DIFF
--- a/terraform/deployments/github/repos.yml
+++ b/terraform/deployments/github/repos.yml
@@ -914,6 +914,7 @@ repos:
     required_status_checks:
       additional_contexts:
         - Test
+    standard_contexts: *standard_security_checks
 
   govuk-user-reviewer:
     visibility: private


terraform/deployments/github/repos.yml
```diff
@@ -909,6 +909,12 @@ repos:
       additional_contexts:
         - Test
 
+  govuk-synthetic-test-app-canary:
+    can_be_deployed: true
+    required_status_checks:
```
AgaDufrat, line 914: Does it need `standard_security_checks`? https://github.com/alphagov/govuk-infrastructure/blob/7a789203d6cf9db545fa9f4358eda5e9eaee1fba/terraform/deployments/github/repos.yml#L2-L4

kentsanggds: yes, probably a good idea, I'll add it in

PR #2987, opened 2025-10-15
step 1 in archiving the govuk-analytics-engineering repo

Nyzl:
I'm archiving this repo because it is not used and not maintained

### DIFF
--- a/terraform/deployments/github/repos.yml
+++ b/terraform/deployments/github/repos.yml
@@ -283,7 +283,9 @@ repos:
       additional_contexts:
         - test
 
-  govuk-analytics-engineering: {}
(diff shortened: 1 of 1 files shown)


terraform/deployments/github/repos.yml
```diff
@@ -283,8 +283,7 @@ repos:
       additional_contexts:
         - test
 
-  govuk-analytics-engineering:
-    visibility: private
+  govuk-analytics-engineering: {}
```
AgaDufrat: You may want to keep it private. Then it will be just one step to archive the repo.

```suggestion
  govuk-analytics-engineering:
    visibility: private
    archived: true
```

PR #2989, opened 2025-10-16
Mention visibility when archiving repos

AgaDufrat:
There maybe a situation when we want to keep archived repo private


nimalank7 (approved):
LGTM!

PR #3036, opened 2025-10-22
Set log_autovauum_min_duration to pending_reboot apply method to stop continual drift

jfharden:
Every terraform plan for this parameter setting is trying to change it from `pending_reboot` to `immediate`, to lets set the terraform to reflect the reality and stop seeing `1 change` on every single terraform plan for all RDS workspaces.


dj-maisy (approved):

dj-maisy:
I'm surprised `tflint` didn't get the big sad about two properties being on one line inside a single set of brackets. 😂

jfharden:
> I'm surprised `tflint` didn't get the big sad about two properties being on one line inside a single set of brackets. 😂

Me too, other lines around this are using the same format so I followed them rather than change them all, and tflint was surprisingly nonplussed :D

PR #3062, opened 2025-10-27
Remove Content Publisher [WHIT-2441]

ChrisBAshton:
Blocked on #3066. 

Content Publisher is now retired.

Jira: https://gov-uk.atlassian.net/browse/WHIT-2441


terraform/deployments/govuk-publishing-infrastructure/content_publisher_s3.tf
```diff
@@ -1,60 +0,0 @@
-resource "aws_s3_bucket" "content_publisher_activestorage" {
```
jfharden, line 1: One thing here, terraform won't be able to delete the bucket until it's emtpy. You'll need to log in as fulladmin and follow one of the methods in https://docs.aws.amazon.com/AmazonS3/latest/userguide/empty-bucket.html

Also it's very likely to trigger a cyber alert, so prior to deleting the bucket you should submit a cyber notification to let cyber know.

PR #3127, opened 2025-11-05
Create "Green" ES 6.8 Clusters

dj-maisy:
## What?
This creates a separate Terraform Deployment so we can start running additional Elasticsearch 6.8 clusters side-by-side with the existing ones to enable the search team to perform a migration and step away from the older "blue" clusters.

### DIFF
--- a/terraform/deployments/elasticsearch-green/main.tf
+++ b/terraform/deployments/elasticsearch-green/main.tf
@@ -102,7 +102,7 @@ resource "aws_elasticsearch_domain" "opensearch" {
   }
 
   encrypt_at_rest {
-    enabled = var.encryption_at_rest
+    enabled = true
   }
 
   domain_endpoint_options {
--- a/terraform/deployments/elasticsearch-green/variables.tf
+++ b/terraform/deployments/elasticsearch-green/variables.tf
@@ -47,11 +47,6 @@ variable "ebs" {
   default     = null
 }
 
-variable "encryption_at_rest" {
-  type        = bool
-  description = "Enable encryption at rest in the opensearch cluster"
-}
-
 variable "tls_security_policy" {
   type        = string
   description = "The pre-canned AWS security policy to enforce for connections to opensearch"


terraform/deployments/elasticsearch-green/iam.tf
```diff
@@ -0,0 +1,63 @@
...
+}
+
+resource "aws_iam_role_policy_attachment" "manual_snapshot_role_policy" {
+  role       = aws_iam_role.manual_snapshot_role.name
+  policy_arn = aws_iam_policy.manual_snapshot_bucket_policy.arn
+
+}
+
+resource "aws_iam_policy" "can_configure_es_snapshots" {
+  name        = "govuk-${var.govuk_environment}-${var.stackname}-elasticsearch6-manual-snapshot-domain-configuration-policy"
```
AP-Hunt, line 43: Double check this isn't longer than allowed

jfharden:
If feels unusual to have a whole separate terraform deployment for green & blue, rather than the same module used mutliple times (potentially with different terraform roots) with different parameters.

It seems incredibly likely that there will either be now, or in the future, unknown differences between the blue and the green clusters since their terraform code (as opposed to the actual state) will drift apart from each other.

terraform/deployments/elasticsearch-green/iam.tf
dj-maisy: An IAM Policy name [can be up to 128 characters in length](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html#reference_iam-quotas-entity-length). This should fit IMO.

terraform/deployments/elasticsearch-green/main.tf
```diff
@@ -0,0 +1,177 @@
...
+      availability_zone_count = var.zone_awareness_enabled ? length(local.subnet_ids) : null
+    }
+  }
+
+  advanced_security_options {
+    enabled = false
+  }
+
+  encrypt_at_rest {
+    enabled = var.encryption_at_rest
```
AP-Hunt: Should this is be always true?

dj-maisy:
> If feels unusual to have a whole separate terraform deployment for green & blue, rather than the same module used mutliple times (potentially with different terraform roots) with different parameters.
> 
> It seems incredibly likely that there will either be now, or in the future, unknown differences between the blue and the green clusters since their terraform code (as opposed to the actual state) will drift apart from each other.

Absolutely, ordinarily I would agree - but unlike other resources such as our RDS instances which have been designed to be repeatable, our Elasticsearch Resources haven't exactly been set up to do this and this "green" cluster is somewhat temporary as it is a stepping-stone to a later upgrade. This just feels like a quick way to migrate from the existing "blue" deployment to a like-for-like "green" deployment unless we believe this should all just exist in the same deployment, but I think there would be a larger need for refactoring here if we wanted to do it this way.

This is effectively a warm-up for when the "hairier" upgrade to one of the newer major versions which would be a breaking change.

My suggestion is that when the work begins on the migration to ES/OS 7 or 8, we create a new module-style deployment for Elasticsearch to make it configurable more like a module rather than the current one-off set of resources.

terraform/deployments/elasticsearch-green/main.tf
dj-maisy: Will set the variable to `default = true` to address this.

jfharden:
> > If feels unusual to have a whole separate terraform deployment for green & blue, rather than the same module used mutliple times (potentially with different terraform roots) with different parameters.
> > It seems incredibly likely that there will either be now, or in the future, unknown differences between the blue and the green clusters since their terraform code (as opposed to the actual state) will drift apart from each other.
> 
> Absolutely, ordinarily I would agree - but unlike other resources such as our RDS instances which have been designed to be repeatable, our Elasticsearch Resources haven't exactly been set up to do this and this "green" cluster is somewhat temporary as it is a stepping-stone to a later upgrade. This just feels like a quick way to migrate from the existing "blue" deployment to a like-for-like "green" deployment unless we believe this should all just exist in the same deployment, but I think there would be a larger need for refactoring here if we wanted to do it this way.
> 
> This is effectively a warm-up for when the "hairier" upgrade to one of the newer major versions which would be a breaking change.
> 
> My suggestion is that when the work begins on the migration to ES/OS 7 or 8, we create a new module-style deployment for Elasticsearch to make it configurable more like a module rather than the current one-off set of resources.

In that case I'd call this module blue-green and use it for both blue now, and green later, otherwise there's going to be at least 3 copies of all of this code which is incredibly unlikely to ever be in sync in the future, and also confusing for people coming to it.

The code you have here seems to tolerate being blue or green (given you take stack name as a variable) so it should just be a renaming of the module.

Then a future PR (when they go up a major version) should be able to remove all the old module code and just stick with this blue-green module

Issue #3052, opened 2025-10-24
Investigate why the ECR pull through cache for govuk-synthetic-test-app repo is not working


kentsanggds:
**User Need**

**As a** platform engineer
**I want** the latest release build for govuk-synthetic-test-app to be available in ECR
**so that** I can run the latest synthetic tests without having to manually pull the image from GHCR

---

### What’s Needed

List anything the solution must do or be (behaviour, performance, security, UX, etc.).

- [ ] investigate why the pull through cache stopped working on the 26th September for release [v1756914788](https://github.com/orgs/alphagov/packages/container/govuk%2Fgovuk-synthetic-test-app/527446296?tag=v1756914788) - this might not be available after 30 days
- [ ] look at logs to see if there are any error messages, kubernetes events logs might be a starting place
- [ ] fix the ECR pull through cache
- [ ] update the README in [govuk-synthetic-test-app](https://github.com/alphagov/govuk-synthetic-test-app/blob/main/README.md) to remove the section regarding manual trigger of the ECR pull through cache.

---

### Acceptance Criteria

- [ ] Any new images are correctly pulled from ECR
- [ ] document any findings about why the ECR pull through cache is not working to prevent other repos from not being able to pull images through
- [ ] We will present this to the team in the sprint review

---

### Assumptions (optional)

- ECR pull through cache is working for all repos except govuk-synthetic-test-app repo.

---

### Notes

  - the manual trigger is working and documented in the README
  - there may be an issue with the infra code adding the repo to the list of [deployable repos](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/ecr/main.tf#L51-L71) but  
when I tried to add govuk-synthetic-test-app as in the `extra_repositories` list TFC complained of duplicate entries.

kentsanggds:
I wonder if its because for the pull through cache to bring in the image from GHCR, the image tag in the govuk-helm chart needs to be updated for the test, as it is not currently updated by the deploy workflow, the tag needs to be manually updated so that the ECR pull through happens.

kentsanggds:
Yes this theory was correct, I've proven this by updating the local manifest to refer to the latest image tag on ghcr and it pulled that image into ECR. This would normally be done by updating the image tags in the image files for each app but as the cronjob needs to be updated manually this isn't done automatically. I'll update the README in the govuk.synthetic-test-app to provide some context around this.

The govuk-synthetic-test-runner repo can be retired as it is not needed.

PR #3128, opened 2025-11-05
Archive govuk-synthetic-test-runner

kentsanggds:
This was created as it was initially thought that the ECR pull through was not working for govuk-synthetic-test-app but this was because there wasn't a demand for the latest image tag so the image wasn't pulled through.

As the ECR pull through is now working and that repo has more historical context this repo will be retired.

https://github.com/alphagov/govuk-infrastructure/issues/3052


terraform/deployments/github/repos.yml
```diff
@@ -910,11 +910,7 @@ repos:
     standard_contexts: *standard_security_checks
 
   govuk-synthetic-test-runner:
-    can_be_deployed: true
-    required_status_checks:
-      additional_contexts:
-        - Test
-    standard_contexts: *standard_security_checks
+    archived: true
```
AgaDufrat, line 913: You will need to do this in 2 PRs as described in the README: https://github.com/alphagov/govuk-infrastructure/tree/main/terraform/deployments/github#archiving-repositories

kentsanggds: OK, I'll split it up

PR #3130, opened 2025-11-06
Test out launching new DBs

jfharden:
Now prepare has been run, lets test the second stage of launching new instances


jfharden:
> Are we not missing 1? I thought we set prepare_to_launch_new_db for more than 24 DB in integration.

No, I just double checked manually (I did the change with search and replace) and everything that had prepare set now has launch set. There is 1 db which doesn't have either set, but that has neither prepare nor launch set.

PR #3212, opened 2025-11-19
[PP-7451] Allow language specification for CodeQL

yndajas:
The CodeQL action decides which languages to analyse based on the file extensions present in a repo. However, if it finds no analysable code for one of those languages, it will error

https://docs.github.com/en/code-security/code-scanning/troubleshooting-code-scanning/no-source-code-seen-during-build

This change will allow us to specify which languages should be analysed so that we can exclude a language for which there's no analysable code

https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/customizing-your-advanced-setup-for-code-scanning#changing-the-languages-that-are-analyzed

This does mean that if analysable code is later added to a repo where its language has been ignored, we wouldn't be analysing it. That doesn't feel great, but this seems to be a limitation of CodeQL

The following PR exhibited this error: https://github.com/alphagov/govuk-content-api-docs/pull/204. We have .js files in that repo but they only contain (magic) comments. This wasn't an issue with older versions of CodeQL. We're currently using 2.23.5 - the last merged PR used 2.23.2 and passed the CodeQL checks: https://github.com/alphagov/govuk-content-api-docs/pull/203

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -2,6 +2,12 @@ name: CodeQL Analysis
 
 on:
   workflow_call:
+    inputs:
+      languages:
+        # see https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/customizing-your-advanced-setup-for-code-scanning#changing-the-languages-that-are-analyzed
+        description: Languages to analyse. Defaults to languages of found file extensions
+        required: false
+        type: string
 
 jobs:
   analyze:
@@ -34,6 +40,7 @@ jobs:
                 # ensure that developers are not distracted by low severity, none issues.
                 # Explanation of security severity scores: https://docs.github.com/en/code-security/code-scanning/managing-code-scanning-alerts/about-code-scanning-alerts#about-security-severity-levels
                 security-severity: /([7-9]|10)\.(\d)+/
+        languages: ${{ inputs.languages }}
 
     - name: Autobuild
       uses: github/codeql-action/autobuild@014f16e7ab1402f30e7c3329d33797e7948572db # v4.31.3


.github/workflows/codeql-analysis.yml
```diff
@@ -2,6 +2,12 @@ name: CodeQL Analysis
 
 on:
   workflow_call:
+    inputs:
+      languages:
+        # see https://docs.github.com/en/code-security/code-scanning/creating-an-advanced-setup-for-code-scanning/customizing-your-advanced-setup-for-code-scanning#changing-the-languages-that-are-analyzed
+        description: Languages to analyse. Defaults to languages of found file extensions
+        required: false
```
AgaDufrat, line 9: Have you tested if it falls back to the automatic detection correctly? It will [default to an empty string](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onworkflow_callinputs). I'm not sure if `""` is the same as not providing an argument to the `languages` input 🤔

AgaDufrat (commented):
Looks like a sensible workaround. We should also update [the docs](https://docs.publishing.service.gov.uk/manual/codeql.html#troubleshooting) once it's merged.

**Not for this PR (and team)** but I wonder if we could use the language matrix strategy, which appears to the recommended method. We could probably use the [resolve-languages command](https://docs.github.com/en/code-security/codeql-cli/codeql-cli-manual/resolve-languages) to detect “buildable” languages and then pass them into the matrix, which would avoid having hardcoded lists.

.github/workflows/codeql-analysis.yml
yndajas: I tried it with the `languages:` key and no value (with the full job copied into the Content API Docs workflow) and it tried analysing JavaScript again. I didn't try `languages: ""`, but I suspect `languages:` is the same given YAML strings don't _need_ quoting?

AgaDufrat: Cool, thanks. Just wanted to check

yndajas:
> **Not for this PR (and team)** but I wonder if we could use the language matrix strategy, which appears to the recommended method. We could probably use the [resolve-languages command](https://docs.github.com/en/code-security/codeql-cli/codeql-cli-manual/resolve-languages) to detect “buildable” languages and then pass them into the matrix, which would avoid having hardcoded lists.

Interesting! I did see the matrix strategy but the examples weren't entirely clear to me. A better form of automatic detection would be good though

.github/workflows/codeql-analysis.yml
yndajas: Just confirmed the following on [this PR](https://github.com/alphagov/govuk-content-api-docs/pull/205):
- no `with: ...` = all three
- `with: ... languages:` = all three
- `with: ... languages: actions,ruby,javascript-typescript` = all three
- `with: ... languages: actions,ruby` = no JavaScript

PR #3218, opened 2025-11-19
Add script to create snapshots

jfharden:
This adds a bash script which, using GNU parallel, will create snapshots, and then snapshot copies in parallel. Each parallel process does the complete process for a single db instance.

If any step of the process has already been completed the script will resume from where it was, this includes if snapshots are currently being created, it will resume waiting.

I moved the lists of instances to operate on into a separate file and sourced it in the reboot script, and now in this script, this ensures we're always operating on the same list of instances.

You'll notice though right now the script is using 3 instances in our test account which have my name, I created these especially for testing this.

It's also using a KMS key I created just for this test. After merge I'll do a follow on very small PR to change to the correct env and list of instances for the next maintenance.

I've used ANSI escape sequences to colour the most important bits of output.

I've deleted the snapshots I took testing this in the test account. You can try it out in test.

### DIFF
--- a/terraform/deployments/rds/create-encrypted-snapshots.sh
+++ b/terraform/deployments/rds/create-encrypted-snapshots.sh
@@ -10,7 +10,7 @@ fi
 # shellcheck disable=SC1091
 source ./db-maintenance-lists.sh
 
-GOVUK_ENVIRONMENT="test"
+GOVUK_ENVIRONMENT="integration"
 
 DBS=("${LITTLE_7_DBS[@]}")
 export KMS_KEY="alias/govuk/rds"


terraform/deployments/rds/create-encrypted-snapshots.sh
```diff
@@ -0,0 +1,297 @@
...
+# shellcheck disable=SC1091
+source ./db-maintenance-lists.sh
+
+GOVUK_ENVIRONMENT="test"
+
+DBS=(
+  "jfharden-test-1"
+  "jfharden-test-2"
+  "jfharden-test-3"
+)
```
jasonBirchall: Will these be removed later?

terraform/deployments/rds/create-encrypted-snapshots.sh
```diff
@@ -0,0 +1,297 @@
...
+
+export KMS_KEY="alias/govuk/jfharden/rds"
```
jasonBirchall: same here

terraform/deployments/rds/create-encrypted-snapshots.sh
```diff
@@ -0,0 +1,297 @@
...
+    "430354129336")
+      echo "test"
+      return;;
+    *)
+      return 1
+  esac
+}
+
+if [[ -z "${AWS_ACCESS_KEY_ID:-}" ]]; then
+  echo "Error: You execute this script with AWS credentials in your env"
```
jasonBirchall: Missing the "must", e.g. 
"You MUST execute this script"

terraform/deployments/rds/create-encrypted-snapshots.sh
```diff
@@ -0,0 +1,297 @@
...
+  # Args:
+  #   $1: Database identifier
+  #   $1: Snapshot identiifer
+  local DB="$1"
+  local SNAPSHOT_IDENTIFIER="$2"
+
+  local STATE
+  while true; do
+    if ! STATE=$(aws rds describe-db-snapshots --db-snapshot-identifier "$SNAPSHOT_IDENTIFIER" --query "DBSnapshots[0].Status" --output text 2>&1); then
+      >&2 echo -e "DB: Unknown error describing RDS snapshot, will retry:\n$STATE"
```
jasonBirchall: should this be `$DB...`?

terraform/deployments/rds/create-encrypted-snapshots.sh
```diff
@@ -0,0 +1,297 @@
...
+  if snapshot_already_exists "$DB" "$ENCRYPTED_SNAPSHOT_NAME"; then
+    echo "$DB: snapshot copy $ENCRYPTED_SNAPSHOT_NAME already exists..."
+  else
+    echo "$DB: snapshot copy doesn't exist, creating"
+    if ! copy_to_encrypted_snapshot "$DB" "$UNENCRYPTED_SNAPSHOT_NAME" "$ENCRYPTED_SNAPSHOT_NAME"; then
+      >&2 echo "$DB: FAILED  to create encrypted snapshot copy!"
+      echo -e "$(colour_red "$DB: FAILED  to create encrypted snapshot copy!")"
+      return 1
+    fi
+  fi
```
jasonBirchall: [nitpick] There are trailing spaces on 262

terraform/deployments/rds/create-encrypted-snapshots.sh
```diff
@@ -0,0 +1,297 @@
...
+
+confirm_snapshots
+
+FAILURE_LOG=$(mktemp -t create-snapshot-error-log)
+FULL_LOG_FILE=$(mktemp -t create-snapshot-log)
+echo "STDERR log file, tail this if you want to see errors and also info output showing the current snapshot status while waiting for them to complete: $FAILURE_LOG"
+
+if ! parallel --line-buffer perform_snapshot {} ::: "${DBS[@]}" 2>> "$FAILURE_LOG" | tee -a "$FULL_LOG_FILE"; then
+  echo "-----------------------------------------------------------------------------------------------"
+  echo "ERRORS OCCURED: Some errors occured, see failure log file: $FAILURE_LOG"
```
jasonBirchall: [nitpick] "ERRORS OCCURED" should be "ERRORS OCCURRED"

jasonBirchall (approved):

terraform/deployments/rds/create-encrypted-snapshots.sh
jfharden: Yeah, I mentioned these in the description :)

terraform/deployments/rds/create-encrypted-snapshots.sh
jfharden: Doh! In vim I have spellcheck running, apparently not in helix :)

PR #3249, opened 2025-11-25
Disable the blue Elasticsearch cluster in Integration

koetsier:
A new green cluster has been instantiated in Integration and now serves all traffic. The blue cluster is no longer required and can be safely disabled until it is needed again for future upgrades.

This change introduces a toggle variable to enable or disable the blue Elasticsearch cluster, and sets it to disabled for the Integration environment only.

### DIFF
--- a/terraform/deployments/elasticsearch/iam.tf
+++ b/terraform/deployments/elasticsearch/iam.tf
@@ -56,8 +56,12 @@ data "aws_iam_policy_document" "can_configure_es_snapshots" {
     actions   = ["iam:PassRole"]
     resources = [aws_iam_role.manual_snapshot_role.arn]
   }
-  statement {
-    actions   = ["es:ESHttpPut"]
-    resources = ["${aws_elasticsearch_domain.opensearch.arn}/*"]
+
+  dynamic "statement" {
+    for_each = length(aws_elasticsearch_domain.opensearch) == 1 ? [aws_elasticsearch_domain.opensearch[0]] : []
+    content {
+      actions   = ["es:ESHttpPut"]
+      resources = ["${statement.value.arn}/*"]
+    }
   }
 }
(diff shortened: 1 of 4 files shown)


dj-maisy:
@koetsier If you re-run the checks (do a blank force-push or something) it should pass now. 🤞

terraform/deployments/elasticsearch/iam.tf
```diff
@@ -58,6 +58,6 @@ data "aws_iam_policy_document" "can_configure_es_snapshots" {
   }
   statement {
     actions   = ["es:ESHttpPut"]
-    resources = ["${aws_elasticsearch_domain.opensearch.arn}/*"]
+    resources = [for d in aws_elasticsearch_domain.opensearch : "${d.arn}/*"]
```
jfharden: Sadly I'm 99% certain this will fail on apply, terraform doesn't validate it, but IAM policy statements which don't reference any resources aren't valid and when the count is 0 that will make this an empty array.

You'll have to use a [dynamic block](https://developer.hashicorp.com/terraform/language/expressions/dynamic-blocks) for this statement so the whole statement doesn't get created

koetsier: Thanks @jfharden. I changed it to a dynamic block

terraform/deployments/elasticsearch/iam.tf
```diff
@@ -56,8 +56,12 @@ data "aws_iam_policy_document" "can_configure_es_snapshots" {
...
   }
-  statement {
-    actions   = ["es:ESHttpPut"]
-    resources = ["${aws_elasticsearch_domain.opensearch.arn}/*"]
+
+  dynamic "statement" {
+    for_each = length(aws_elasticsearch_domain.opensearch) == 1 ? [aws_elasticsearch_domain.opensearch[0]] : []
+    content {
+      actions   = ["es:ESHttpPut"]
+      resources = ["${aws_elasticsearch_domain.opensearch[0].arn}/*"]
```
AP-Hunt: The value in the resources bit should be `statement.value.arn`

AP-Hunt (approved):

PR #3264, opened 2025-11-27
Allow apply immediately to be set per instance

jfharden:
Allow apply immediately to be set per instance and continue to default to false in prod

I have put this in it's own PR so we can see nice clean no change diffs before I actually override it for some RDS instances in the lower envs.


AP-Hunt (approved):
I don't love string comparisons with the environment var, but in the interests if expedience I'll approve it

jfharden:
> environment

I agree completely, I wanted the minimal chance of making a mistake at this point so kept the old comparison.

In the aftermath of all this work I'm going to turn the databases input into an object with defined fields and defaults and get rid of a lot of this kind of string comparison checking, and also all these uses of try which damage my soul a little each time I add a new one 😆

PR #3265, opened 2025-11-28
Add `fork` option to GitHub repository

nimalank7:
Description:
- Add fork configuration to GitHub repository
- Add a precondition check that fails to create the forked resource if `source_owner` and `source_repo` aren't present as these are required
- Set `terraform-govuk-tfe-workspacer`, `markdown-toolbar-element`, `fastly-exporter` and `github-trello-poster` as a fork to stop GitHub terraform provider deleting and re-creating the repository


jfharden:
Hmm there's some shenanigans going on in the plan, it's going to delete the codecommit repo for terraform-tfe-workspacer and add it for terraform-tfe-workspacer-empty.

nimalank7:
@jfharden yep after this PR is merged we can do some terraform state manipulation then we can run it

PR #3277, opened 2025-12-01
Point release cname to new instance in integration

jfharden:
In preparation for a rollback test


jfharden:
Rolling back instead of continuing

Issue #3233, opened 2025-11-20
Renew www.gov.uk Certificate


dj-maisy:
**User Need**

**As a** Platform Engineer
**I want** to renew the www.gov.uk TLS Certificate
**so that** users do not risk seeing a certificate warning in their browser.

---

### What’s Needed

List anything the solution must do or be (behaviour, performance, security, UX, etc.).

- [ ] Renew the above certificate with GlobalSign.
- [ ] Install the new certificate

---

### Acceptance Criteria

Clear, measurable conditions to verify the story is complete and valuable.

- [ ] The certificate presented by www.gov.uk is no longer due to expire on the 17th of December 2025.
- [ ] The process for renewal of certificates is documented and shared with the team (including any places where the cert is used outside Fastly)

---

### Risks & Mitigation (optional)

- Bad things will happen if this renewal is missed.

dj-maisy:
Actions:

* https://github.com/alphagov/govuk-fastly/pull/269
* https://github.com/alphagov/govuk-fastly/pull/270
* https://github.com/alphagov/govuk-fastly-secrets/pull/72
* https://github.com/alphagov/govuk-fastly-secrets/pull/73

dj-maisy:
Have contacted the relevant people Re:

* gov.uk
* dfid.gov.uk

All other Domains are now validated.

AngelaConway:
Moving to blocked - no response from CDDO - escalating

dj-maisy:
CDDO Domains support apparently do not know who manages the DNS records.

jfharden:
No idea how but both the apex gov.uk and also dfid.gov.uk are now renewed and valid until Dec 28 2026

nicholsj:
In the Fastly dashboard, the validation for gov.uk and dfid.gov.uk is still failing. 

We've spoken to the Domains Team and they are raising a request with Nominet to add the necessary config for `gov.uk`, and will review their process.

nicholsj:
The necessary CNAME record has been added and I've triggered validation (it's still pending)

dj-maisy:
### Update (2025-12-18)

* The necessary gov.uk Certificate has been renewed
* There is a technical issue on the Fastly Dashboard which has been picked up by Fastly Support and will be addressed after the Christmas break, this is preventing the validation process from displaying correctly
* dfid.gov.uk is still being chased to find who is responsible for the DNS entries
* All other SANs have been split into their own certificate subscriptions

[...]

PR #3312, opened 2025-12-04
Use ACME Challenge from Fastly for Publishing

dj-maisy:
## What?
This creates the necessary ACME Challenge CNAME record for `publishing.service.gov.uk` domain for validation by Fastly.

### Related

* https://github.com/alphagov/govuk-infrastructure/issues/3233

### DIFF
--- a/terraform/deployments/root-dns/root_dns_zones.tf
+++ b/terraform/deployments/root-dns/root_dns_zones.tf
@@ -18,10 +18,10 @@ resource "aws_route53_zone" "publishing_subdomain" {
 }
 
 resource "aws_route53_record" "publishing_fastly_acme_challenge" {
-  count = var.govuk_environment == "production" ? 1 : 0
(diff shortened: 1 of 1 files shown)


terraform/deployments/root-dns/remote.tf
```diff
@@ -2,3 +2,11 @@ data "tfe_outputs" "vpc" {
   organization = "govuk"
   workspace    = "vpc-${var.govuk_environment}"
 }
+
+data "tfe_outputs" "fastly_www" {
```
samsimpson1, line 6: Does this create a circular dependency? In a DR scenario, would the fastly root depend on this one to work?

dj-maisy: I don't think so. The only data resource in govuk-fastly that calls anywhere is for logging. So I think this should be fine?

PR #3314, opened 2025-12-04
Update db maintenance scripts for big 20 in production

jfharden:


jfharden:
I already did this in https://github.com/alphagov/govuk-infrastructure/pull/3291/

PR #3320, opened 2025-12-08
Update the databases variable to be a list of objects with defined fields

jfharden:
Having to work on the rds instances when the variables defining them are free form objects of any type is really painful. It also means we need to use lookup and/or try everywhere which introduces many possible typos, and the need to repeatedly specify what their defaults are.

So this PR:

* Changes the databases variable to be a list of objects with defined attributes and specified defaults.
* Updates the rds module to remove all the lookups and trys and use the defaults instead

The ultimate test of this is that the plans should show no changes and no errors


jfharden:
FYI I iteratively got this working and then rebased to squash all the commits

Issue #3358, opened 2025-12-15
Upgrade publishing-api database to Postgres 17


nicholsj:
These are not intended as a complete set of all tasks - those are in the runbooks (linked below). Instead these are the key stages that it's useful to track

- [x] Make a copy of the dev runbook and update with relevant changes

### Integration

[Developer runbook for staging & integration](https://docs.google.com/document/d/1A0zxAvIUgGlvHMlby9K6oBM7ZpYJb8HiATsGW5l4ZVA/edit?usp=sharing) - updated version of the runbook for publishing-api specifically in integration and staging
[SRE runbook for publishing-api staging & integration](https://docs.google.com/document/d/1gClCnuC3yCcutjLwHkhH9_a6SUNPc0IySOvdNRSaCoI/edit?tab=t.0#heading=h.fp6hiaqcco5f) - updated version of the runbook for publishing-api specifically in integration & staging

- [x] [SRE] set up parameter group
- [x] [Dev] Perform blue/green deployment and switchover
- [x] [Dev] Clean up tasks
- [x] [SRE] Terraform state manipulation
- [x] [SRE] Update [sheet](https://docs.google.com/spreadsheets/u/0/d/1dFCIqKstWa2ubNY_GzPq8WRvX_m0Lkc0au8eBhQ8P_k/edit)

### Staging 

[Developer runbook for staging & integration](https://docs.google.com/document/d/1A0zxAvIUgGlvHMlby9K6oBM7ZpYJb8HiATsGW5l4ZVA/edit?usp=sharing) - updated version of the runbook for publishing-api specifically in integration and staging
[SRE runbook for publishing-api staging & integration](https://docs.google.com/document/d/1gClCnuC3yCcutjLwHkhH9_a6SUNPc0IySOvdNRSaCoI/edit?tab=t.0#heading=h.fp6hiaqcco5f) - updated version of the runbook for publishing-api specifically in integration & staging

- [x] [SRE] set up parameter group
- [x] [Dev] Perform blue/green deployment and switchover
- [x] [Dev] Clean up tasks
- [x] [SRE] Terraform state manipulation
- [x] [SRE] Update [sheet](https://docs.google.com/spreadsheets/u/0/d/1dFCIqKstWa2ubNY_GzPq8WRvX_m0Lkc0au8eBhQ8P_k/edit)

### Production

[...]

PR #3359, opened 2025-12-15
Set `publishing-api` param group to 17

nimalank7:
Description:
- Parameter group set to 17 for postgres upgrade
- https://github.com/alphagov/govuk-infrastructure/issues/3358


jfharden:
The file name is still saying pg 14, but it doesn't matter enough to ask for changes

Issue #2842, opened 2025-09-18
🔒 Add aws:SecureTransport enforcement to all S3 buckets


nicholsj:
This issue connects to https://github.com/alphagov/govuk-platform-internal/issues/64

nicholsj:
[Backlog refinement] Agreed to change to Medium urgency

AP-Hunt:
PR merged and deployed to integration. Giving it a couple of days to prove nothing breaks.

PR #3365, opened 2025-12-15
Enforce TLS connections to S3 buckets across the board

AP-Hunt:
We had a small number of buckets whose policies didn't enforce it, and a smaller number which already did. THey all now deny access if not over TLS.


dj-maisy (approved):

nicholsj:
Fixes https://github.com/alphagov/govuk-infrastructure/issues/2842

PR #3378, opened 2025-12-17
chore: 🤖 remove old unused imports

jaskaransarkaria:
NOOP

We have upgraded the kubernetes provider to v3 and that deprecated some k8s resources. After importing these to the new resources we can remove the old import statements


nimalank7 (approved):
LGTM!

PR #3383, opened 2025-12-18
Add GitHub Actions OIDC role for govuk-chat-evaluation Bedrock CI

chaecramb:
Add new `chat-ci` Terraform deployment that creates:

- GitHub Actions OIDC provider in the govuk-test AWS account
- IAM role assumable by `alphagov/govuk-chat-evaluation` workflows
- Bedrock invoke policy scoped to `openai.gpt-oss-120b-1:0` model

This allows govuk-chat-evaluation CI to run `real_bedrock` pytest-marked tests by authenticating via OIDC and invoking the Bedrock model in eu-west-1.

Also adds the `chat-ci` workspace configuration in tfc-configuration.

Trello: https://trello.com/c/hBGL3B9V/2994-can-our-real-openai-auto-eval-tests-be-real-bedrock

### DIFF
--- a/terraform/deployments/chat-evaluation-ci/github_oidc_provider.tf
+++ b/terraform/deployments/chat-evaluation-ci/github_oidc_provider.tf
@@ -1,0 +1,9 @@
+data "tls_certificate" "github_actions" {
+  url = "https://token.actions.githubusercontent.com/.well-known/openid-configuration"
+}
+
+resource "aws_iam_openid_connect_provider" "github_actions" {
+  url             = "https://token.actions.githubusercontent.com"
+  client_id_list  = ["sts.amazonaws.com"]
+  thumbprint_list = [data.tls_certificate.github_actions.certificates[0].sha1_fingerprint]
+}
--- a/terraform/deployments/chat-evaluation-ci/iam_github_actions_bedrock.tf
+++ b/terraform/deployments/chat-evaluation-ci/iam_github_actions_bedrock.tf
@@ -1,0 +1,68 @@
+locals {
+  github_subjects = [
+    "repo:${var.github_repository}:ref:refs/heads/*",
+    "repo:${var.github_repository}:pull_request",
+  ]
+
+  bedrock_model_arns = [for id in var.bedrock_model_ids : "arn:aws:bedrock:${var.aws_region}::foundation-model/${id}"]
+}
+
+data "aws_iam_policy_document" "github_actions_assume_role" {
+  statement {
+    sid     = "GitHubActionsAssumeRoleWithWebIdentity"
+    actions = ["sts:AssumeRoleWithWebIdentity"]
+
+    principals {
+      type        = "Federated"
+      identifiers = [aws_iam_openid_connect_provider.github_actions.arn]
+    }
+
+    condition {
+      test     = "StringEquals"
+      variable = "token.actions.githubusercontent.com:aud"
+      values   = ["sts.amazonaws.com"]
+    }
+
+    condition {
+      test     = "StringLike"
+      variable = "token.actions.githubusercontent.com:sub"
+      values   = local.github_subjects
+    }
+  }
+}
+
+resource "aws_iam_role" "github_actions_bedrock_ci" {
+  name                 = var.role_name
+  assume_role_policy   = data.aws_iam_policy_document.github_actions_assume_role.json
+  max_session_duration = 3600
+}
+
+data "aws_iam_policy_document" "bedrock_invoke_policy" {
+  statement {
+    sid    = "InvokeBedrock"
+    effect = "Allow"
+    actions = [
+      "bedrock:InvokeModel",
+      "bedrock:InvokeModelWithResponseStream",
+    ]
+    resources = local.bedrock_model_arns
+  }
+
+  statement {
+    sid       = "ListFoundationModels"
+    effect    = "Allow"
+    actions   = ["bedrock:ListFoundationModels"]
+    resources = ["*"]
+  }
+}
+
+resource "aws_iam_policy" "bedrock_invoke_policy" {
+  name        = "${var.role_name}-policy"
+  description = "Allow GitHub Actions CI to invoke specified Bedrock models in ${var.aws_region}"
+  policy      = data.aws_iam_policy_document.bedrock_invoke_policy.json
+}
+
+resource "aws_iam_role_policy_attachment" "attach_bedrock_invoke" {
+  role       = aws_iam_role.github_actions_bedrock_ci.name
+  policy_arn = aws_iam_policy.bedrock_invoke_policy.arn
+}
--- a/terraform/deployments/chat-evaluation-ci/main.tf
+++ b/terraform/deployments/chat-evaluation-ci/main.tf
@@ -1,0 +1,37 @@
+terraform {
+  cloud {
+    organization = "govuk"
+    workspaces {
+      tags = ["chat-evaluation-ci", "test", "aws"]
+    }
+  }
+
+  required_version = "~> 1.10"
+
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "~> 6.0"
+    }
+    tls = {
+      source  = "hashicorp/tls"
+      version = "~> 4.0"
+    }
+  }
+}
+
+provider "aws" {
+  region = var.aws_region
+
+  default_tags {
+    tags = {
+      product              = "govuk"
+      system               = "govuk-chat-evaluation"
+      environment          = var.govuk_environment
+      owner                = "govuk-platform-engineering@digital.cabinet-office.gov.uk"
+      repository           = "govuk-infrastructure"
+      terraform-deployment = basename(abspath(path.root))
+      Service              = "govuk-chat-evaluation-ci"
+    }
+  }
+}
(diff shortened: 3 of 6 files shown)


terraform/deployments/tfc-configuration/chat.tf
```diff
@@ -93,3 +93,33 @@ module "chat-production" {
     module.variable-set-chat-production.id
   ]
 }
+
+module "chat-ci" {
```
kevindew: I think this should be in a separate, terraform/deployments/tfc-configuration/chat-ci.tf, file where this module is named chat-ci-test.

My logic for this is that we have a new deployment chat-ci, so separate to the other deployments in this file and that these ones have a suffix of the environment but this does not.

terraform/deployments/tfc-configuration/chat.tf
```diff
@@ -93,3 +93,33 @@ module "chat-production" {
...
+  source = "github.com/alphagov/terraform-govuk-tfe-workspacer"
+
+  organization        = var.organization
+  workspace_name      = "chat-ci"
```
kevindew:
```suggestion
  workspace_name      = "chat-ci-test"
```

terraform/deployments/tfc-configuration/chat.tf
```diff
@@ -93,3 +93,33 @@ module "chat-production" {
...
+  workspace_desc      = "Creates the GitHub Actions OIDC role in govuk-test for govuk-chat-evaluation Bedrock CI tests"
+  workspace_tags      = ["chat:ci", "test", "aws"]
```
kevindew:
```suggestion
  workspace_tags      = ["test", "chat-ci", "aws"]
```

terraform/deployments/tfc-configuration/chat.tf
```diff
@@ -93,3 +93,33 @@ module "chat-production" {
     module.variable-set-chat-production.id
   ]
 }
+
+module "chat-ci" {
+  source = "github.com/alphagov/terraform-govuk-tfe-workspacer"
+
+  organization        = var.organization
+  workspace_name      = "chat-ci"
+  workspace_desc      = "Creates the GitHub Actions OIDC role in govuk-test for govuk-chat-evaluation Bedrock CI tests"
```
kevindew:
```suggestion
  workspace_desc      = "This module manages resources needed to operate the CI of GOV.UK Chat"
```

just a bit more scope for future

terraform/deployments/tfc-configuration/chat.tf
```diff
@@ -93,3 +93,33 @@ module "chat-production" {
...
+  project_name = "govuk-infrastructure"
+
+  vcs_repo = {
+    identifier     = "alphagov/govuk-infrastructure"
+    branch         = "main"
+    oauth_token_id = data.tfe_oauth_client.github.oauth_token_id
+  }
+
+  team_access = {
+    "GOV.UK Non-Production (r/o)" = "write"
```
kevindew: I don't know quite what this time access means so I could be wrong here (and maybe something to check), however my expectation is that "GOV.UK Production" is a more privileged role than "GOV.UK Non-Production" and thus any application of the latter should be accompanied by the former.

I could be very wrong though.

terraform/deployments/tfc-configuration/chat.tf
```diff
@@ -93,3 +93,33 @@ module "chat-production" {
...
+  }
+
+  variable_set_ids = [
+    local.aws_credentials["test"],
+  ]
```
kevindew: I assume we don't need things like:

```
    module.variable-set-common.id,
    module.variable-set-test.id,
```

Are you confident?

Prior to this review I had no idea what these are so not sure myself on what's needed but looking at [variables-test.tf](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/tfc-configuration/variables-test.tf) there's perhaps a common-ephemeral, but I'm not sure whether that is meant for our AWS test environment or for devs testing.

terraform/deployments/chat-ci/main.tf
```diff
@@ -0,0 +1,37 @@
+terraform {
+  cloud {
+    organization = "govuk"
+    workspaces {
+      tags = ["chat:ci", "test", "aws"]
```
kevindew: Is there convention for this colon approach to a tag (chat:ci) otherwise would wonder if chat-ci is more consistent for this project

terraform/deployments/chat-evaluation-ci/main.tf
```diff
@@ -0,0 +1,37 @@
...
+  }
+}
+
+provider "aws" {
+  region = var.aws_region
+
+  default_tags {
+    tags = {
+      product              = "govuk"
+      system               = "govuk-chat-evaluation"
```
kevindew, line 29: Although I've suggested going with chat-ci naming previously perhaps we'd be better going with chat-evaluation-ci as it would be consistent with this (and indeed the underlying repo is govuk-chat-evaluation)

terraform/deployments/tfc-configuration/chat.tf
kevindew: As per my later comment, maybe all our chat-ci naming should be chat-evaluation-ci

terraform/deployments/chat-evaluation-ci/github_oidc_provider.tf
kevindew, line 1: Should we have a reference to github in this file name instead? it feels like these resources are github data

terraform/deployments/chat-ci/variables.tf
```diff
@@ -0,0 +1,29 @@
...
+}
+
+variable "role_name" {
+  type        = string
+  description = "Name of IAM role assumed by GitHub Actions to run Bedrock tests."
+  default     = "github_action_govuk_chat_evaluation_bedrock_ci"
+}
+
+variable "bedrock_model_id" {
+  type        = string
```
kevindew: Can this be set as a collection (I guess type array) rather than an individual string. I was just thinking that we don't want it to be too tricky to have multiple models that can be used.

Obviously, we don't need this now so if it's a pain let's not worry.

terraform/deployments/tfc-configuration/chat.tf
chaecramb: I'm don't have much knowledge about this but that make sense. Given this deploys to the test account account, presumably both teams should have access.

And looking at the existing patterns for Chat: https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/tfc-configuration/chat.tf                                                                                                                                         
                                                                                                                                                                                                     
- chat-integration has both                                                                                          
- chat-staging / chat-production only has"GOV.UK Production"/write                                                                                                                               
                                                                                                                                                                                                     
I'd have thought this should match integration.

terraform/deployments/tfc-configuration/chat.tf
chaecramb: As far as I can tell we don't need those variable sets. But being brand new to terraform I'd have to talk to Ian to be confident, he did have a look over this and thought it looked fine though.
                                                                                                                                                                                                     
Looking at the variable sets, they both seem to be about EKS cluster auth, which we don't need here.

terraform/deployments/chat-ci/main.tf
chaecramb: Good point, I've switched it to `-`. That seems to be what's exclusively used here.

terraform/deployments/chat-evaluation-ci/github_oidc_provider.tf
chaecramb: Yes, that seems reasonable. I was following the convention of: https://github.com/alphagov/govuk-infrastructure/blob/ce2ef090fd84cb87e4f349191235a45b4c799381/terraform/deployments/github/aws_oidc_provider.tf

But we also have https://github.com/alphagov/govuk-infrastructure/blob/ce2ef090fd84cb87e4f349191235a45b4c799381/terraform/deployments/govuk-publishing-infrastructure/github_oidc_provider.tf

I'll rename it to match the second instead.

terraform/deployments/chat-ci/variables.tf
chaecramb: Good idea, I've updated it.

chaecramb:
@kevindew Comments have all been addressed

kevindew:
This looks good to me now @chaecramb - great job.

Shall we mark it as ready for review and see if @samsimpson1 is available to review it?

terraform/deployments/chat-evaluation-ci/iam_github_actions_bedrock.tf
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+}
+
+resource "aws_iam_role" "github_actions_bedrock_ci" {
+  name                 = var.role_name
+  assume_role_policy   = data.aws_iam_policy_document.github_actions_assume_role.json
+  max_session_duration = 3600
+
+  depends_on = [aws_iam_openid_connect_provider.github_actions]
```
samsimpson1: Is this necessary? There should be an implicit dependency on this resource from the entry in the assume role policy document

chaecramb: Good catch, thanks. Yes, this depends_on is unnecessary. The assume role policy document references aws_iam_openid_connect_provider.github_actions.arn, so Terraform already has the implicit dependency. I’ll remove it to reduce noise.

PR #3388, opened 2025-12-18
Remove logical replication params from publishing-api in staging and also old imports

jfharden:
Imports are complete and we no longer need logical replication on publishing-api rds instance


nimalank7 (approved):
LGTM!

PR #3459, opened 2026-01-13
Remove deletion protection for old DBs from prod

jfharden:
Remove deletion protection from the big 20 old instances prior to deleting them


jaskaransarkaria (approved):
I can see all 18 `deletion_protection  = false` in the plan, so LGTM

PR #3460, opened 2026-01-13
Destroy old 'big 20' in production

jfharden:
These databases are no longer in use, destroy them in production


jaskaransarkaria (approved):
confirmed 18 changed `destroy_old_instance  = true` in tf plan, so LGTM

PR #3461, opened 2026-01-13
Correct permissiosn of  mirror sync role

AP-Hunt:
The mirror sync role's permissions have been adjusted in two ways

1. Granted "glue:BatchGetTable" on the relevant table ARN
2. Granted "s3:ListBucket" and "s3:GetObject" on the S3 bucket in which the Fastly logs are stored.

These permissions are required when using Athena to query an S3-backed data source. We missed them before because the IAM roles we used to test run the mirror had implicit permission to do these things.

The Fastly lgos bucket ARN is pulled in from its Terraform workspace. It doesn't create a circular dependency.

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/govuk_mirror_sync.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/govuk_mirror_sync.tf
@@ -440,7 +440,7 @@ data "aws_iam_policy_document" "govuk_mirror_sync" {
   }
 
   statement {
-    sid = "AthenaS3Permissions"
+    sid = "AthenaS3PermissionsInResults"
     actions = [
       "s3:AbortMultipartUpload",
       "s3:GetBucketLocation",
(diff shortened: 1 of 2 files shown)


terraform/deployments/govuk-publishing-infrastructure/govuk_mirror_sync.tf
```diff
@@ -459,9 +459,27 @@ data "aws_iam_policy_document" "govuk_mirror_sync" {
...
+      test     = "ForAnyValue:StringEquals"
+      values   = ["athena.amazonaws.com"]
+      variable = "aws:CalledVia"
+    }
+  }
+
   statement {
     sid = "AthenaGluePermissions"
     actions = [
+      "glue:BatchGetTable",
```
jfharden, line 482: According to the [iam policy actions page](https://docs.aws.amazon.com/service-authorization/latest/reference/list_awsglue.html) BatchGetTables isn't a policy action

terraform/deployments/govuk-publishing-infrastructure/govuk_mirror_sync.tf
```diff
@@ -472,6 +490,7 @@ data "aws_iam_policy_document" "govuk_mirror_sync" {
       "arn:aws:glue:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:catalog",
       "arn:aws:glue:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:database/fastly_logs",
       "arn:aws:glue:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:database/fastly_logs/govuk_www",
+      "arn:aws:glue:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:table/fastly_logs/govuk_www",
```
jfharden, line 492: Doh this is my fault, the line above sould be table/ not database/, given `database/{database}/{table}` isn't valid (from line 474) it's probably better to change my incorrect line above

terraform/deployments/govuk-publishing-infrastructure/govuk_mirror_sync.tf
AP-Hunt: It's an error in the docs. CloudTrail shows it happening during an Athena call. 
image

jfharden: I've submitted an "Incomplete content" feedback for the policy actions page saying it's missing from the page :(

PR #3392, opened 2025-12-19
Vale prose linting

AP-Hunt:
[Vale](https://vale.sh/) is a linting tool for written language, similar to a linter for a programming language. There are three goals in introducing it

1. Make our writing more consistent
2. Reduce the amount of time it takes to review prose
3. Improve the quality of our prose reviews

~~This is a proof of concept implementation, whose goal is to give a flavour of the complexity involved in doing a full implementation.~~

This is a full implementation of the idea, with all of our written docs in the repository linted and corrected.

PR #3472, opened 2026-01-13
Introduce Vale configuration

AP-Hunt:
This PR is a cut down and amalgamated version of https://github.com/alphagov/govuk-infrastructure/pull/3392 

It was created by running `git format-patch` to generate patches from every commit on the branch which touched certain files, then re-applying them on a new branch with `git apply`

For example

```
git checkout vale_prose_linting
git format-patch -o ../patches/vocabulary/ main..vale_prose_linting -- .vale/styles/config/vocabularies .gitignore

git checkout vale_configuration
git apply ../patches/vocabulary/*

git add .vale/
git commit
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,93 @@ See the [`docs/` directory](docs/).
 
 There are also docs in [the `terraform/docs/` directory](terraform/docs/) and inline in READMEs in some directories.
 
+## Documentation linting
+
+This repository uses [Vale](https://vale.sh/) to lint our written English prose. Vale applies a variety of rules and 
+heuristics, such as avoiding overly long sentences and not using the passive voice, as well as doing spelling and grammar
+checks.
+
+To run the documentation linting locally, run 
+
+```shell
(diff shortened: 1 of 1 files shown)


.github/workflows/lint_docs.yml
```diff
@@ -0,0 +1,31 @@
...
+  workflow_dispatch: {}
+  pull_request:
+    paths:
+      - "docs/**/*"
+
+env:
+  VALE_VERSION: "3.13.0"
+jobs:
+  lint_docs:
+    runs-on: ubuntu-latest
```
jaskaransarkaria, line 14: worth adding a permissions block to avoid annoying code scanning security issues like https://github.com/alphagov/govuk-helm-charts/security/code-scanning/46

AP-Hunt: Good call. I've done that and squashed it back into the original commit as a fixup.

Issue #3423, opened 2026-01-06
Configure renovate to update EKS addons


jfharden:
**User Need**

**As a** Platform Engineer
**I want** to keep our EKS add-ons up to date in a controlled manner, and at a time of our choosing
**so that** we remain security patched, and also don't have unexpected add-on updates appear when trying to apply other changes

---

### What’s Needed

List anything the solution must do or be (behaviour, performance, security, UX, etc.).

- [ ] Configure renovate to open PRs to update the aws eks addons (See [renovate docs for aws_eks_addon data source](https://docs.renovatebot.com/modules/datasource/aws-eks-addon/))

---

### Acceptance Criteria

Clear, measurable conditions to verify the story is complete and valuable.

- [ ] When new add on versions are available a PR is opened to update the addon

---

### Assumptions (optional)

- renovate runs inside EKS so we can give it the required permissions via [IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html)
- We have already pinned to specific EKS addon versions in https://github.com/alphagov/govuk-infrastructure/pull/3422

jasonBirchall:
[Backlog refinement] since the versions are now pinned, we won't be paying a lot of attention to this. It's important for us to know when a new addon is available. We'll keep the urgency as it is.

jasonBirchall:
[preplanning] @samsimpson1 will give a kick-off for others on Tuesday, 13th Jan.

PR #3473, opened 2026-01-13
ci: 🎡 update renovate.json to watch eks addons

jaskaransarkaria:
This is the 3/3 prs to enable renovate to watch eks addons for updates:

- ~~[create the new renovate irsa](https://github.com/alphagov/govuk-infrastructure/pull/3471)~~
- ~~[pass the new service account with correct aws permissions to the renovate cron job](https://github.com/alphagov/govuk-helm-charts/pull/3983)~~
- configure the renovate.json to check for any updated eks addons and add gha to ensure that the hardcoded k8s version in `renovate.json` is up to date

config validated with the following command:

```
LOG_LEVEL=debug npx --yes --package renovate -- renovate --token github_pat_XXXXXX --platform local --repository-cache reset
```

I haven't managed to eradicate one bit of hardcoding (the kubernetes version in renovate.json), I think, in the grand scheme of things, it's ok as we don't change k8s versions regularly. But to make sure it's not completely forgotten about, there is a gha in the pr that will cross-check the value and fail unless it's correct and in a subsequent pr I'll add it to the upgrade docs too.

related: https://github.com/alphagov/govuk-infrastructure/issues/3423

### DIFF
--- a/.github/workflows/validate-renovate-k8s-version.yml
+++ b/.github/workflows/validate-renovate-k8s-version.yml
@@ -1,7 +1,7 @@
 name: Validate renovate.json is using the correct k8s version
 on:
   push:
-    paths: ['terraform/deployments/tfc-configuration/variables-production.tf']
+    paths: ['terraform/deployments/tfc-configuration/variables-production.tf', 'renovate.json']
 jobs:
   k8s_version_validate:
     name: Validate renovate.json is using the correct k8s version


.github/workflows/validate-renovate-k8s-version.yml
```diff
@@ -0,0 +1,26 @@
+name: Validate renovate.json is using the correct k8s version
+on:
+  push:
+    paths: ['terraform/deployments/tfc-configuration/variables-production.tf']
```
jfharden: Probably want to include the renovate.json as well

jaskaransarkaria: good shout, added via a fixup 👍🏽

jfharden (approved):

PR #3481, opened 2026-01-14
Specify snapshot identifiers

jfharden:
Now we aren't just launching from a known snapshot, use snapshot_identifier only and manually specify all the snapshot identifiers

### DIFF
--- a/terraform/deployments/rds/rds.tf
+++ b/terraform/deployments/rds/rds.tf
@@ -48,7 +48,7 @@ resource "aws_db_instance" "normalised_instance" {
 
   // This is purposefully not referencing the resource so that we can create snapshots outside of terraform and use them to launch
   // this instance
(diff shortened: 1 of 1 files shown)


terraform/deployments/rds/rds.tf
```diff
@@ -48,7 +48,7 @@ resource "aws_db_instance" "normalised_instance" {
 
   // This is purposefully not referencing the resource so that we can create snapshots outside of terraform and use them to launch
   // this instance
-  snapshot_identifier = each.value.launch_new_db_from_snapshot ? "${local.identifier_prefix}${each.value.name}-${each.value.engine}-post-encryption" : null
+  snapshot_identifier = each.value.snapshot_identifier != null ? each.value.snapshot_identifier : null
```
jfharden: Why am I doing this, if it's null, null is correct, I'll get rid of the ternary

jfharden: Boom, gone!

PR #3490, opened 2026-01-14
Set all rds instance engine versions to Major for PG and Major.Minor for MySQL

jfharden:
We want auto minor version updates to happen and our terraform not try and change them back, so set the engine_version to major only for postgres or major.minor only for MySQL.

Where the version specfied is not the current default (which is what terraform will try to set it to) I've set apply immediately to false so it doesn't interrupt the working day.

### DIFF
--- a/terraform/deployments/tfc-configuration/variables-integration.tf
+++ b/terraform/deployments/tfc-configuration/variables-integration.tf
@@ -132,8 +132,9 @@ module "variable-set-rds-integration" {
 
     databases = {
       account_api = {
-        engine         = "postgres"
-        engine_version = "14.18"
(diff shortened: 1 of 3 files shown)


jfharden:
> There's a few dbs across the envs that don't have `apply_immediately = false`. I guess they are either defaulting to false somewhere or noops?

The default is true in non-prod envs and false in prod :) 

If the version specified was already the default RDS version for the engine version then the change is a no-op, that's 14.19, 16.10, and 17.6 for postgres, and 8.0.43 for MySQL LTS version 8.0

jaskaransarkaria (approved):

PR #3498, opened 2026-01-15
Add normalised cnames

jfharden:
We want to change the cnames for places-manager (previously called imminence), content-data-api (cname is blue-content-data-api), and the publishing-api replica.

This PR

* Creates a new normalised_instance_cname resource
* Moves all apart from imminence and content-data-api into the new normalised resource
* Adds a filter on the old resource to only keep the records for imminence and content-data-api
* Creates a new normalised_replica_cname resource

Note:

I can't do a moved block for fact-check-manager since it only exists in integration for now, for this single record only I'll manually run:

```
terraform workspace select rds-integration
terraform state mv 'aws_route53_record.instance_cname["fact_check_manager"]' 'ws_route53_record.normalised_instance_cname["fact_check_manager"]'
```

Once this PR is applied we will have new records (in normalised_instance_cname and normalised_replica_cname) and I can change the URL env vars to use the new cnames, before finally deleting the moved blocks, the instance_cname and the replica_cname resources

### DIFF
--- a/terraform/deployments/rds/rds.tf
+++ b/terraform/deployments/rds/rds.tf
@@ -287,7 +287,7 @@ resource "aws_route53_record" "normalised_instance_cname" {
   zone_id = data.tfe_outputs.root_dns.nonsensitive_values.internal_root_zone_id
 
   // Right now the names are stuck as the old names. Hopefuilly we can change this soon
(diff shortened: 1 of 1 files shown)


jfharden:
Have run the state mv

Issue #3441, opened 2026-01-08
Setup DGU Tenancy Access for Kubernetes


dj-maisy:
**User Need**

**As a** DGU Engineer
**I want** to be able to access and interact with workloads in the DGU Namespace on GOV.UK's EKS Clusters
**so that** we can develop, diagnose, maintain and troubleshoot our application stack

---

### What’s Needed

List anything the solution must do or be (behaviour, performance, security, UX, etc.).

- [x] Create a `dguengineer` IAM Role for AWS in `govuk-user-reviewer` (use `licensinguser` as a template)
- [x] Assign the National Data Library (DGU) Engineers via the `external_users.yml` file
- [x] Refactor the EKS Access Entries (if possible) configuration to make it easier to manage roles and tenants, as the existing file is starting to sprawl
- [x] Add the necessary EKS Access Entries resources to `cluster-access/eks_access.tf` in [alphagov/govuk-infrastructure] for `dguengineer` users that grants them access to the `datagovuk` Namespace only
- [ ] Revoke access to the `datagovuk` Namespace for any roles that are not `platformengineer`

---

### Acceptance Criteria

Clear, measurable conditions to verify the story is complete and valuable.

- [ ] DGU Engineers have tested and can confirm they can access Kubernetes resources via `kubectl`.

---

### Not in Scope

* Granting or Managing Access to shared Platform Resources (e.g. Grafana, Argo, Dex, etc.) - this will be addressed in a separate ticket (#3442)

### Assumption

- We are assuming DGU developers already have AWS gds-users access.

jfharden:
@jasonBirchall @nicholsj This is REALLY interesting for our use case in the future world of log in via SSO & EntraID. We definitely need more than just "developer" since we are now in the business of running a multi-tenant platform

samsimpson1:
* A new IAM role, `dguengineer` has been created. It has read-only access to most AWS resources and read/write access to the `datagovuk` K8s namespace.
* Two new role types have been added to govuk-user-reviewer, `dguadmin` and `dguengineer`. `dguadmin` grants access to the `dguengineer` and `fulladmin` roles. `dguengineer` only allows access to the role named the same.
* The GDS CLI has been updated to support this new IAM role
* Ken has been informed, and has tested the new role. He has confirmed it is working as expected

PR #3500, opened 2026-01-16
Modularise access entries in cluster-access and add new dguengineer access entry

samsimpson1:
This adds a modularised copy of the TF resources we use to configure EKS cluster access for all other roles.

The new dguengineer role grants read/write access to only the datagovuk namespace, and nothing else. Similar to how the licensing role is configured.

All other roles will be reworked to use the new module once testing of the new dguengineer role is complete.

https://github.com/alphagov/govuk-infrastructure/issues/3441

### DIFF
--- a/terraform/deployments/cluster-access/eks_access.tf
+++ b/terraform/deployments/cluster-access/eks_access.tf
@@ -420,3 +420,33 @@ resource "kubernetes_role_binding_v1" "readonly" {
     api_group = "rbac.authorization.k8s.io"
   }
 }
+
+module "dguengineer" {
+  source = "./modules/access-entry"
+
+  name = "dguengineer"
+
+  cluster_name = local.cluster_name
+
+  access_policy_arn        = "arn:aws:eks::aws:cluster-access-policy/AmazonEKSAdminPolicy"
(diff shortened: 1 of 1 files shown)


terraform/deployments/cluster-access/eks_access.tf
```diff
@@ -420,3 +420,33 @@ resource "kubernetes_role_binding_v1" "readonly" {
...
+
+  name = "dguengineer"
+
+  cluster_name = local.cluster_name
+
+  access_policy_arn        = "arn:aws:eks::aws:cluster-access-policy/AmazonEKSAdminPolicy"
+  access_policy_scope      = "namespace"
+  access_policy_namespaces = ["datagovuk"]
+
+  role_rules = [
```
jaskaransarkaria, line 435: Is it worth splitting out `cluster_role_rules` and `ns_role_rules` they might not always be the same

samsimpson1: Right now we only provide either namespaced or cluster-wide access to a role, not both. This might need to change in the future if we need to start mixing namespaced and cluster-wide permissions but it should be fine for now, and its neat

Issue #2187, opened 2025-05-20
Audit and Upgrade Terraform Versions


dj-maisy:
### User Need

**As a** Platform Engineer
**I want** to identify and upgrade any out-of-date Terraform versions
**so that** we stay on top of technical debt, bugs, security issues and can benefit from new features

---

### Functional Requirements (What):

- [ ] Identify each Terraform root in the GOV.UK managed repositories (see list below) and document their current version
- [ ] Upgrade all of the identified entities to the latest available Terraform version
- [ ] Check that the Terraform plans and applies as expected across all environments
~- [ ] Ensure Renovate is managing Terraform versions for every TFC workspace~ [preplanning 9/1/26] we've decided not to do this part as we would like to wait until post-concourse work.

---

### Acceptance Criteria:

- [ ] All Terraform across the known repos below is up to date.
- [ ] If a way to improve Terraform version management is identified, this should either be implemented, or written into a new User Story and linked back to this ticket.
- [ ] Changes (if any) during delivery are reflected in this story.
- [ ] Impacted users have been notified or communicated with appropriately (e.g., change logs, Slack, documentation updates).

---

### Known Repos with Terraform:

- [ ] [govuk-fastly](https://github.com/alphagov/govuk-fastly) (and related sensitive repo)
- [ ] [govuk-infrastructure](https://github.com/alphagov/govuk-infrastructure) (and related sensitive repo)
- [ ] [govuk-s3-mirror](https://github.com/alphagov/govuk-s3-mirror)
- [ ] [govuk-user-reviewer](https://github.com/alphagov/govuk-user-reviewer)
- [ ] [terraform-govuk-tfe-workspacer](https://github.com/alphagov/terraform-govuk-tfe-workspacer)

**User Communication Plan (if applicable):**

Describe briefly how users will be informed about the change (e.g., Slack message, email, platform announcement).

---

**Notes:**

(Add any additional context, links to diagrams, stakeholder interviews, etc.)

---

nimalank7:
- `tfc-aws-config` Terraform needs to be run
- Renovate config needs to be setup
- `terraform_version` is set on `govuk-infrastructure`
- Some Terraform workspaces were created manually (e.g. `govuk-dns-tf`, `govuk-user-reviewer`, `github`, `sentry`, `govuk-fastly-secrets` and need to be imported in before `terraform_version` can be set on these

nimalank7:
[...]

PR #3538, opened 2026-01-20
Import `govuk-user-management` Terraform Workspace into Terraform

nimalank7:
Description:
- Import `govuk-user-management` workspace into Terraform as its currently clickopsed
- See the Terraform plan in `tfc-configuration` to review
- https://github.com/alphagov/govuk-infrastructure/issues/2187

### DIFF
--- a/terraform/deployments/tfc-configuration/govuk-user-management.tf
+++ b/terraform/deployments/tfc-configuration/govuk-user-management.tf
@@ -1,0 +1,65 @@
+module "govuk-user-management" {
+  source = "github.com/alphagov/terraform-govuk-tfe-workspacer"
+
+  organization        = var.organization
+  workspace_name      = "govuk-user-management"
+  workspace_desc      = "This module manages user access to Fastly, PagerDuty and GitHub Teams"
+  workspace_tags      = ["govuk-user-management"]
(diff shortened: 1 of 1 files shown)


terraform/deployments/tfc-configuration/govuk-user-management.tf
```diff
@@ -0,0 +1,65 @@
...
+  source = "github.com/alphagov/terraform-govuk-tfe-workspacer"
+
+  organization        = var.organization
+  workspace_name      = "govuk-user-management"
+  workspace_desc      = "This module manages user access to Fastly, PagerDuty and GitHub Teams"
+  workspace_tags      = ["govuk-user-management"]
+  assessments_enabled = true
+  terraform_version   = var.terraform_version
+  execution_mode      = "remote"
+  working_directory   = "/terraform/saas/"
```
jaskaransarkaria: In the plan this dir used to be `./terraform/saas`, will that break anything given that has changed to `/terraform/saas` ?

nimalank7: Nah don't think so but I'll change it back to have a no-op.

PR #3541, opened 2026-01-21
fix: 🐛 add before_compute to add vpc-cni before the workers

jaskaransarkaria:
- add `before_compute = true` to `vpc-cni` addon.

Without this flag new node groups cannot join the cluster. The nodes come up and are a healthy state within the Auto-Scaling Group but has the following cni error in the kubelet logs:

`sudo journalctl -u kubelet -f`

```sh
kubelet.go:3126] "Container runtime network not ready" networkReady="NetworkReady=false reason:NetworkPluginNotReady message:Network plugin returns error: cni plugin not initialized"
```

The `before_compute` var instructs the addon to be provisioned before data plane compute is added [docs example](https://github.com/terraform-aws-modules/terraform-aws-eks?tab=readme-ov-file#eks-managed-node-group).

image

This functionality was added in [v19.9](https://github.com/terraform-aws-modules/terraform-aws-eks/releases/tag/v19.9.0). This was released back in 2023, and because we only [recently pinned our addon versions](https://github.com/alphagov/govuk-infrastructure/commit/2e374df73d3dcbeba47faabf48f20857f5d3783f), it's not clear when we actually adopted this version.

## Testing instructions

- spin up an eph cluster from main, you'll eventually timeout and hit a `Create Failed` for your node group.
- then spin up an eph cluster from this branch `eph-jas-210126`, it should succeed.

## Other changes

- update the ephemeral readme steps
- move ephemeral `cluster_services` above `cluster_access`

### DIFF
--- a/terraform/deployments/ephemeral/README.md
+++ b/terraform/deployments/ephemeral/README.md
@@ -28,20 +28,18 @@ If the shutdown script fails or is in a pending state you can still remove the e
 
 1. During the run of the `shutdown.sh` script, click on the terrform cloud link during the running of the shutdown.sh script to monitor the status of the run, it might be in a `pending` state if there are other runs blocking the destroy run so you will need to force and unlock the workspace and discard earlier runs for the shutdown script to continue.
 2. For any remaining ephemeral resources managed by terraform cloud, locate them using the `eph-` search on workspaces and under the `Settings/Destruction and Deletion` section:
-   - manually queue a deletion of your ephemeral cluster infrastructure and monitor it`s progress on the terraform cloud website.
-   - when the infrastructure resources have been deleted then delete the workspace itself.
-
+   a. manually queue a deletion of your ephemeral cluster infrastructure and monitor it`s progress on the terraform cloud website.
+   b. when the infrastructure resources have been deleted then delete the workspace itself.
 3. If there are still some ephemeral resources remaining you will need to use the terraform cli to remove these last resources:
-
-- update your local terraform state
-    `terraform refresh -var "ephemeral_cluster_id=$EPH_CLUSTER_ID"`
-- list the resources
-    `terraform state list`
-- inspect the resources to make sure that you are destroying your ephemeral resources
-    `terrform show <resource name>`
-- if only ephemeral resources remain then run
-    `terraform destroy`
-- NOTE - terraform cloud workspaces may need to be removed manually on the terraform cloud website under the `Settings/Destruction and Deletion` section.
+   a. update your local terraform state
+       `terraform refresh -var "ephemeral_cluster_id=$EPH_CLUSTER_ID"`
+   b. list the resources
+       `terraform state list`
+   c. inspect the resources to make sure that you are destroying your ephemeral resources
+       `terrform show <resource name>`
+   d. if only ephemeral resources remain then run
+       `terraform destroy`
+   e. NOTE - terraform cloud workspaces may need to be removed manually on the terraform cloud website under the `Settings/Destruction and Deletion` section.
 
 ### Alertmanager, Prometheus and Grafana
 


terraform/deployments/ephemeral/ws/workspace.tf
```diff
@@ -11,6 +11,7 @@ module "workspace" {
   trigger_patterns    = ["/terraform/deployments/${var.name}/**/*"]
   global_remote_state = true
   queue_all_runs      = false
+  force_delete        = true
```
jfharden: We shouldn't add force_delete, this will make it trivial to delete the workspaces without destroying the resources they create.

terraform/deployments/ephemeral/README.md
```diff
@@ -2,13 +2,15 @@ This module is used to provision ephemeral EKS clusters via Terraform Cloud.
...
 
-Set an ephemeral cluster ID before doing anything: `export EPH_CLUSTER_ID=eph-aaa111`
-
+1. `export EPH_CLUSTER_ID=eph-aaa111`
+1. ensure you are in the correct dir `terraform/deployments/ephemeral`
 1. Ensure you have logged in to Terraform Cloud via the Terraform CLI (`terraform login`)
-2. Do a `terraform init`
-3. Run an apply with your chosen ephemeral cluster ID (this isn't generated for you)
+1. Do a `terraform init`
+1. `terraform workspace new ${EPH_CLUSTER_ID}`
```
jfharden: You shouldn't need to do this I don't think, 🤔

terraform/deployments/ephemeral/README.md
```diff
@@ -27,18 +29,21 @@ If the shutdown script fails or is in a pending state you can still remove the e
 
 1. During the run of the `shutdown.sh` script, click on the terrform cloud link during the running of the shutdown.sh script to monitor the status of the run, it might be in a `pending` state if there are other runs blocking the destroy run so you will need to force and unlock the workspace and discard earlier runs for the shutdown script to continue.
 2. For any remaining ephemeral resources managed by terraform cloud, locate them using the `eph-` search on workspaces and under the `Settings/Destruction and Deletion` section:
-  - manually queue a deletion of your ephemeral cluster infrastructure and monitor it`s progress on the terraform cloud website.
-  - when the infrastructure resources have been deleted then delete the workspace itself.
+
+- manually queue a deletion of your ephemeral cluster infrastructure and monitor it`s progress on the terraform cloud website.
```
jfharden: These are supposed to be an unordered sublist inside 2. Dedenting them breaks the semantic meaning of the lists, so they should be left indented within 2

terraform/deployments/ephemeral/README.md
jaskaransarkaria: Ran through it this morning following these steps. You need to run that command to create the workspace.

terraform/deployments/ephemeral/tfe.tf
```diff
@@ -69,34 +69,36 @@ module "cluster_infrastructure" {
...
     ship_kubernetes_events_to_logit = false
   }
 
-  depends_on = [module.cluster_infrastructure, module.cluster_services, tfe_project.project]
+  depends_on = [module.cluster_infrastructure, tfe_project.project]
 }
 
-module "cluster_services" {
+
+module "cluster_access" {
```
samsimpson1, line 87: cluster-access has been applied before cluster-services because it makes it impossible to debug cluster-services if it fails to run, since nobody would have cluster access. Are you sure you want to swap these around?

terraform/deployments/ephemeral/README.md
samsimpson1: The ephemeral root runs locally, not on TFC so there should be no workspace to create. The default one is fine (as long as you don't have multiple ephemeral clusters to manage)

terraform/deployments/ephemeral/tfe.tf
jaskaransarkaria: I don't think the ordering of tf blocks affects the order in which something is applied (but the dependency block on `cluster_access` does mean that it's only applied after `cluster_services` atm).

I inferred from the `cluster_access` `depends_on` that there needed to be a hard dependency that `cluster_access` _must_ come after `cluster_services`, so the movement of code is more to highlight that logic. Having played around with `cluster_access` more, I'm not sure that there is any reason for the `depends_on = [*, cluster_services, *]` exists, all we need is the cluster up?

jaskaransarkaria: I tried to apply the `cluster_access` ws before `cluster_services` and it errors waiting for namespaces created in `cluster_services` so there is a hard dependency there. There might be an argument to pull those namespaces out and create them elsewhere but I think htat's a different conversation

PR #3585, opened 2026-01-27
Increase concurrent PRs to 20

nimalank7:
Description:
- Increase the amount of concurrent PRs that Renovate can open to 20
- For `required_version` bump the range
- https://github.com/alphagov/govuk-infrastructure/issues/2187

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -20,12 +20,13 @@
     ],
     "packageRules": [
       {
-        "matchDepTypes": ["helm_release", "provider", "modules", "required_provider", "required_version"],
+        "matchDepTypes": ["required_version"],
         "rangeStrategy": "bump"
       }
     ]
   },
   "prHourlyLimit": 20,
+  "prConcurrentLimit": 20,
   "labels": [
     "dependencies",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -20,12 +20,13 @@
     ],
     "packageRules": [
       {
-        "matchDepTypes": ["helm_release", "provider", "modules", "required_provider", "required_version"],
```
jfharden, line 23: How come you're removing these? They seem like they are important things to keep getting updates for?

nimalank7: Renovate will continue to update on those as those are part of the `terraform` package manager. All I'm doing is restricting the bump to just `required_versions.` The default behaviour for the rest is `auto` - see https://docs.renovatebot.com/configuration-options/#rangestrategy

jfharden: Ah ok cool, I gave the docs a decent read and yeah it's clear this is very granular and does what you describe.

It's possible auto might be a bit bold for us, but we can always deal with that if it is :)

PR #3594, opened 2026-01-28
Added second, private, repo for the Publishing AI Alpha.

minglis:
Some of our MSP code is not yet allowed to be open, so separate repo created for this work to be managed in. to be revisited at the end of the alpha.

PR adds repo. and also maps this repo to the github team for this work.

### DIFF
--- a/terraform/deployments/github/main.tf
+++ b/terraform/deployments/github/main.tf
@@ -123,7 +123,7 @@ data "github_team" "co_platform_engineering" {
 
 moved {
   from = github_team_repository.govuk_ai_accelerator_repos
-  to = github_team_repository.govuk_ai_accelerator_repos["govuk-ai-accelerator"]
+  to   = github_team_repository.govuk_ai_accelerator_repos["govuk-ai-accelerator"]
 }
 
 resource "github_team_repository" "govuk_ai_accelerator_repos" {


terraform/deployments/github/main.tf
```diff
@@ -121,8 +121,15 @@ data "github_team" "co_platform_engineering" {
...
 
+variable "govuk_ai_accelerator_repo_names" {
+  # repos to be used in the GOV.UK Publishing AI alpha
+  type    = list(string)
+  default = ["govuk-ai-accelerator", "govuk-ai-accelerator-tooling"]
+}
+
 resource "github_team_repository" "govuk_ai_accelerator_repos" {
-  repository = "govuk-ai-accelerator"
+  for_each   = toset(var.govuk_ai_accelerator_repo_names)
```
jfharden, line 130: Moving this into a for_each makes terraform want to destroy and the govuk-ai-accelerator repo

Image

To fix this you need to add a moved block (outside the resource) :
```
moved {
    from = github_team_repository.govuk_ai_accelerator_repos
    to  = github_team_repository.govuk_ai_accelerator_repos["govuk-ai-accelerator"]
}
```

terraform/deployments/github/main.tf
```diff
@@ -121,8 +121,15 @@ data "github_team" "co_platform_engineering" {
   slug = "co-platform-engineering"
 }
 
+variable "govuk_ai_accelerator_repo_names" {
```
jfharden: Please could you move this to the variables.tf file

jfharden (changes requested):
Using the request-changes feature because it will try to destroy your repo as is

terraform/deployments/github/main.tf
jfharden: I should mention, once the merged PR has been applied with the moved block in, then it can be deleted (it only needs to apply the moved a single time (and will ignore it for future runs, but it's confusing for future devs if it is still in there))

minglis: Good spot. fixed.

terraform/deployments/github/main.tf
minglis: done

terraform/deployments/github/main.tf
```diff
@@ -82,6 +82,11 @@ locals {
   ]
 }
 
+moved {
+  from = github_team_repository.govuk_ai_accelerator_repos
+  to = github_team_repository.govuk_ai_accelerator_repos["govuk-ai-accelerator"]
```
jfharden: The linter is complaining, sorry about this but you have to align the `=` signs


```suggestion
  to   = github_team_repository.govuk_ai_accelerator_repos["govuk-ai-accelerator"]
```

Worth knowing you can run `terraform fmt` which will fix this stuff automatically (we also have it on a pre-commit hook so you could `pre-commit install` in this repo and it will tell you it's wrong and auto fix it (you'll have to add its fixes and commit again)

PR #3618, opened 2026-02-03
Update repos.yml with new govuk-chat-v1-iterations-2026 private repo

exfalsoquodlibet:
This will be a private python repo used for experimenting and iterating on gov.uk chat.
It has not yet been created.

### DIFF
--- a/terraform/deployments/github/repos.yml
+++ b/terraform/deployments/github/repos.yml
@@ -353,6 +353,14 @@ repos:
         - Format
         - Type checks
 
+  govuk-chat-v1-iterations-2026:
+    visibility: private
+    required_status_checks:
+      # standard security checks are disabled as not configured for a private repo
+      additional_contexts:
+        - Lint
+        - Format
+
   govuk-chat-gradio-user-research:
     visibility: private
     archived: true


terraform/deployments/github/repos.yml
```diff
@@ -353,6 +353,16 @@ repos:
...
 
+  govuk-chat-v1-iterations-2026:
+    visibility: private
+    required_status_checks:
+      # standard security checks are disabled as not configured for a private repo
+      additional_contexts:
+        - Tests
+        - Lint
+        - Format
+        - Type checks
```
kevindew:
```suggestion
```

I think you want to remove the required_status_checks altogether as I don't believe you want any GitHub workflows to be required.

Only thing I'm not 100% sure on is whether we should just remove it or set it as an empty object (`required_status_checks: {}`) but as it looks like there are [others with it removed](https://github.com/alphagov/govuk-infrastructure/blob/9f73d754dc057fc3a461999be49d45222bb88040/terraform/deployments/github/repos.yml#L462-L463) I think that is the right route.

exfalsoquodlibet: Thanks Kev.

I followed the guidance for private beta:
> Note that private repositories can't use the predefined standard_govuk_rails_checks or standard_security_checks. In such cases, include a comment: # standard security checks are disabled as not configured for a private repo. **You should define equivalent jobs in your repository’s CI workflow and reference them in the additional_contexts section of the required status checks**

And it looks like we did the same for other private repos: https://github.com/alphagov/govuk-infrastructure/blob/0a8387d929fd1850f84954ca2af475d3a1c90924/terraform/deployments/github/repos.yml#L345-L354

But very happy to remove it if that's a cleaner approach.

kevindew: So adding these is only worthwhile if you expect to be adding those tasks as GitHub actions workflows. 

In case you're not aware, they essentially define the list of checks that GitHub will require before a pull request is merged. So for the govuk-chat-gradio-prototype that list effectively requires these workflows to succeed before a PR can be merged: [Tests](https://github.com/alphagov/govuk-chat-gradio-prototype/blob/f01392f1f5cef9ad7b60030bc6679b50a0d486a4/.github/workflows/ci.yaml#L13-L28), [Lint](https://github.com/alphagov/govuk-chat-gradio-prototype/blob/f01392f1f5cef9ad7b60030bc6679b50a0d486a4/.github/workflows/ci.yaml#L30-L45), [Format](https://github.com/alphagov/govuk-chat-gradio-prototype/blob/f01392f1f5cef9ad7b60030bc6679b50a0d486a4/.github/workflows/ci.yaml#L47-L62), [Type checks](https://github.com/alphagov/govuk-chat-gradio-prototype/blob/f01392f1f5cef9ad7b60030bc6679b50a0d486a4/.github/workflows/ci.yaml#L64-L79)

My expectation is that for a workspace style repo where people are exploring prototype ideas you'd not be having a CI layer - but if you are that is great. I could imagine tests and type checks being a pain, but lint or format checks could likely be good for consistent code.

exfalsoquodlibet: Thanks Kev! 
> they essentially define the list of checks that GitHub will require before a pull request is merged

yes, that was clear to me and I thought it was a good idea since to have checks for these. 

I take your point though that it may be a pain for tests and type checks given the purpose of the repo. I have removed these and we can always add them later.

kevindew: Sounds good. 

If you do expect to write tests consistently for the work then it's great to have the check - my expectation was that we've not done that before though with experimental work like this?

exfalsoquodlibet: > my expectation was that we've not done that before though with experimental work like this?

yes, we have been very inconsistent, and it is something I'd like to encourage us to do more, but - granted - it "slows down" the experimentation process.

PR #3627, opened 2026-02-04
chore: 🤖 make services dependent on access for better debugging

jaskaransarkaria:
# Improve ephemeral cluster creation flow

```
tl;dr remove namespace creation from cluster-services and import them into cluster-access
```

- when creating ephemeral clusters, `cluster-services` fails frequently, and because `cluster-access` is dependent on services, when it fails we have no access to the cluster to debug.
  - pull out the dependent resources (namespaces) and invert the dependency

# Cluster access module fix
  
`cluster-access` errored when trying to for_each over a data block. This error has shown up since the terraform version upgrade work.

image

---

cluster-services is a NOOP
cluster-access is effectively a NOOP it's showing that it will add some labels to the namespaces that already exist on the actual namespaces

### DIFF
--- a/terraform/deployments/cluster-access/eks_access.tf
+++ b/terraform/deployments/cluster-access/eks_access.tf
@@ -9,6 +9,7 @@ module "platformengineer" {
 
   cluster_name = local.cluster_name
 
+  aws_iam_role_arns   = data.aws_iam_roles.platformengineer.arns
   access_policy_arn   = "arn:aws:eks::aws:cluster-access-policy/AmazonEKSClusterAdminPolicy"
   access_policy_scope = "cluster"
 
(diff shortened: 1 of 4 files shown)


terraform/deployments/cluster-access/namespaces.tf
```diff
@@ -0,0 +1,66 @@
...
+  id = "apps"
+}
+
+resource "kubernetes_namespace_v1" "apps" {
+  metadata {
+    name = var.apps_namespace
+    annotations = {
+      "argocd.argoproj.io/sync-options" = "ServerSideApply=true"
+    }
+    labels = {
```
jfharden, line 12: Interestingly some of these labels show as going to be added 🤔 But it's clearly the exact same resource you have removed and imported 🤔

jfharden (approved):

PR #3628, opened 2026-02-04
IA-2490: create GCP project ga4-aggregate-analytics and TFC workspace

aaronfowles:
This PR:

- create a TFC workspace to manage deployments for a single environment/project
- stub out minimal deployment for `gcp-ga4-aggregate-analytics` project creation with project-level permission groups specified.

This deployment depends on existing [gcp-credentials](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/tfc-configuration/remote.tf#L20). I don't know if that service account has permissions to create projects having not done this before.


nimalank7 (approved):
LGTM!

PR #3642, opened 2026-02-05
Adding a second private repo to store thoughtworks owned code

minglis:
As part of the AI accelerator work we will be working with some thoughtworks code that we don't want to be in a public repo at this stage. This PR creates a repo to have a private fork.

### DIFF
--- a/terraform/deployments/github/repos.yml
+++ b/terraform/deployments/github/repos.yml
@@ -1041,3 +1041,9 @@ repos:
     # standard security checks are disabled as not configured for a private repo
     can_be_deployed: false
     visibility: private
+
+  govuk-ai-accelerator-tw-accelerator:
+    # standard security checks are disabled as not configured for a private repo
+    can_be_deployed: false
+    visibility: private
+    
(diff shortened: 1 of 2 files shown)


terraform/deployments/github/repos.yml
```diff
@@ -1041,3 +1041,9 @@ repos:
     # standard security checks are disabled as not configured for a private repo
     can_be_deployed: false
     visibility: private
+
+  govuk-ai-accelerator-tw-accelerator:
+    # standard security checks are disabled as not configured for a private repo
+    can_be_deployed: false
+    visibility: private
+
```
nimalank7, line 1049: If it's a fork there is some additional config that needs to be added - see https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/github/README.md?plain=1#L96

nimalank7 (changes requested):

terraform/deployments/github/repos.yml
minglis: thanks - in this case it just needs t be an empty repo, the origin code isn't on Github. It's coming from within the owners own code management system.

PR #3646, opened 2026-02-05
Add s3 buckets for ai accelarator

minglis:
PR to add a new bucket for the AI Accelerator work in GOV.UK Publishing.

Should be encrypted, versioned and not public.

This configuration is based on the work by GOV.UK Forms: https://github.com/alphagov/forms-deploy/blob/main/infra/modules/secure-bucket/main.tf

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/ai_accelerator_s3.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/ai_accelerator_s3.tf
@@ -61,7 +61,7 @@ data "aws_iam_policy_document" "https_only" {
 
 resource "aws_s3_bucket_policy" "govuk_ai_accelerator_data_bucket_policy" {
   bucket = aws_s3_bucket.govuk_ai_accelerator_data[0].id
-  policy = data.aws_iam_policy_document.https_only.json
+  policy = data.aws_iam_policy_document.https_only[0].json
 
   count = var.govuk_environment == "integration" ? 1 : 0
 }


terraform/deployments/govuk-publishing-infrastructure/ai_accelerator_s3.tf
```diff
@@ -0,0 +1,69 @@
+resource "aws_s3_bucket" "govuk_ai_accelerator_data" {
```
jfharden, line 1: Because this same terraform root is used for workspaces for integration, staging, and production, this bucket will try to be created in every environment.

You'll need to add something like


```suggestion
resource "aws_s3_bucket" "govuk_ai_accelerator_data" {
  count = var.govuk_environment == "integration" ? 1 : 0
```

to every resource, and most annoyingly, have to update every reference to those resources to specify array index 0, like chaging:

```
aws_s3_bucket.govuk_ai_accelerator_data.id
```

to 
```
aws_s3_bucket.govuk_ai_accelerator_data[0].id
```

minglis: ouch - will do

minglis: Added count block based on integration environment for wach resouce block. The bucket is referred to in the other resources, so updated to aws_s3_bucket.govuk_ai_accelerator_data[0].id or similar everywhere.

PR #3669, opened 2026-02-11
Set permissions for the AI Alpha github team to maintain. 

minglis:
This team only has access to the 3 repos that constitute the Publishing alpha. this is to allow the MSP team to manage their code within the team.

Currently they are in the setup phase of the alpha and are making frequent code changes as they setup the code.


minglis:
not needed, closed.

PR #3676, opened 2026-02-12
IA-2491: create gcp-gds-bq-processing TFC workspace to manage GCP project gds-bq-processing

aaronfowles:
`gds-bq-processing` is a GCP project with existing resources. This PR creates a TFC workspace and minimal configuration to import the project so it can be managed via terraform.

### DIFF
--- a/terraform/deployments/gcp-gds-bq-processing/README.md
+++ b/terraform/deployments/gcp-gds-bq-processing/README.md
@@ -1,0 +1,1 @@
+# gcp-gds-bq-processing
--- a/terraform/deployments/gcp-gds-bq-processing/main.tf
+++ b/terraform/deployments/gcp-gds-bq-processing/main.tf
@@ -1,0 +1,18 @@
+terraform {
+  cloud {
+    organization = "govuk"
+    workspaces {
+      project = "govuk-data-engineering"
+      name    = "gcp-gds-bq-processing"
+    }
+  }
+
+  required_providers {
+    google = {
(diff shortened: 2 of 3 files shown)


terraform/deployments/gcp-gds-bq-processing/project_iam_binding.tf
```diff
@@ -0,0 +1,25 @@
+resource "google_project_iam_binding" "project_owners" {
```
jfharden: You'll need an import statement for every resource since these already exist.

The docs pages for the individual resources should include examples of how to import, for example the iam binding:

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_project_iam#importing-iam-bindings

aaronfowles: Ok thanks. I think I'll change this this PR to just import the project. I've updated the membership of the groups to reflect current state but I agree it would be good to double-check with a plan in a separate PR.

aaronfowles: Actually, I'll just limit this to create the workspace as you (rightly) spooked me 😅

aaronfowles: Hi @jfharden, thanks for the review. I've scoped this down to just creating the workspace in this PR now. Would you mind re-reviewing?

PR #3682, opened 2026-02-13
Adding postgres17 database to support the AI Accelerator alpha

minglis:
Adds a postgres17 database for the AI Alpha project.

Uses same engine type and storage as the publishing_api as may end up working on similar large datasets.

### DIFF
--- a/terraform/deployments/tfc-configuration/variables-integration.tf
+++ b/terraform/deployments/tfc-configuration/variables-integration.tf
@@ -383,7 +383,6 @@ module "variable-set-rds-integration" {
         instance_class               = "db.m6g.large"
         performance_insights_enabled = true
         project                      = "GOV.UK - AI Accelerator"
-        snapshot_identifier          = "govuk-ai-accelerator-postgres-post-encryption"
       }
 
       link_checker_api = {


terraform/deployments/tfc-configuration/variables-integration.tf
```diff
@@ -598,6 +598,24 @@ module "variable-set-rds-integration" {
...
+          deadlock_timeout           = { value = 2500 }
+          log_lock_waits             = { value = 1 }
+        }
+        engine_params_family         = "postgres17"
+        name                         = "govuk-ai-accelerator"
+        allocated_storage            = 1000
+        instance_class               = "db.m6g.large"
+        performance_insights_enabled = true
+        project                      = "GOV.UK - AI Accelerator"
+        snapshot_identifier          = "govuk-ai-accelerator-postgres-post-encryption"
```
jfharden: You should leave this off, this launches the db from a snapshot, but you don't have a snapshot. All the others have it because of the database maintenance work

terraform/deployments/tfc-configuration/variables-integration.tf
```diff
@@ -598,6 +598,24 @@ module "variable-set-rds-integration" {
         project                      = "GOV.UK - Publishing"
         snapshot_identifier          = "whitehall-mysql-post-encryption"
       }
+
+      govuk_ai_accelerator = {
```
jfharden: They are in (almost!) alphabetical order, could you move this block to be in the correct place alphabetically

terraform/deployments/tfc-configuration/variables-integration.tf
minglis: Done

terraform/deployments/tfc-configuration/variables-integration.tf
minglis: Done

terraform/deployments/tfc-configuration/variables-integration.tf
```diff
@@ -368,6 +368,23 @@ module "variable-set-rds-integration" {
...
+        engine_params = {
+          log_min_duration_statement = { value = 10000 }
+          log_statement              = { value = "all" }
+          deadlock_timeout           = { value = 2500 }
+          log_lock_waits             = { value = 1 }
+        }
+        engine_params_family         = "postgres17"
+        name                         = "govuk-ai-accelerator"
+        allocated_storage            = 1000
+        instance_class               = "db.m6g.large"
```
dj-maisy, line 383: Just adding that `db.m7g.large` and `db.m8g.large` are also available instance types that are using more modern CPUs and higher network throughput for a similar price.

minglis: thanks, for now I'll leave it - this whole DB should get deleted once the alpha is finished, at which point will need to do some proper sizing based on usage.

PR #3689, opened 2026-02-16
[CHAT-128] Add runbook link to Cloudwatch alarms for GOV.UK Chat

davidgisbey:
## Description

We've recently made some changes to the dev docs for Chat AI alerts in this PR: https://github.com/alphagov/govuk-developer-docs/pull/5422.

We've added new sections to the docs for each alert. This means that we're now able to link to the relevant section for each alert.

This adds a runbook link to the Cloudwatch alarms for GOV.UK Chat, which will help us quickly access the relevant documentation when an alert is triggered and sent to Slack.

## Jira ticket

https://gdsgovukagents.atlassian.net/jira/software/c/projects/CHAT/boards/269?label=Backend_Dev&selectedIssue=CHAT-128

### DIFF
--- a/terraform/deployments/chat/cloudwatch_alarms.tf
+++ b/terraform/deployments/chat/cloudwatch_alarms.tf
@@ -8,7 +8,12 @@ locals {
 # Bedrock token usage over 50% alarm
 resource "aws_cloudwatch_metric_alarm" "bedrock_token_threshold_50" {
   alarm_name          = "govuk-chat-${var.govuk_environment}-bedrock-token-threshold-50"
-  alarm_description   = "WARNING - The current ${var.govuk_environment} Bedrock token usage > 50%"
+  alarm_description   = <<-EOF
+  WARNING - The current ${var.govuk_environment} Bedrock token usage > 50%
+
+  Runbook:
+  https://docs.publishing.service.gov.uk/manual/alerts/chat-ai-alerts.html#bedrock-token-threshold-alerts
+  EOF
   comparison_operator = "GreaterThanOrEqualToThreshold"
   threshold           = 50
   evaluation_periods  = 1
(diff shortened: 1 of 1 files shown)


kevindew (approved):
Changes look good to me. Only thing I'm not clear on is if using HEREDOC and multiline description is something that we have precedent for and are confident it'll work as expected. Perhaps the platform team can help confirm that if you're not sure?

terraform/deployments/chat/cloudwatch_alarms.tf
```diff
@@ -8,7 +8,12 @@ locals {
 # Bedrock token usage over 50% alarm
 resource "aws_cloudwatch_metric_alarm" "bedrock_token_threshold_50" {
   alarm_name          = "govuk-chat-${var.govuk_environment}-bedrock-token-threshold-50"
-  alarm_description   = "WARNING - The current ${var.govuk_environment} Bedrock token usage > 50%"
+  alarm_description   = <<-EOF
```
davidgisbey, line 11: Just highlighting Kev's comment [here](https://github.com/alphagov/govuk-infrastructure/pull/3689#pullrequestreview-3818700559)

> Changes look good to me. Only thing I'm not clear on is if using HEREDOC and multiline description is something that we have precedent for and are confident it'll work as expected. Perhaps the platform team can help confirm that if you're not sure?

If it's not a precedent we want to set/it'll cause any issues we could just do 

```
"WARNING - The current ${var.govuk_environment} Bedrock token usage > 50%. runbook_url: https://docs.publishing.service.gov.uk/manual/alerts/chat-ai-alerts.html#bedrock-token-threshold-alerts"
```

jaskaransarkaria: We haven't got anywhere else using a multiline string for an alarm_description but I see no reason why this wouldn't work, so happy to accept this

Issue #3645, opened 2026-02-05
[Tetragon] - Create a PoC in integration that ships logs to an encrypted s3 bucket


jaskaransarkaria:
**User Need**

**As a** Platform engineer
**I want** to record all `kubectl exec` actions in pods
**so that** we know what users need `kubectl exec` for

## Context

If the PoC is accepted, the feature would be useful in creating an audit trail of `kubectl exec` commands, which were previously invisible. This would help trace any potential attacker's activities.

---

### What’s Needed

<img width="831" height="643" alt="Image" src="https://github.com/user-attachments/assets/ab22710e-0f0c-493e-aeb3-031ca0424eae" />

- [ ] Install and configure tetragon. Prevent logs from being sent to logit (consider logging to a file)
- [ ] ship the logs to s3 (consider using fluent-bit)
- [ ] once you have logs in s3, create a script to parse and pull the data out in a meaningful way (consider using [duckdb](https://duckdb.org/))

---

### Acceptance Criteria

Clear, measurable conditions to verify the story is complete and valuable.

- [ ] we are capturing meaningful `kubectl exec` logs

---

### User Comms Plan (if applicable)

Briefly describe how you’ll let users know about the change.

---

### Assumptions (optional)

-

---

### Risks & Mitigation (optional)

- consider applying `tetragon.redactionFilters` if you are seeing common patterns of secrets leaking, Although we should be ok for the PoC as we are not outputting to stdout and shipping to an encrypted s3

---

### Notes

- The helm configuration values on github and in their docs are frustrating because you have to scroll across the table, use https://artifacthub.io/packages/helm/cilium/tetragon instead.
- be cautious of the log volume, consider the following daemonset configuration to filter out irrelevant logs

```
  export-allowlist: '{"in_init_tree":false,"namespace": ["apps"],"event_set":["PROCESS_EXEC",
    "PROCESS_EXIT", "PROCESS_KPROBE", "PROCESS_UPROBE", "PROCESS_TRACEPOINT", "PROCESS_LSM"]}'
  export-denylist: |-
    {"health_check":true}
    {"namespace":["", "kube-system"]}```

- consider logging via gRPC instead of json

jaskaransarkaria:
Next steps: https://github.com/alphagov/govuk-infrastructure/issues/3698

PR #3695, opened 2026-02-17
[POC] [Tetragon] - add Tetragon, S3 modules and Terraform docs gha

jaskaransarkaria:
1 / 2 - Create the modules to add Tetragon and ship the logs to a secure s3 bucket

relates #3645 

- add tetragon module and configure to log to a file not stdout
- add fluent bit to ship log file to s3
    - uses irsa role to push to s3
    - pushes an gzip'ed object for every hour of kubectl exec logs (pushes none if there are no logs that hour). Logs are organised into the following dir structure: `yyyy/mm/dd/hh-<hash>`
    - log format example.: 
    ```
    {<exec_event_data>}
    {<exec_event_data>}
    ```
    [link to bucket](https://430354129336-pxa5tnuj.eu-west-1.console.aws.amazon.com/s3/buckets/eph-jas-100226-eks-exec-audit-logs?region=eu-west-1&prefix=2026/02/13/&showversions=false)
- add s3 module as a "shared-module"
- add tf docs to output to `USAGE.md` following examples in their docs (this allows us to keep any README in our README and not in tf files) eg. https://github.com/terraform-docs/gh-actions/tree/v1.4.1/examples

Terraform plans are NOOPs.

The next PR will enable Tetragon with logging in `Integration`.

## Testing instructions

- spin up an ephemeral cluster from this branch with `enable_tetragon = true`
- deploy an app to the `apps` namespace and exec onto it
- put a command into a watch and wait (up to 10 mins)
- examine the files in the bucket and find you command

You can use my cluster if you prefer. init a aws on the test account and then follow the commands below:

1. aws eks update-kubeconfig --name eph-jas-100226 --kubeconfig ~/.kube/eph-jas-100226
2. export KUBECONFIG=~/.kube/eph-jas-100226
3. kubectl exec -n apps -it demo-deploy-56978fbf48-84lks -- sh
4. follow the fluent bit logs in the `logging` namespace (I use stern for that https://github.com/stern/stern)
5. find and download your logs from the s3 bucket `eph-jas-100226-eks-exec-audit-logs` (wait up to 10 mins for the logs to come through)

# Next steps

https://github.com/alphagov/govuk-infrastructure/issues/3698

### DIFF
--- a/terraform/shared-modules/s3/USAGE.md
+++ b/terraform/shared-modules/s3/USAGE.md
@@ -1,41 +1,17 @@
 <!-- BEGIN_TF_DOCS -->
-## Requirements
-
-No requirements.
-
 ## Providers
 
 | Name | Version |
 |------|---------|
 | <a name="provider_aws"></a> [aws](#provider\_aws) | n/a |
 
-## Modules
-
-No modules.
-
-## Resources
-
-| Name | Type |
-|------|------|
-| [aws_iam_policy.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/iam_policy) | resource |
-| [aws_s3_bucket.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket) | resource |
-| [aws_s3_bucket_lifecycle_configuration.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_lifecycle_configuration) | resource |
-| [aws_s3_bucket_ownership_controls.owner](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_ownership_controls) | resource |
-| [aws_s3_bucket_policy.bucket_policy](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_policy) | resource |
-| [aws_s3_bucket_public_access_block.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_public_access_block) | resource |
-| [aws_s3_bucket_server_side_encryption_configuration.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_server_side_encryption_configuration) | resource |
-| [aws_s3_bucket_versioning.this](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_versioning) | resource |
-| [aws_iam_policy_document.https_only](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document) | data source |
-| [aws_iam_policy_document.s3_combined_policy](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document) | data source |
-
 ## Inputs
 
 | Name | Description | Type | Default | Required |
 |------|-------------|------|---------|:--------:|
-| <a name="input_AES256_encryption_configuration"></a> [AES256\_encryption\_configuration](#input\_AES256\_encryption\_configuration) | Whether to use AES256 as the algorithm for server side encryption. If false, the caller should set their own configuration | `bool` | `true` | no |
+| <a name="input_name"></a> [name](#input\_name) | The bucket and IAM policy name. NOTE: This must be a globally unique name for AWS S3 | `string` | n/a | yes |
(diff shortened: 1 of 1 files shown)


terraform/shared-modules/s3/USAGE.md
```diff
@@ -0,0 +1,48 @@
+<!-- BEGIN_TF_DOCS -->
```
jfharden, line 1: I'd give the options to not show requirements, modules, or resources

terraform/shared-modules/s3/main.tf
```diff
@@ -0,0 +1,157 @@
+resource "aws_s3_bucket" "this" {
+  bucket = var.name
+
+  tags = {
+    Name = var.name
+  }
+}
+
+resource "aws_s3_bucket_lifecycle_configuration" "this" {
```
jfharden, line 9: This should probably be conditional on there being any rules specified. I think if you don't specify any this will fail to apply

terraform/shared-modules/s3/main.tf
```diff
@@ -0,0 +1,157 @@
...
+}
+
+data "aws_iam_policy_document" "s3_combined_policy" {
+  source_policy_documents = flatten([
+    data.aws_iam_policy_document.https_only.json,
+  var.extra_bucket_policies])
+}
+
+resource "aws_iam_policy" "this" {
+  name        = "${var.name}-eks-exec-audit-logs"
```
jfharden: This name isn't generic, given it's in a reusable module it should be more generic than this

terraform/shared-modules/s3/main.tf
```diff
@@ -0,0 +1,157 @@
...
+resource "aws_s3_bucket_versioning" "this" {
+  bucket = aws_s3_bucket.this.id
+
+  versioning_configuration {
+    status = var.versioning_enabled ? "Enabled" : "Disabled"
+  }
+}
+
+resource "aws_s3_bucket_server_side_encryption_configuration" "this" {
+  count  = var.AES256_encryption_configuration ? 1 : 0
```
jfharden: I wouldn't even include this as an option, just always enforce SSE.

I would say include options to vary the config of it, but it's likely we will never need it, so I'd just keep what you have below and enforce SSE with AES256. If we ever want to vary we can adapt the module once we know in what ways we want to vary it

terraform/deployments/cluster-services/modules/tetragon/fluent_bit.tf
```diff
@@ -0,0 +1,112 @@
...
+      "app.kubernetes.io/managed-by"       = "Terraform"
+      "argocd.argoproj.io/managed-by"      = "cluster-services"
+      "pod-security.kubernetes.io/audit"   = "privileged"
+      "pod-security.kubernetes.io/enforce" = "privileged"
+      "pod-security.kubernetes.io/warn"    = "privileged"
+    }
+  }
+}
+
+resource "helm_release" "fluent_bit" {
```
jfharden, line 17: Annoying that filebeat can't write to s3 :(

We do already have a fluentbit running (it's shipping logs from the kubernetes events api to elasticsearch).

I think in the PoC doing what you have done is the best idea. For productionising it in the future we should probably extract the various fluent bit configs and have them all centralised and shipping all the non-logit-logs

terraform/deployments/cluster-services/modules/tetragon/fluent_bit.tf
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    serviceAccount = {
+      create = false
+      name   = "${kubernetes_service_account_v1.this.metadata[0].name}"
+    }
+
+    securityContext = {
+      privileged               = false
+      allowPrivilegeEscalation = true
```
jfharden: Can't this be false?

jfharden (commented):
I didn't quite finish the review but I've got meetings now and at least this gets some way forward

terraform/deployments/cluster-services/modules/tetragon/fluent_bit.tf
jaskaransarkaria: That's a good point, I have a follow-up issue, of things to do if the POC is accepted, and I'll add the step to centralise the fluent bit configs

jaskaransarkaria: https://github.com/alphagov/govuk-infrastructure/issues/3698

PR #3703, opened 2026-02-18
fix: 🐛 remove extra policy block and update principals

jaskaransarkaria:
- fix malformed policy
- remove unused policy resource
- remove unnecessary output

image


nimalank7 (approved):
LGTM!

PR #3724, opened 2026-02-20
docs: ✏️ add shared tf modules

jaskaransarkaria:
Add some user facing docs to sign post users to our new shared modules dir

### DIFF
--- a/docs/terraform-shared-modules.md
+++ b/docs/terraform-shared-modules.md
@@ -14,4 +14,4 @@ module "secure_s3_bucket" {
 }
 ```
 
-Find our shared modules [here](https://github.com/alphagov/govuk-infrastructure/tree/main/terraform/shared-modules).
+The shared modules are in [the `terraform/shared-modules` directory](https://github.com/alphagov/govuk-infrastructure/tree/main/terraform/shared-modules) in this repository.


docs/terraform-shared-modules.md
```diff
@@ -0,0 +1,17 @@
...
+module "secure_s3_bucket" {
+  source = "github.com/alphagov/govuk-infrastructure/terraform/shared-modules/s3?ref=3f260111d76ce69eeb1f6b9b8d3ea52e1bd467b4"
+
+  name               = local.bucket_name
+  versioning_enabled = true
+  lifecycle_rules    = []
+}
+```
+
+Find our shared modules [here](https://github.com/alphagov/govuk-infrastructure/tree/main/terraform/shared-modules).
```
jfharden: It's bad form to say here for a link, instead something like:


```suggestion
The shared modules are stored in [the terraform/shared-modules folder](https://github.com/alphagov/govuk-infrastructure/tree/main/terraform/shared-modules) in this repository.
```

jaskaransarkaria: Argh it's annoying I had something like this earlier but Vale shouted at me that "are found" (same for "are stored") is the passive voice

PR #3725, opened 2026-02-20
docs: ✏️ fix vale issues

jaskaransarkaria:
- fix vale linting errors

### DIFF
--- a/docs/logit/logit/README.md
+++ b/docs/logit/logit/README.md
@@ -1,6 +1,6 @@
 # Intro
 
-A collection of config files (mainly Logstash) for use on https://logit.io.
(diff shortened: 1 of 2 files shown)


docs/logit/logit/README.md
```diff
@@ -1,6 +1,6 @@
 # Intro
 
-A collection of config files (mainly logstash) for use on logit.io.
+A collection of config files (mainly Logstash) for use on https://logit.io.
```
jfharden: So I don't agree with vale here, logit.io is the company name, as well as being a URL. It would probably be better to add logit.io to one of the allowed word lists

PR #3732, opened 2026-02-23
IA-2490 create dataset for raw data and add Service Accounts

aaronfowles:
Add dataset and retention period - fd7013d2f5877a95edc2583810ff9bf31bbbbbb8
Add service accounts - 6aca8f23697a1722c59f4c4a317a50d14e759a41

### DIFF
--- a/terraform/deployments/gcp-ga4-aggregate-analytics/project_iam_binding.tf
+++ b/terraform/deployments/gcp-ga4-aggregate-analytics/project_iam_binding.tf
@@ -13,6 +13,7 @@ resource "google_project_iam_binding" "project_editors" {
   role    = "roles/editor"
   members = [
     "group:gcp-ga4-aggregate-analytics-editors@digital.cabinet-office.gov.uk",
+    "serviceAccount:firebase-measurement@system.gserviceaccount.com", // Write access for GA4 exports 
   ]
 }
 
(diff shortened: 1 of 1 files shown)


terraform/deployments/gcp-ga4-aggregate-analytics/dataset.tf
Nyzl, line 1: the raw data from GA will be sharded rather than partitioned so i think we'll have to use [default_table_expiration_ms](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/bigquery_dataset#default_table_expiration_ms-1)

or maybe setting both, to cover shards and partitions

aaronfowles: Ah yes good shout. Yeah, I think you're right - under the hood the shards will be treated as tables. I'll have a read but I think setting both sounds sensible.

aaronfowles: I've added both. ready for another look whenever you have a chance 👍

terraform/deployments/gcp-ga4-aggregate-analytics/dataset.tf
```diff
@@ -0,0 +1,13 @@
+locals {
+  # 1000ms * 60s * 60m * 24h * 90 days
+  ninety_days_in_ms = 1000 * 60 * 60 * 24 * 90
```
jfharden, line 3: ❤️

jfharden (approved):

PR #3746, opened 2026-02-24
Add elasticsearch legacy subnets to test account

jfharden:
I need to deploy an opensearch cluster into my ephemeral cluster, but the es subnets don't exist in test, so this PR adds them.


jfharden:
Nope, this needs to go in the eph cluster definition instead

PR #3764, opened 2026-02-26
Create and destroy ephemeral clusters with a Make target

AP-Hunt:
## What

We're making more and more use of ephemeral clusters, so it feels right to start to put a little more tooling around them.

Also includes some small tweaks to the Terraform to make it safer to automate.

## How to review

1. Compare the code I've written to the written instructions in the readme in `terraform/deployments/ephemeral/``
2. Test it! Spin up a new one and try tear it down

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,3 +31,21 @@ lint_docs:
   	  	1>&2 echo -e "\tmake lint_docs LINT_DOCS=./docs/README.md WATCH_DOCS=true"; \
 	fi; \
 	exit $${EXIT_CODE};
+
+.PHONY: ephemeral_cluster
+ephemeral_cluster:
+	@ if [[ -z "$${EPH_CLUSTER_ID}" ]] || [[ ! -v "EPH_CLUSTER_ID" ]]; then \
+		echo "Set the ephemeral cluster id with the EPH_CLUSTER_ID variable"; \
+		printf "\t make create_ephemeral_cluster EPH_CLUSTER_ID=\"eph-new-cluster\"\n"; \
+		exit 1; \
+  	fi; \
+  	cd terraform/deployments/ephemeral; \
+	terraform init; \
+	echo "Ephemeral cluster $${EPH_CLUSTER_ID} will be built by Terraform in Terraform Cloud."; \
+	echo "When the 'cluster_access' workspace is complete you should be able to access to the cluster"; \
+	printf "\t aws eks update-kubeconfig --name $${EPH_CLUSTER_ID}\n"; \
+	echo "Once all workspaces are complete, log into the cluster and run './validate.sh' to test the cluster is functioning"; \
+	open "https://app.terraform.io/app/govuk/projects"; \
+	echo "Press enter to continue"; \
+	read; \
+	terraform apply -var ephemeral_cluster_id="$${EPH_CLUSTER_ID}"; \
--- a/README.md
+++ b/README.md
@@ -39,6 +39,19 @@ We set the constraints with minor version precision. However when using this Ter
 We have some [recommended pre-commit hooks](.pre-commit-config.yaml). You need
 to [install `pre-commit`](https://pre-commit.com/#install) for these to run.
 
+## Starting and updating an ephemeral cluster
+
+This repository supports creating ephemeral Kubernetes clusters for the purposes of short-lived tests. To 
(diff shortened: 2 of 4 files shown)


terraform/deployments/ephemeral/ws/workspace.tf
```diff
@@ -30,6 +30,8 @@ module "workspace" {
     "common-ephemeral"
   ]
 
+  force_delete = true
```
jfharden, line 33: This isn't such a good idea, it will leave real resources that have been created live in the accounts. If the cluster fails to be deleted you need to do something about it or we will have things living forever

README.md
```diff
@@ -39,6 +39,23 @@ We set the constraints with minor version precision. However when using this Ter
...
+This repository supports creating ephemeral Kubernetes clusters for the purposes of short-lived tests. To 
+create a new cluster, or update the workspaces for an existing one, use the Make target:
+
+```shell
+make ephemeral_cluster EPH_CLUSTER_ID=eph-new-cluster
+```
+
+## Stopping an ephemeral cluster
+
+To stop an existing ephemeral cluster, use the Make test:
```
jfharden:
```suggestion
To stop an existing ephemeral cluster, use the Make task:
```

Makefile
```diff
@@ -30,4 +30,31 @@ lint_docs:
...
+ephemeral_cluster:
+	@ if [[ -z "$${EPH_CLUSTER_ID}" ]] || [[ ! -v "EPH_CLUSTER_ID" ]]; then \
+		echo "Set the ephemeral cluster id with the EPH_CLUSTER_ID variable"; \
+		printf "\t make create_ephemeral_cluster EPH_CLUSTER_ID=\"eph-new-cluster\"\n"; \
+		exit 1; \
+  	fi; \
+  	cd terraform/deployments/ephemeral; \
+	terraform init; \
+	terraform apply -var ephemeral_cluster_id="$${EPH_CLUSTER_ID}"; \
+	echo "Ephemeral cluster $${EPH_CLUSTER_ID} is being built by Terraform in Terraform Cloud."; \
```
jfharden: This wont appear until the terraform apply is complete right? terraform apply will block until it's complete and terraform apply is executing runs of all the workspaces so will block until they have all complete (or one fails)?

terraform/deployments/ephemeral/ws/workspace.tf
AP-Hunt: In which case, I think it may be impossible to script the deletion. If you leave it false, Terraform Cloud says it can't delete the project because it has dependencies :/

Makefile
AP-Hunt: Bums! Yes, correct. I added the echoes while it was running the `terraform apply` for me 😂 I'll move them between the `init` and `apply`

terraform/deployments/ephemeral/ws/workspace.tf
jfharden: If you just launch a cluster, then destroy it with the shutdown script it you successfully terraform destroy it, if things drift outside the terraform state, or you create things which you don't destroy and the shutdown script doesn't know about (e.g. if you add anotrher workspace with stuff in, but don't manually do a destroy run on it (the shutdown script hard codes which workspaces to destroy)) then you end up with stuff left, but if you just force destroy the workspace all that does is delete the workspace from terraform cloud, it doesn't make it delete dependent resources

jfharden: This has me thinking, it would be better if the shutdown script enumerated all the workspaces in the ephemeral project and destroyed them all rather than hard code. We would need a dependency tree for at least the ones we know about (this is way outside the scope of this PR though), but it would means like the opensearch workspace I just made would get cleaned up

AP-Hunt: > If you just launch a cluster, then destroy it with the shutdown script it you successfully terraform destroy it

That wasn't my experience :( I ran the destroy task as it appears in this PR to kill off a cluster I'd tried to create that failed because of a name that was too long. There was no infrastructure outside the VPC, so the shutdown script didn't need to do anything, but Terraform Cloud still failed with its complaint about dependencies

I think for now I'll just remove the destruction. It's clearly something that needs more thought.

jasonBirchall:
Closing and we can reopen when ready

PR #3773, opened 2026-02-27
Use GCP creds for GCP project

JonathanHallam:


terraform/deployments/tfc-configuration/gov-graph.tf
```diff
@@ -27,7 +27,7 @@ module "gov-graph-dev" {
   }
 
   variable_set_ids = [
-    local.aws_credentials["integration"],
```
aaronfowles, line 30: I've only used the "production" service account in the work I've done so far as they were "data" projects as opposed to "app" projects with different envs. This looks sensible to me but someone from platform engineering will be able to comment with more knowledge.

PR #3774, opened 2026-02-27
Add opensearch deployment for ai-accelerator

jfharden:
**NOTE** This is already deployed, you can see it in the [ai-accelerator terraform workspace](https://app.terraform.io/app/govuk/workspaces/ai-accelerator/)

This PR adds:

* An opensearch-blue-green-deployment module which can be used to easily launch new opensearch clusters, launch blue and/or green domains with differing configs (including launching a mix of opensearch and elasticsearch domains). It includes a lot of validations which give you positive error messages when you provide invalid configuration (I hit a lot of these, so they are born out of my own learning journey)
* An ai-accelerator deployment which uses the shared module 
* Out of necessity it fixes a bug in the s3 shared module which means you can't make a logging config without a target_object_key_format

There is already [a variable set deployed in tfc-configuration](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/tfc-configuration/variables-integration.tf#L762-L786) and [a workspace which uses it](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/tfc-configuration/ai-accelerator.tf) (this currently points to my branch, once this PR is merged I will change it to point at main) 

The register snaphsot python file is lifted almost (but not quite) wholesale from the chat deployment (which was in turn lifted from the elasticsearch deployment). I do not suggest we put a lot of time into reviewing that in this PR since it could absolutely be better, but it's not possible to automate easily and is only done once ever in the life of an opensearch domain

### DIFF
--- a/terraform/shared-modules/opensearch-blue-green-deployment/.tool-versions
+++ b/terraform/shared-modules/opensearch-blue-green-deployment/.tool-versions
@@ -1,1 +1,0 @@
-terraform 1.14.6


terraform/deployments/ai-accelerator/opensearch.tf
```diff
@@ -0,0 +1,17 @@
+module "opensearch" {
+  source = "../../shared-modules/opensearch-blue-green-deployment"
```
jaskaransarkaria, line 2: I would recommend pointing this towards the git url and hash so the module can be pinned

terraform/shared-modules/opensearch-blue-green-deployment/README.md
```diff
@@ -0,0 +1,145 @@
...
+## Usage
+
+See the examples below, and see [USAGE.md](./USAGE.md) for a complete list of all options with descriptions.
+
+## Examples
+
+### Just launch a blue domain with minimal config
+```tf
+module "opensearch" {
+  source = "../../shared-modules/opensearch-blue-green-deployment"
```
jaskaransarkaria, line 21: If you swap out and use the github url, you should update that here too

terraform/shared-modules/opensearch-blue-green-deployment/README.md
```diff
@@ -0,0 +1,145 @@
...
+}
+```
+
+### Launch a Blue and Green cluster 
+
+**NOTE**: You would switch which the CNAME points to by changing `current_live_domain` to `green`.
+
+```tf
+module "opensearch" {
+  source = "../../shared-modules/opensearch-blue-green-deployment"
```
jaskaransarkaria, line 56: and here

terraform/shared-modules/opensearch-blue-green-deployment/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+  aws_region = "eu-west-1"
+}
+```
+
+### Complete configuration example launching only a blue cluster
+
+```tf
+module "opensearch" {
+  source = "../../shared-modules/opensearch-blue-green-deployment"
```
jaskaransarkaria, line 102: and here

terraform/shared-modules/opensearch-blue-green-deployment/s3.tf
```diff
@@ -0,0 +1,55 @@
+locals {
+  bucket_suffix        = var.s3_bucket_custom_suffix == null ? "opensearch-snapshots" : var.s3_bucket_custom_suffix
+  snapshot_bucket_name = "govuk-${var.govuk_environment}-${var.opensearch_domain_name}-${local.bucket_suffix}"
+}
+
+module "snapshot_bucket" {
+  # soource = "github.com/alphagov/govuk-infrastructure/terraform/shared-modules/s3?depth=1&ref=main"
+  source = "../s3"
```
jaskaransarkaria, line 8: Annoying you can't use the commit hash to pin because you need to add in the change first lol

terraform/deployments/ai-accelerator/opensearch.tf
jfharden: Yeah I will, but I don't want to do it in this PR since the commit it's pinned to might never appear in main. I'll do a follow up PR where I just change these.

terraform/shared-modules/opensearch-blue-green-deployment/README.md
jfharden: In the docs I'm not keen on putting an actual sha, but I can put a placeholder (we can discuss that in the PR that changes this :))

PR #3776, opened 2026-02-27
chore: 🤖 grant dgu perms to update their argo apps in ui

jaskaransarkaria:
https://gds.slack.com/archives/C013F737737/p1772103958506269


terraform/deployments/cluster-services/argo.tf
```diff
@@ -33,6 +33,9 @@ locals {
     g, ${var.github_national_data_library_team}, role:readonly
```
samsimpson1, line 33: Will this cause problems with the new role assignment on line 38?

jaskaransarkaria: No, I've tested in staging with Ken

PR #3799, opened 2026-03-04
BAU - Expand National Data Library ArgoCD permissions

nimalank7:
Description:
- Give NDL developers ability to manually sync applications in the `datagovuk` namespace
- Give NDL developers ability to sync to an arbitrary branch in the `datagovuk` namespace to allow for testing custom branches
- See https://argo-cd.readthedocs.io/en/stable/operator-manual/rbac/#the-override-action

### DIFF
--- a/terraform/deployments/cluster-services/argo.tf
+++ b/terraform/deployments/cluster-services/argo.tf
@@ -35,6 +35,10 @@ locals {
     g, ${var.github_read_write_team}, role:admin
     p, role:nationaldatalibrary, applications, update, datagovuk/*, allow
     p, role:nationaldatalibrary, applications, update, default/dgu-app-of-apps, allow
+    p, role:nationaldatalibrary, applications, sync, datagovuk/*, allow
+    p, role:nationaldatalibrary, applications, sync, default/dgu-app-of-apps, allow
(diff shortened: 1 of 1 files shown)


terraform/deployments/cluster-services/argo.tf
```diff
@@ -75,7 +80,8 @@ resource "helm_release" "argo_cd" {
 
     configs = {
       cm = {
-        url = "https://${local.argo_host}"
+        url                                                        = "https://${local.argo_host}"
+        "application.sync.requireOverridePrivilegeForRevisionSync" = "true"
```
jaskaransarkaria: Will this affect non dgu users?

jaskaransarkaria: Will they also require the override?

nimalank7: Upon reflection I decided to remove `"application.sync.requireOverridePrivilegeForRevisionSync" = "true"` to keep the existing installation.

Non-DGU users won't need an override as 99.9% of GOV.UK Developers have `admin` permission to ArgoCD (via the `gov-uk-production-deploy` GitHub team) which allows the `override`.

PR #3800, opened 2026-03-04
Use custom service account for search dataform

emmalowe:
GCP is enforcing a new access control model for Dataform called "strict act-as mode". As part of this, GCP is disabling the ability for Dataform instances to be run using the Default Dataform service Account, so all Dataform workflows must switch to a custom service account, which the Dataform Service Agent is given permissions on. The custom service account must include the iam.serviceAccounts.actAs permission to configure Dataform workflows.

Steps covered here:
- Move creation of custom service account "dataform-sa"
- Change BigQuery internal project permissions from default service account to custom service account
- Give default service account permissions to impersonate custom service account
- Add custom service account to repo set up (to use this as the default account for running workflows)
- Give custom service account secrets permissions to connect to our dataform Github repo
- Give custom service accounts permission to read from the GA4 Analytics project

See:
- Background on "strict act-as" mode: https://docs.cloud.google.com/dataform/docs/strict-act-as-mode
- GCP guidance on granting Dataform the required access: https://docs.cloud.google.com/dataform/docs/access-control#grant-dataform-required-access
- Service account migration guidance: https://gov-uk.atlassian.net/wiki/spaces/GIAT/pages/5133271058/Dataform+Service+Account+Migration

### DIFF
--- a/terraform/deployments/ga4-analytics/project_iam_custom_roles.tf
+++ b/terraform/deployments/ga4-analytics/project_iam_custom_roles.tf
@@ -4,7 +4,8 @@ resource "google_project_iam_custom_role" "gds_bigquery_read_access" {
     "bigquery.datasets.get",
     "bigquery.datasets.getIamPolicy",
     "bigquery.tables.get",
-    "bigquery.tables.getData"
+    "bigquery.tables.getData",
+    "bigquery.tables.list"
   ]
   role_id = "GDS_BQ_read_access"
   title   = "GDS BQ read access"
--- a/terraform/deployments/ga4-analytics/dataset_iam_binding.tf
+++ b/terraform/deployments/ga4-analytics/dataset_iam_binding.tf
@@ -1,0 +1,21 @@
+locals {
+  members = [
+    "serviceAccount:dataform-sa@search-api-v2-integration.iam.gserviceaccount.com",
+    "serviceAccount:dataform-sa@search-api-v2-staging.iam.gserviceaccount.com",
+    "serviceAccount:dataform-sa@search-api-v2-production.iam.gserviceaccount.com",
+  ]
+}
+
+resource "google_bigquery_dataset_iam_member" "flattened_dataset_reader" {
+  for_each   = toset(local.members)
+  dataset_id = "${google_project.project.project_id}.flattened_dataset"
+  role       = google_project_iam_custom_role.gds_bigquery_read_access.name
+  member     = each.key
+}
+
+resource "google_bigquery_dataset_iam_member" "events_reader" {
+  for_each   = toset(local.members)
+  dataset_id = "${google_project.project.project_id}.analytics_330577055"
+  role       = google_project_iam_custom_role.gds_bigquery_read_access.name
+  member     = each.key
+}


emmalowe:
Since this is Terraform, I assume the approach to testing is to get this merged and then test it out on integration before applying the changes in production.

terraform/deployments/search-api-v2/dataform.tf
```diff
@@ -106,22 +107,36 @@ resource "google_dataform_repository_workflow_config" "search-monthly" {
...
+  role               = "roles/iam.serviceAccountUser"
+  member             = "serviceAccount:service-${var.gcp_project_number}@gcp-sa-dataform.iam.gserviceaccount.com"
+}
+
+resource "google_service_account_iam_member" "dataform_sa_token_creator" {
+  service_account_id = google_service_account.dataform_service_account.name
+  role               = "roles/iam.serviceAccountTokenCreator"
+  member             = "serviceAccount:service-${var.gcp_project_number}@gcp-sa-dataform.iam.gserviceaccount.com"
+}
+
```
aaronfowles, line 127: I thought that specifying the new service account on the `google_dataform_repository` as you've done [here](https://github.com/alphagov/govuk-infrastructure/pull/3800/changes#diff-90d8f65767027b57cfad51520af370711202ce17c57134ced8476792c342da22R33) would make the default SA (and any changes to it) not required. But if you've found through testing that it actually needs that then that's fine.

@oliverroberts1-gds I think I'd have to defer to you on this one as you've been much closer to the SA changes than I have...

aaronfowles: Yes, looking at the [docs](https://docs.cloud.google.com/dataform/docs/access-control#about-service-accounts) I think you've done the right thing. But I'll let @oliverroberts1-gds confirm 😄.

emmalowe:
> The terraform [plan](https://app.terraform.io/app/govuk/workspaces/search-api-v2-integration/runs/run-qADqWwUn5v1WuK4d) looks like it failed with what could be a transient error. Might be worth triggering a retry from the UI to see if that resolves...

Thanks @aaronfowles ! I reran the plan and it looks like it's working now.

terraform/deployments/search-api-v2/dataform.tf
oliverroberts1-gds: @aaronfowles We do need to give the Dataform Service Agent both serviceAccountUser, and tokencretor on the new custom service account. Tested with Ed Billet's instance as it would throw permission issues otherwise.

emmalowe: Thanks @oliverroberts1-gds - does this PR look ok to approve then? (I'll ask @hannako and someone from Platform Engineering to approve as well, so responsibility is not all on you!)

oliverroberts1-gds: I don't have permission to approve, but looks good on my end @emmalowe

terraform/deployments/search-api-v2/dataform.tf
```diff
@@ -106,22 +107,36 @@ resource "google_dataform_repository_workflow_config" "search-monthly" {
   }
 }
 
+# Permissions for Default Dataform Service Agent to impersonate our custom Dataform Service Account
+# This allows Dataform to set up and run new workflows
+resource "google_service_account_iam_member" "dataform_sa_impersonation" {
```
emmalowe: I'm wondering if I should use `google_service_account_iam_binding` for the serviceAccountUser and serviceAccountTokenCreator roles here, since (I think) I want to only allow the default service account to impersonate our custom one.

emmalowe:
Hello reviewers 👋🏻  Apologies for sitting on this a while - I realised I had to do some more thinking on cross-project permissions before I could merge this. The short version is, our Dataform pipelines need permission to read from the GA4 Analytics Project, so I've added those in a [new commit](https://github.com/alphagov/govuk-infrastructure/pull/3800/changes/e6323a68e98cebb662aa4f3594c3ab75a52498eb).

@AP-Hunt it might be most efficient if you could review this first, please? Most of my questions (that I'll leave inline) are about Terraform best practice, so you might be best to answer those. @hannako feel free to have a look too, since we've been chatting about this already.

Thanks 🌟

terraform/deployments/ga4-analytics/project_iam_binding.tf
```diff
@@ -71,3 +70,25 @@ resource "google_project_iam_binding" "gds_log_alert_writer" {
     "group:govuk-performance-analysts@digital.cabinet-office.gov.uk"
   ]
 }
+
+locals {
```
emmalowe: Since we only need to read from 2/17 datasets in the GA4 Analytics project, I've added these as dataset permissions instead of project permissions (which would give access to all the datasets). This might be overkill.

I've chosen to use `iam_member` instead of `iam_binding` for two reasons:

1. I'm not confident how dataset permissions interact with project permissions, and I didn't want to wipe out any permissions for anyone else. From what I've read, dataset binding and project bindings together are additive, so that shouldn't be the case, but I'm not sure.
2. It seems like some of GA4 Analytics permissions have been set up via ClickOps and again, I don't want to delete permissions for anyone else. But maybe I need to do more work to check what is there and rectify any issues because we probably should use binding if we can (see point 1).

If the permission set up looks okay, I'm not clear on how this new code should be organised. I've seen locals.tf files elsewhere. Maybe we should also have a separate file for dataset permissions 🤷🏻‍♀️

terraform/deployments/ga4-analytics/project_iam_custom_roles.tf
```diff
@@ -10,6 +10,19 @@ resource "google_project_iam_custom_role" "gds_bigquery_read_access" {
   title   = "GDS BQ read access"
 }
 
+resource "google_project_iam_custom_role" "gds_bigquery_read_and_list_access" {
```
emmalowe: I'm adding this custom role, because for some of our pipelines that use a table wildcard, we need the `bigquery.tables.list` permission, which isn't in `gds_bigquery_read_access`. This also might be overkill, since I could just add `bigquery.tables.list` to `gds_bigquery_read_access`.

emmalowe: I think because I'm only adding a list permission (rather than create or delete), it's not worth making a separate role, but I'm curious to know what other people think.

aaronfowles: I tend to agree that extending the existing `gds_bigquery_read_access` role with `bigquery.tables.list` permission feels like the right thing to do. As you say, the ability to list feels like something the read role should be allowed to do.

terraform/deployments/ga4-analytics/project_iam_binding.tf
AP-Hunt: I think that's a good and sensible decision :+1:

aaronfowles: on point 2 - the GA4 User Admin tool also manages permissions on these datasets. Because that's a lot of users and cross-gov is was decided to use a tool to manage that rather than terraform. For that reason, I think you'll have to use the non-authoritative method to avoid blatting those existing permissions.

This does look like a sensible approach but I think we'd just to need to make sure we (the Insights & Analytics team) are happy to have dataset-level permissions configured in two different places (i.e. via the tool and in terraform) and document that accordingly. @Nyzl do you have any thoughts on this?

emmalowe: Ah interesting - thanks for that context @aaronfowles, this is the first I've heard of that tool! Do you have a link, or more info you could share about it?

aaronfowles: @emmalowe sure, here are some user-facing [docs](https://docs.data-community.publishing.service.gov.uk/tools/ga4-user-admin-tool/#ga4-user-admin-tool) and the [repo](https://github.com/alphagov/ga4-user-admin).

Migrating the terraform for that repo to TFC is a bit tricky for various reasons and hasn't been tackled yet so still lives in the app repo.

Nyzl: yeah i'd be happy with this approach, however i'd prefer the config to live in a separate file because it's not project wide iam, so living in something like dataset_iam_binding.tf would be more appropriate

emmalowe:
Thanks all for your comments. All should be addressed now, if you wouldn't mind giving this a re-review
@AP-Hunt @aaronfowles @Nyzl @hannako

Issue #3710, opened 2026-02-19
✨ Provision Neptune database for GOV.UK AI Accelerator in integration


jasonBirchall:
**User Need**

**As a** GOV.UK AI Accelerator developer
**I want** an Amazon Neptune graph database provisioned in the integration environment
**so that** the AI Accelerator application can store and query graph data (e.g. content relationships, knowledge graphs) in integration.

---

### Context

The GOV.UK AI Accelerator project needs an Amazon Neptune graph database in the integration environment. This is entirely new infrastructure — there are no existing Neptune resources anywhere in the govuk-infrastructure repository.

There is some related prior art in the repository: the `terraform/deployments/gcp-gov-graph/` deployment manages a GCP-based gov-graph project, suggesting graph-based data modelling is an established concept within GOV.UK.

We should create a new `terraform/deployments/ai-accelerator-neptune/` deployment with its own TFC workspace. This keeps Neptune cleanly isolated, follows the pattern established by other data store deployments (opensearch, elasticsearch, elasticache, rds), and makes future promotion straightforward.

Additionally, Neptune offers two deployment models:

- **Neptune Serverless** — scales capacity automatically, good for unpredictable or bursty workloads, simpler to manage, pay-per-use. Well-suited for an integration/experimental environment.
- **Neptune Provisioned** — fixed instance types, more predictable costs at sustained load, more configuration options.

The owner of this ticket should confirm the deployment model with the AI Accelerator team.

Raised by @minglis (Lead Tech Architect, Publishing team). We should work closely with Martyn on this ticket — Neptune is new to our estate, and he has the context on the graph data use case, query language preferences, and expected workload.

---

### What's Needed

- [ ] An Amazon Neptune cluster provisioned in the integration AWS account
- [ ] VPC placement within existing private subnets
- [x] Security group restricting access to the EKS cluster and appropriate services
- [x] Encryption at rest and in transit enabled
- [x] CloudWatch logging enabled (audit logs at minimum)
- [x] A corresponding TFC workspace configured in `terraform/deployments/tfc-configuration/`
- [x] Route53 CNAME record for internal DNS resolution (consistent with other data store patterns)
- [ ] Integration environment only — no staging or production resources at this stage
- [x] Tag the infrastructure correctly [(as per GDS way)](https://gds-way.digital.cabinet-office.gov.uk/manuals/aws-tagging.html#mandatory)

---

### Acceptance Criteria

[...]

PR #3803, opened 2026-03-04
Add neptune module

jaskaransarkaria:
- add neptune module
- add neptune workspace
- add neptune subnets (integration and test)
- deploy neptune module to integration for ai-accelerators

✅ closes: https://github.com/alphagov/govuk-infrastructure/issues/3710

### DIFF
--- a/terraform/deployments/neptune/kms.tf
+++ b/terraform/deployments/neptune/kms.tf
@@ -1,0 +1,122 @@
+resource "aws_kms_key" "neptune" {
+  description = "GOV.UK Key for Encrypting Neptune databases"
+  key_usage   = "ENCRYPT_DECRYPT"
+  policy      = data.aws_iam_policy_document.neptune_kms.json
+}
+
+resource "aws_kms_alias" "neptune" {
+  name          = "alias/govuk/neptune"
+  target_key_id = aws_kms_key.neptune.key_id
+}
+
+data "aws_iam_policy_document" "neptune_kms" {
+  statement {
+    sid = "Allow terraform cloud to manage the key"
+
+    principals {
+      type        = "AWS"
+      identifiers = ["arn:aws:iam::${data.aws_caller_identity.current.account_id}:role/terraform-cloud"]
+    }
+
+    actions = [
+      "kms:CancelKeyDeletion",
+      "kms:Create*",
+      "kms:Delete*",
+      "kms:Describe*",
+      "kms:Disable*",
+      "kms:Enable*",
+      "kms:Get*",
+      "kms:List*",
+      "kms:Put*",
+      "kms:Revoke*",
+      "kms:ScheduleKeyDeletion",
+      "kms:Update*",
+    ]
+
+    resources = ["*"]
+  }
+
+  statement {
+    sid = "Allow fulladmin roles to manage the key"
+
+    principals {
+      type        = "AWS"
+      identifiers = ["*"]
+    }
+
+    actions = [
+      "kms:CancelKeyDeletion",
+      "kms:Create*",
+      "kms:Delete*",
+      "kms:Describe*",
+      "kms:Disable*",
+      "kms:Enable*",
+      "kms:Get*",
+      "kms:List*",
+      "kms:Put*",
+      "kms:Revoke*",
+      "kms:ScheduleKeyDeletion",
+      "kms:Update*",
+    ]
+
+    resources = ["*"]
+
+    condition {
+      test     = "StringLike"
+      variable = "aws:PrincipalArn"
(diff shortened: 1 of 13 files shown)


terraform/deployments/neptune/irsa.tf
```diff
@@ -0,0 +1,49 @@
...
+
+  statement {
+    sid = "AllowDataAccessForNeptune"
+
+    effect = "Allow"
+
+    actions = [
+      "neptune-db:*"
+    ]
+    resources = [
```
samsimpson1: How come these are done differently? Is the neptune cluster ARN different from the one you construct below it?

terraform/deployments/neptune/kms.tf
```diff
@@ -0,0 +1,122 @@
...
+    condition {
+      test     = "StringEquals"
+      variable = "kms:CallerAccount"
+      values   = [data.aws_caller_identity.current.account_id]
+    }
+
+    condition {
+      test     = "StringEquals"
+      variable = "kms:ViaService"
+      values   = ["rds.${var.aws_region}.amazonaws.com"]
```
samsimpson1, line 100: Is this meant to be rds?

terraform/deployments/neptune/neptune.tf
```diff
@@ -0,0 +1,149 @@
...
+  storage_encrypted                     = true
+  storage_type                          = each.value.storage_type
+  kms_key_arn                           = aws_kms_key.neptune.arn
+  neptune_subnet_group_name             = aws_neptune_subnet_group.this.name
+  availability_zones                    = local.azs
+  iam_database_authentication_enabled   = true
+  iam_roles                             = each.value.iam_roles
+
+  dynamic "serverless_v2_scaling_configuration" {
+    for_each = try(each.value.serverless_config, null)[*]
```
samsimpson1, line 101: Does this work if serverless_config isn't set? I'd be surprised if `null[*]` is valid

terraform/deployments/neptune/service_account.tf
```diff
@@ -0,0 +1,52 @@
...
+      "app.kubernetes.io/instance" = "neptune"
+    }
+  }
+
+  depends_on = [
+    module.iam_assumable_role
+  ]
+}
+
+resource "kubernetes_cluster_role_v1" "this" {
```
samsimpson1: Why does this need a ClusterRole?

terraform/deployments/neptune/service_account.tf
```diff
@@ -0,0 +1,52 @@
+resource "kubernetes_service_account_v1" "this" {
```
samsimpson1: Pattern we normally use for service accounts is that they're managed by the chart that manages the app, not in TF

terraform/deployments/neptune/kms.tf
jaskaransarkaria: yeah under the hood it's actually rds

terraform/deployments/neptune/irsa.tf
jaskaransarkaria: yeah the arns are different the first is an rds one and the second is `neptune-db`

terraform/deployments/neptune/service_account.tf
jaskaransarkaria: Good catch, it doesn't need to be.

samsimpson1: Does it need a role at all?

terraform/deployments/neptune/service_account.tf
jaskaransarkaria: In this case there is no helm chart _and_ we need to attach our irsa annotation to it.

terraform/deployments/neptune/neptune.tf
jaskaransarkaria: Yeah this should work https://developer.hashicorp.com/terraform/language/expressions/splat#single-values-as-lists

terraform/deployments/neptune/service_account.tf
jaskaransarkaria: I think you're right actually, I'll remove it

terraform/deployments/neptune/service_account.tf
samsimpson1: We do the IRSA annotation in helm too: https://github.com/alphagov/govuk-helm-charts/blob/main/charts/app-config/values-production.yaml#L561-L565

Personally I feel like this should be managed with the rest of the kubernetes resources. Its a bit awkward having some of your k8s resources managed via helm and others via terraform

Issue #3820, opened 2026-03-06
Use tfvars files for variables in Terraform Cloud workspaces


samsimpson1:
**User Need**

**As a** Terraform user
**I want** to use tfvars files instead of TFC var sets to manage TF variables
**so that** I can easily understand Terraform changes in pull requests

---

### Context

(Add any extra context, links, diagrams, stakeholder interviews, etc.)

---

### What’s Needed

List anything the solution must do or be (behaviour, performance, security, UX, etc.).

- [x] Update the terraform-govuk-tfe-workspacer module to support tfvars files
- [x] Create equivalent tfvars files for each TFC var set
- [x] Update all TFC workspaces to use these new tfvars files instead of variable sets
- [x] Remove the now-unused variable sets
- [x] Create blank tfvars files for each workspace (for future use)

---

### Acceptance Criteria

Clear, measurable conditions to verify the story is complete and valuable.

- [x] Variable sets are no longer used for configuring TFC workspaces (with exceptions such as AWS credentials and secrets)
- [x] Terraform plans are run against affected workspaces in PRs that update variables
- [x] Documentation is updated to reflect the new way of configuring TFC workspaces

---

### User Comms Plan (if applicable)

Briefly describe how you’ll let users know about the change.

---

### Assumptions (optional)

-

---

### Risks & Mitigation (optional)

-

---

samsimpson1:
* the workspacer module has been updated to take a list of tfvars files to include in plans and applies
* all variable sets in govuk-infrastructure have been converted to tfvars files
* the variable sets have been replaced with tfvars files on all workspaces
* for workspaces that don't already have their own specific tfvars files, placeholder ones have been created to make adding new variables to these workspaces easier in the future
* all mentions of how to update variables in govuk-developer-docs have been updated to reflect the changes

PR #3823, opened 2026-03-06
Use tfvars files for common variables

samsimpson1:
Replace the common variable set with tfvars files in all workspaces


samsimpson1:
This PR will be split into smaller, more reviewable pieces.

PR #3828, opened 2026-03-09
Specify principal * for OpenSearch username and password auth

jfharden:
We're doing username and password auth which means we need to set "Principal" to "*". The terraform docs say to set both type to "*" and identifiers to `["*"]` to achieve this.

### DIFF
--- a/terraform/shared-modules/opensearch-blue-green-deployment/opensearch-domain/main.tf
+++ b/terraform/shared-modules/opensearch-blue-green-deployment/opensearch-domain/main.tf
@@ -84,8 +84,8 @@ data "aws_iam_policy_document" "opensearch_domain" {
     sid = "AllowOpenSearchAccessFromThisAccount"
 
     principals {
-      type        = "AWS"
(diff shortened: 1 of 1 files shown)


terraform/shared-modules/opensearch-blue-green-deployment/opensearch-domain/main.tf
```diff
@@ -84,8 +84,8 @@ data "aws_iam_policy_document" "opensearch_domain" {
     sid = "AllowOpenSearchAccessFromThisAccount"
 
     principals {
-      type        = "AWS"
-      identifiers = [data.aws_caller_identity.current.account_id]
+      type        = "*"
```
samsimpson1, line 87: Your comment says this should be set to an empty string

jfharden: 🫠 the comment should say `"*"`

From https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document#principals-and-not_principals

PR #3848, opened 2026-03-09
Migrate CSP Bucket to Secure S3 Module

dj-maisy:
## What?
This rewrites the CSP Reporter Bucket configuration to use the new shared "Secure S3 Bucket" module.

### DIFF
--- a/terraform/deployments/csp-reporter/buckets.tf
+++ b/terraform/deployments/csp-reporter/buckets.tf
@@ -1,6 +1,8 @@
 module "secure_s3_bucket_csp_reports" {
-  source = "github.com/alphagov/govuk-infrastructure/terraform/shared-modules/s3?ref=3f260111d76ce69eeb1f6b9b8d3ea52e1bd467b4"
-  name   = "govuk-${var.govuk_environment}-csp-reports"
+  source = "../../shared-modules/s3"
+
+  govuk_environment = var.govuk_environment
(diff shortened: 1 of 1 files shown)


terraform/deployments/csp-reporter/buckets.tf
```diff
@@ -1,44 +1,20 @@
-# Bucket to store data from Kinesis Firehose, stores both successes and errors
-resource "aws_s3_bucket" "csp_reports" {
-  bucket = "govuk-${var.govuk_environment}-csp-reports"
+module "secure_s3_bucket_csp_reports" {
+  source = "github.com/alphagov/govuk-infrastructure/terraform/shared-modules/s3?ref=3f260111d76ce69eeb1f6b9b8d3ea52e1bd467b4"
```
AP-Hunt: Are we not able to reference the module by file path?

terraform/deployments/csp-reporter/buckets.tf
```diff
@@ -1,44 +1,20 @@
...
-    }
-    actions   = ["s3:*"]
-    resources = ["${aws_s3_bucket.csp_reports.arn}/*"]
-    condition {
-      test     = "Bool"
-      values   = [false]
-      variable = "aws:SecureTransport"
-    }
-  }
+moved {
```
AP-Hunt, line 19: We could also add moved blocks for the bucket policy and lifecycle configuration

PR #3849, opened 2026-03-10
Use tfvars files for all govuk-publishing-infrastructure workspaces

samsimpson1:

### DIFF
--- a/terraform/deployments/tfc-configuration/govuk-publishing-infrastructure.tf
+++ b/terraform/deployments/tfc-configuration/govuk-publishing-infrastructure.tf
@@ -8,7 +8,11 @@ module "govuk-publishing-infrastructure-integration" {
   terraform_version = var.terraform_version
   execution_mode    = "remote"
   working_directory = "/terraform/deployments/govuk-publishing-infrastructure/"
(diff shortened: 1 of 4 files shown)


terraform/deployments/tfc-configuration/govuk-publishing-infrastructure.tf
```diff
@@ -8,7 +8,12 @@ module "govuk-publishing-infrastructure-integration" {
   terraform_version = var.terraform_version
   execution_mode    = "remote"
   working_directory = "/terraform/deployments/govuk-publishing-infrastructure/"
-  trigger_patterns  = ["/terraform/deployments/govuk-publishing-infrastructure/**/*"]
+  trigger_patterns = [
+    "/terraform/deployments/govuk-publishing-infrastructure/**/*",
+    "/terraform/variables/integration/common.tfvars",
+    "/terraform/variables/integration/amazonmq.tfvars",
```
AP-Hunt: Are the `amazonmq` set of variables used by another root? If not, I think we should merge them into the `govuk-publishing-infrastructure` file. That way every root has exactly 1 file for variables unique to it.

samsimpson1: They are not. I'll move them to the publishing-infra file.

PR #3850, opened 2026-03-10
Use tfvars files for all gov-graph workspaces

samsimpson1:


terraform/deployments/tfc-configuration/gov-graph.tf
```diff
@@ -26,9 +29,12 @@ module "gov-graph-dev" {
     "GOV.UK Production"           = "write"
   }
 
+  tfvars_files = [
+    "integration/gov-graph.tfvars"
```
AP-Hunt: In all other instances, the tfvars file is named exactly the same as the directory the root lives in. We should continue that here.

samsimpson1: This is the only instance of the use of 'dev' as an environment name in our entire estate. Do we want to make an exception here just for this one workspace?

nimalank7: Yeah I'm with @samsimpson1 as otherwise we'd have to create a directory for `dev` with just one `tfvars` file in.

AP-Hunt: Sorry, I didn't explain myself very well. I meant that the root is called `gcp-gov-graph`  under `terraform/deployments/`. Every other tfvars file is named the same as the root. I'm not as worried about having this one having its env name and tfvars directory not line up. Env name is almost an implementation detail.

samsimpson1: Ah, I understand now. That makes sense. I'll rename the file, and maybe look at correcting the name of the module later on.

PR #3879, opened 2026-03-11
BAU - Refactor `gcp-gov-graph` Terraform

nimalank7:
Description:
- Currently the Terraform will create 9 sets of all the resources in https://github.com/alphagov/govuk-infrastructure/tree/main/terraform/deployments/gcp-gov-graph/modules (e.g. `govuk-graph`dev` workspace will create `module.integration`, `module.staging`, `module.production` etc...)
- Instead each workspace will only create the GCP resources in the corresponding folder (e.g. `govuk-graph-dev` will only create GCP resources in the integration project etc...) which can be done without the modules
- Move the additional variables the modules defined into `tfc-configuration`


terraform/deployments/gcp-gov-graph/variables.tf
```diff
@@ -20,13 +30,44 @@ variable "google_cloud_billing_account" {
...
+    "cloudfunctions.googleapis.com",
+    "run.googleapis.com",
+    "cloudscheduler.googleapis.com",
+    # Required for observability
+    "logging.googleapis.com",
+    "monitoring.googleapis.com",
+  ]
 }
 
-variable "tfe_project_name" {
```
JonathanHallam, line 28: I think renaming this has caused some errors

nimalank7: Ah I forgot to add this back in I've done that now.

PR #3901, opened 2026-03-12
[CHAT-341] Amend GOV.UK Chat cloudwatch integration

davidgisbey:
## Description 

 We had some issues with the the tpm threshold being exceeded when large amounts of content ingestion was taking place. We've decided to use multiple regions for titan embedding invocations where we use 

- eu-west-1 for embeddings related to user questions
- eu-west 2 for embeddings related to document ingestion

This PR does a few things:

- configures /aws/bedrock log group with 30 day retention for eu-west-2 to support
- updates the Bedrock IAM role to allow writing to the new log group
- configures the new log group
- updates current cloudwatch alarms to specify they're using dublin
- duplicates the 2 titan dublin alarms and sets the region to eu-west-2 (london). Also updates the descriptions to make it clear which region it's configured to alert on

## Jira ticket 

https://gdsgovukagents.atlassian.net/jira/software/c/projects/CHAT/boards/269?label=Backend_Dev&selectedIssue=CHAT-341


terraform/deployments/chat/cloudwatch.tf
```diff
@@ -1,4 +1,10 @@
-resource "aws_cloudwatch_log_group" "bedrock_log_group" {
+resource "aws_cloudwatch_log_group" "bedrock_log_group_dublin" {
   name              = "/aws/bedrock"
   retention_in_days = 30
```
kevindew, line 3: should we actually put the explicit region in here as well, as we name it in the resource name it feels like it's a bit strange that it's implicit

davidgisbey:
this had merge conflicts so i resolved them. Sorry if that messes with your review. Edit: sorry i can see you just left the one comment.

terraform/deployments/chat/cloudwatch.tf
davidgisbey: Yea good shout i'll sort that.

PR #3904, opened 2026-03-12
BAU - Refactor `gcp-gov-graph` variables

nimalank7:
Description:
- Refactor Terraform variables

### DIFF
--- a/terraform/deployments/gcp-gov-graph/variables.tf
+++ b/terraform/deployments/gcp-gov-graph/variables.tf
@@ -10,11 +10,6 @@ variable "tfc_organization_name" {
   description = "The name of the Terraform Cloud organization"
 }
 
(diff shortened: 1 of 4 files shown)


terraform/deployments/gcp-gov-graph/main.tf
```diff
@@ -25,6 +25,11 @@ locals {
   display_name = title(var.name)
 }
 
+data "tfe_project" "govuk-data-engineering" {
```
samsimpson1: What is this data source for? Only the name is used, so it just looks like you've replaced a TF variable with a hardcoded value

nimalank7: Yeah true. Let me re-introduce the variable again.

PR #3907, opened 2026-03-12
Remove redundant code

JonathanHallam:
Originally the gov-graph project was copied from search api. There is a good amount on terraform that's specific to search that can and should be removed

### DIFF
--- a/terraform/deployments/gcp-gov-graph/imports.tf
+++ b/terraform/deployments/gcp-gov-graph/imports.tf
@@ -1,0 +1,4 @@
+import {
+  id = var.project_id
+  to = google_project.environment_project
+}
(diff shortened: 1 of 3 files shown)


terraform/deployments/gcp-gov-graph/imports.tf
```diff
@@ -0,0 +1,4 @@
+import {
+  id = var.project_id
```
nimalank7, line 2: This doesn't make any sense. You're importing a variable? I think you meant to import an actual GCP project in?

JonathanHallam: The Id's just a string that points to the project right? So I think this just points to the appropriate project for each environment

PR #3909, opened 2026-03-12
Create secure bucket for new DB Dumps Process

dj-maisy:
## What?
This creates a new S3 bucket using the "secure bucket" module for our new DB dumps process. It includes a policy to allow the backup script to have access to the bucket across the necessary environments, and:

* Has data retention set to 30 days
* Has versioning disabled to reduce storage bloat


dj-maisy:
Re-created in another PR and merged

Issue #3754, opened 2026-02-25
Migrate terraform/deployments/govuk-publishing-infrastructure/db_backup_s3.tf to S3 Shared Module


dj-maisy:
See the parent issue https://github.com/alphagov/govuk-infrastructure/issues/3718 for details.

This PR to migrate a different bucket may also be useful https://github.com/alphagov/govuk-infrastructure/pull/3848 

### Acceptance Criteria

- [ ] The bucket named in the title is using the shared module

PR #3932, opened 2026-03-17
Migrate DB Backup S3 Buckets to new Module

dj-maisy:
## What?
This migrates the DB Backup and Replica buckets to use the new S3 Module. This was a bit more work than I expected and also required some changes to the module itself to support multiple storage type transitions

### Related Work

* Closes https://github.com/alphagov/govuk-infrastructure/issues/3754

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/db_backup_s3.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/db_backup_s3.tf
@@ -102,7 +102,7 @@ moved {
 }
 
 moved {
-  from = aws_s3_bucket_object_lock_configuration.backup_main
+  from = aws_s3_bucket_object_lock_configuration.backup_main[0]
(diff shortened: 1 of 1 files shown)


terraform/deployments/govuk-publishing-infrastructure/db_backup_s3.tf
```diff
@@ -3,142 +3,163 @@ locals {
...
+    },
+    {
+      id                            = "non-production"
+      status                        = var.govuk_environment != "production" ? "Enabled" : "Disabled"
+      expiration                    = { days = 8 }
+      noncurrent_version_expiration = { noncurrent_days = 1 }
+    }
+  ]
+
+  object_lock_config = var.govuk_environment != "production" ? [
```
jfharden: I think you have this the wrong way round right? If it's prod you do want the compliance lock

PR #3945, opened 2026-03-18
CDN analytics review IAM

Nyzl:
This PR adds read access for sthe service account data-processing@gds-bq-processing.iam.gserviceaccount.com, used to process Fastly logs for analytics.

Access has been removed for the Analytics Team because it is no longer needed


JonathanHallam (approved):
Looks good!

Issue #3913, opened 2026-03-13
Test out CloudWatch Network Monitoring


AP-Hunt:
**User Need**

**As a** platform engineer
**I want** configure and deploy CloudWatch Network Monitoring for EKS
**so that** I can learn more about it
**and** evaluate it for fitness

---

### Context

In https://github.com/alphagov/govuk-platform-internal/issues/104 ([write up](https://docs.google.com/document/d/1rfmltSzV9ZeVoY30_6o3a620OuIEClvb0pyMoXTiqkg/edit?pli=1&tab=t.0)) we looked across the Kubernetes ecosystem for options for automatically generating a Kubernetes service map using live data.

We landed on a couple of options to try out, CloudWatch Network Monitoring being one of them (the other option is Grafana, covered in [this separate issue](https://github.com/alphagov/govuk-infrastructure/issues/3914))

---

### What’s Needed

List anything the solution must do or be (behaviour, performance, security, UX, etc.).

- [x] Learn about CloudWatch Network Monitoring as described in the write up
- [x] Deploy it into the integration environment _only_ so we can get a true picture of it

---

### Acceptance Criteria

Clear, measurable conditions to verify the story is complete and valuable.

- [ ] Have an opinion about the fitness of CloudWatch Network Monitoring as a solution for generating Kubernetes service maps
- [ ] Have demonstrated and explained it to the team
- [ ] Estimate what it will cost each month to run, and how much it might cost if we doubled in size (so we can estimate whether the costs might get out of control in the future)

dj-maisy:
PR:

* https://github.com/alphagov/govuk-infrastructure/pull/3971

dj-maisy:
Another PR:

* https://github.com/alphagov/govuk-infrastructure/pull/3979

dj-maisy:
This is now live in Integration but currently requires the fulladmin role to access until the Role Policies are updated (see above).

dj-maisy:
This task is **on hold** while we gather data.

dj-maisy:
## Cost Estimations

* Network Flow monitoring is charged per EKS Node, at approximately $5.037 per month.
* We have 8 Nodes in Production (at time of writing).
* The Network Flow Monitoring produces one set of 5 Metrics for the whole EKS Cluster billed at approximately $1.50 per month.
* Total Estimated Cost for leaving this on in Production: $40.30 per month.
* Ignoring any free tiers, that's **$41.80 per month**.
[...]

PR #3965, opened 2026-03-23
Tidy-up Kube State Metrics Config (Make it Default)

dj-maisy:
## What?
~~This tweaks our Terraform to allow us to enable the Network Flow Monitoring Agent per environment. To make the config a little cleaner, I am also making the Kube State Metrics Addon a default for the clusters (and supplying a Prometheus Dashboard in a separate PR to help expose these metrics) as it gives us usable health metrics with no noticeable drawbacks.~~ Breaking this up into two PRs.

This tidies-up our Kube State Metrics configuration (before we start configuring the Cluster Addons to enable the Network Flow Monitor in integration).

### Related

* https://github.com/alphagov/govuk-infrastructure/issues/3913

### DIFF
--- a/terraform/deployments/cluster-infrastructure/main.tf
+++ b/terraform/deployments/cluster-infrastructure/main.tf
@@ -40,8 +40,9 @@ locals {
   }
 
   default_cluster_addons = {
-    coredns        = { addon_version = "v1.13.2-eksbuild.1", resolve_conflicts_on_create = "OVERWRITE" }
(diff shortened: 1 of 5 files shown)


AP-Hunt:
I think you should split this PR in two: one for the general improvements, and one for the addition of the network flow add-on. There's a 50% chance we'll not go with CloudFront, and it'd make it easier to unpick without losing the general improvements.

dj-maisy:
> I think you should split this PR in two: one for the general improvements, and one for the addition of the network flow add-on. There's a 50% chance we'll not go with CloudFront, and it'd make it easier to unpick without losing the general improvements.

Done.

PR #3969, opened 2026-03-23
Enable Network Flow Addon (Integration)

dj-maisy:
## What?

This enables the Network Flow Agent on Integration.

### Related

* https://github.com/alphagov/govuk-infrastructure/issues/3913

### DIFF
--- a/terraform/variables/integration/common.tfvars
+++ b/terraform/variables/integration/common.tfvars
@@ -55,8 +55,6 @@ enable_arm_workers_green = true
 enable_tetragon          = true
 enable_x86_workers       = false
 
-enable_network_flow_addon = true
-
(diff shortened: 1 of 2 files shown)


terraform/variables/integration/common.tfvars
```diff
@@ -55,6 +55,8 @@ enable_arm_workers_green = true
 enable_tetragon          = true
 enable_x86_workers       = false
 
+enable_network_flow_addon = true
```
AP-Hunt: This should be in [cluster-infrastructure.tfvars](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/variables/integration/cluster-infrastructure.tfvars) since it's defined in its `variables.tf`

PR #3973, opened 2026-03-24
IA-2582: setup user table and scheduled query

aaronfowles:
This is PR 2 out of 3 for https://gov-uk.atlassian.net/browse/IA-2582. This is the mechanism for determining whether there has been unauthorised access.

- add a configurable allow-list authorised_users
- add scheduled query to identify new rows in audit logs from unauthorised users
- persists record of unauthorised access in unauthorised_access_alerts table
- add I&A team alert channel secret container (to be configured later)

also a bit of housekeeping:
- renaming `main.tf` -> `logging.tf`
- moving dataset definition from `main.tf` -> `bigquery.tf`

The next PR will configure the actual alerting.

### DIFF
--- a/terraform/deployments/gcp-ga4-aggregate-analytics/modules/data-access-monitoring/README.md
+++ b/terraform/deployments/gcp-ga4-aggregate-analytics/modules/data-access-monitoring/README.md
@@ -3,7 +3,7 @@ This module sets up a dataset called `data_access_log` in the specified project 
 
 A configurable allow-list table also exists in that dataset called `authorised_users`. Any users which have queried a table in the project who are not also in `authorised_users` will result in that read being captured in the table `unauthorised_access_alerts`.
 
-It could eventually move to `terraform/shared-modules` once more mature and tested.
+It could eventually move to `terraform/shared-modules` once more mature and tested. The required Google Cloud API services would need to be managed directly by module in this case.
 <!-- BEGIN_TF_DOCS -->
 ## Requirements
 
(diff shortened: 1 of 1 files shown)


terraform/deployments/gcp-ga4-aggregate-analytics/modules/data-access-monitoring/bigquery.tf
```diff
@@ -0,0 +1,89 @@
...
+  location   = "europe-west2"
+}
+
+# The "Allow List" Table
+resource "google_bigquery_table" "authorised_users" {
+  project    = var.project_id
+  dataset_id = google_bigquery_dataset.audit_logs.dataset_id
+  table_id   = "authorised_users"
+
+  schema = <<EOF
```
AP-Hunt, line 13: You might have a nicer experience using [the jsonencode function](https://developer.hashicorp.com/terraform/language/functions/jsonencode). Up to you though. Both are readable.

terraform/deployments/gcp-ga4-aggregate-analytics/modules/data-access-monitoring/bigquery.tf
```diff
@@ -0,0 +1,89 @@
...
+  project        = var.project_id
+  display_name   = "Detect Unauthorised BQ Reads"
+  location       = "europe-west2"
+  data_source_id = "scheduled_query"
+  schedule       = "every 5 minutes"
+
+  service_account_name = google_service_account.query_executor.email
+
+  params = {
+    query = <<EOT
```
AP-Hunt, line 59: You could use [the templatefile function](https://developer.hashicorp.com/terraform/language/functions/templatefile) here to separate the query into its own `.sql` file and get more support from your IDE

terraform/deployments/gcp-ga4-aggregate-analytics/modules/data-access-monitoring/bigquery.tf
aaronfowles: Good shout - I'll do this in the final PR for this work

terraform/deployments/gcp-ga4-aggregate-analytics/modules/data-access-monitoring/bigquery.tf
aaronfowles: I'll also do this in the final PR for this work

PR #3998, opened 2026-03-26
Move gov graph code

aaronfowles:
This adds all of the terraform code from govuk-knowledge-graph over to govuk-infrastructure and terrraform cloud, huzah!


AP-Hunt:
I worked with @aaronfowles and @JonathanHallam to move their existing state file across. That bit doesn't need doing.

This PR brings this repo in line with reality.

PR #4000, opened 2026-03-26
IA-2582: fix issue with filter

aaronfowles:
I'm having issues configuring the filter for a metric based alert. I've found code examples using `resource.label` and `resource.labels` interchangeably. Also `metric.label` and `metric.labels`.

It's also possible that you can use them interchangably and I just got the `metric.type` wrong several times 😞 .

These [docs](https://docs.cloud.google.com/monitoring/api/v3/filters) suggest it's always the plural so I'm going with that in this PR. @oliverroberts1-gds would you mind taking a look as I know you've wrestled with something similar on faas?

### DIFF
--- a/terraform/deployments/gcp-ga4-aggregate-analytics/modules/data-access-monitoring/monitoring.tf
+++ b/terraform/deployments/gcp-ga4-aggregate-analytics/modules/data-access-monitoring/monitoring.tf
@@ -23,7 +23,7 @@ resource "google_monitoring_alert_policy" "dts_failure_alert" {
   conditions {
     display_name = "Detection Query Failed"
     condition_threshold {
-      filter          = "resource.type=\"bigquery_dts_config\" AND resource.labels.config_id=\"${local.detection_query_uuid}\" AND metric.type=\"bigquerydatatransfer.googleapis.com/transfer/completed_runs\" AND metric.label.completion_status=\"FAILED\""
+      filter          = "resource.type=\"bigquery_dts_config\" AND metric.type=\"bigquerydatatransfer.googleapis.com/transfer_config/completed_runs\""
       duration        = "0s"
       comparison      = "COMPARISON_GT"
       threshold_value = 0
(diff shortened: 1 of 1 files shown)


terraform/deployments/gcp-ga4-aggregate-analytics/modules/data-access-monitoring/monitoring.tf
```diff
@@ -23,7 +23,7 @@ resource "google_monitoring_alert_policy" "dts_failure_alert" {
   conditions {
     display_name = "Detection Query Failed"
     condition_threshold {
-      filter          = "resource.type=\"bigquery_dts_config\" AND resource.labels.config_id=\"${local.detection_query_uuid}\" AND metric.type=\"bigquerydatatransfer.googleapis.com/transfer/completed_runs\" AND metric.label.completion_status=\"FAILED\""
+      filter          = "resource.type=\"bigquery_dts_config\" AND resource.labels.config_id=\"${local.detection_query_uuid}\" AND metric.type=\"bigquerydatatransfer.googleapis.com/transfer_config/completed_runs\" AND metric.labels.completion_state=\"FAILED\""
```
oliverroberts1-gds: I used the label link as shown in the [faas logging](https://github.com/alphagov/govuk-feedback-processing/blob/532f333ebe97eea1827f5c87c7fa9d6ebd825540/terraform-dev/logging.tf#L83)

oliverroberts1-gds: I believe you also have to specify the "resource.type=\"global\"  for custom metrics, as they can be cross project

oliverroberts1-gds:
I think a part of the issue may be that you are referring to the metric before you have any instances of it (hence I went through the faff of defining the metric in terraform first). Can you see instances of the metric in your logging/monitoing gui?

aaronfowles:
> I think a part of the issue may be that you are referring to the metric before you have any instances of it (hence I went through the faff of defining the metric in terraform first). Can you see instances of the metric in your logging/monitoing gui?

I think this is the equivalent in the UI now. part of the problem in last merge was definitely that I misspelled the name of the metric 🙄 . So I think the metric does exist as it's a "built-in" from the BigQuery DTS.

image

PR #4006, opened 2026-03-30
IA-2582: refine filters on ga4-aggregate-analytics alert to be more selective

aaronfowles:
This PR:

- adds the docs link in the alert content, otherwise it's difficult to find in the UI.
- changes the metric alignment period to be hourly to match the schedule
- add the labels back in to the metric definition to be more selective now that the basic metric is firing

### DIFF
--- a/terraform/deployments/gcp-ga4-aggregate-analytics/modules/data-access-monitoring/monitoring.tf
+++ b/terraform/deployments/gcp-ga4-aggregate-analytics/modules/data-access-monitoring/monitoring.tf
@@ -23,7 +23,7 @@ resource "google_monitoring_alert_policy" "dts_failure_alert" {
   conditions {
     display_name = "Detection Query Failed"
     condition_threshold {
-      filter          = "resource.type=\"bigquery_dts_config\" AND metric.type=\"bigquerydatatransfer.googleapis.com/transfer_config/completed_runs\""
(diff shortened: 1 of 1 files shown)


terraform/deployments/gcp-ga4-aggregate-analytics/modules/data-access-monitoring/monitoring.tf
```diff
@@ -23,7 +23,7 @@ resource "google_monitoring_alert_policy" "dts_failure_alert" {
   conditions {
     display_name = "Detection Query Failed"
     condition_threshold {
-      filter          = "resource.type=\"bigquery_dts_config\" AND metric.type=\"bigquerydatatransfer.googleapis.com/transfer_config/completed_runs\""
+      filter          = "resource.type=\"bigquery_dts_config\" AND resource.labels.config_id=\"${local.detection_query_uuid}\" AND metric.type=\"bigquerydatatransfer.googleapis.com/transfer_config/completed_runs\" AND metric.labels.completion_status=\"FAILED\""
       duration        = "0s"
```
oliverroberts1-gds, line 27: This look fab looks fab to me, I thought that you needed global but I guess that is a custom metric thing.

aaronfowles: I might have to eat crow on the `global` thing. I _think_ that might only be for custom metrics but I might find out post-merge that I'm wrong 🙈

PR #4061, opened 2026-04-09
Remove things

JonathanHallam:
There were a few things kicking about in the terraform that no longer needed to be there, namely some already imported imports and some old or unnecessary variables that were throwing warnings up


Nyzl (approved):
excellent cleanup

PR #4065, opened 2026-04-13
Use GCP resource for freshness control

emmalowe:
When the site search serving configurations were first defined in Terraform, Google resources to define controls didn't exist so these were written using restapi objects.

Now that a Google resource does exist for controls, we can switch to using that to simplify the code and so that we can benefit from better validation and type management.


emmalowe:
Unfortunately, the resource `google_discovery_engine_control` doesn't seem to support multiple control points at this point in time, so it would be difficult to set up the freshness boosting using it right now.

PR #4066, opened 2026-04-13
Use float for freshness boost

emmalowe:
An attempt to avoid drift in the Terraform configuration, caused by explicitly setting the boostAmount to 0 in Terraform which is converted to a null response in the Discovery Engine API. We need the set boostAmount to match what's in the API response to avoid drift.


stringiest (approved):
Fingers crossed!

jaskaransarkaria (approved):

PR #4071, opened 2026-04-15
Introduce 10 day cooldown period for Dependabot and Renovate

AP-Hunt:
## What

Adds 10 day dependency cooldowns for Dependabot and Renovate as appropriate

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,79 +1,86 @@
 {
-  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": [
-    "config:recommended",
-    "helpers:pinGitHubActionDigests"
-  ],
-  "platformAutomerge": true,
-  "pre-commit": {
-    "enabled": true
-  },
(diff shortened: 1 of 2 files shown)


renovate.json
```diff
@@ -1,79 +1,86 @@
...
-        "aws-eks-addon"
-      ]
-    },
-    {
-      "matchManagers": ["custom.regex"],
-      "matchDepNames": [
-        "Terraform Version"
-      ],
-      "groupName": "Terraform Version"
+    "$schema": "https:\/\/docs.renovatebot.com\/renovate-schema.json",
```
samsimpson1: This URL has been mangled

AP-Hunt: Ugh! PHP! That's probably true on all the Renovate ones :( I'll fix it here.

AP-Hunt: It's actually done it to ALL the slashes!

samsimpson1 (approved):

PR #4079, opened 2026-04-17
Test removing token param from tfc-configuration

jfharden:
I want to test if we really need to provide TFE provider auth credentials when running in terraform cloud. This should prove it by removing the token configuration


jfharden:
We do. https://app.terraform.io/app/govuk/workspaces/tfc-configuration/runs/run-NTVx8fp7yCMFDfEU is the run for this and it's auth error-tastic

PR #4088, opened 2026-04-20
Set blocked_encryption_types in S3 SSE config

samsimpson1:
The TF provider wants to make perpetual changes if this isn't set


jfharden (approved):
Interesting (unrelated) change in the csp reporter workspaces 🤔

PR #4089, opened 2026-04-20
IA-2701: add whitehall tables to data freshness alerts

aaronfowles:
Configure GovGraph to alert if data from `whitehall` source is late.


JonathanHallam (approved):
👍

PR #4091, opened 2026-04-21
Create a new IAM role for govuk-ai-graph-tools

AP-Hunt:
## What
The role is identical in permissions to `govuk-ai-accelerator`, but is separate so that the two things don't share a role. 

## How to review

I cloned `ai_accelerator_iam.tf` as a starting point. Make sure I didn't miss changing any references.


jfharden (approved):
Approved even though I think we shouldn't be putting this in the publishing infra workspace. We're adding extra complexity to an already too-big deployment, and for things that will never go beyond integration

PR #4093, opened 2026-04-21
Add content block embed code highlighter repo

patrickjfl:
This is a library that will allow content block embed codes to be highlighted within publishing apps.


jfharden (approved):

patrickjfl:
Closing because we've decided it makes more sense to keep this work within the existing `content-block-editor` repo.

PR #4119, opened 2026-04-24
IA-2640: add gcp-project-init module

aaronfowles:
First-pass at a modular approach to allow existing GCP projects to be configured with sensible defaults and remove some boilerplate. It will also update the `roles/owner` role to have at least the terraform SA configured and keep all project-level roles explicitly managed by terraform.

A few things to consider:

- is it clear that the terraform account will be added as owner by default?
- I've opted for bindings for project-level permissions as that's what the team currently uses. Adding the terraform SA as a member would be more straight-forward but could get dropped if a binding then replaces it. It feels cleaner to enforce all project level roles must be defined in this module IMO.
- turns out imports are not allowed in modules.
- I've added the users who are currently configured at the project-level. Should we create Google Groups for them now or come to that later?

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -12,3 +12,4 @@
 /terraform/deployments/gcp-gds-bq-processing/ @alphagov/gov-uk-platform-engineering @alphagov/govuk-data-engineering
 /terraform/deployments/gcp-gov-graph/ @alphagov/gov-uk-platform-engineering @alphagov/govuk-data-engineering
 /terraform/variables/**/gcp-gov-graph.tfvars @alphagov/gov-uk-platform-engineering @alphagov/govuk-data-engineering
+/terraform/shared-modules/gcp-project-init/ @alphagov/gov-uk-platform-engineering @alphagov/govuk-data-engineering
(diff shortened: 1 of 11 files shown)


aaronfowles (commented):
- use Google Groups
- check readme is up to scratch - AC: should be usable by another team
- update CODEOWNERS

Nyzl (approved):
LGTM - good design choices that make sense and with default values additional business units should be able to make use of the code as is, or adapt it.
Our aim will then be to roll this out to all GCP projects

jfharden (commented):
This looks great! The only extra thing is you should add the shared module to the trigger patterns of your workspace:

https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/tfc-configuration/gcp-projects.tf#L105

This will mean the workspace does a plan (which will report its status to a PR) whenever a PR opens touching that path. It's especially important for shared modules so people don't unintentionally make changes which negatively impact some of the other things sharing the module. The big benefit though is you get to see the plans when you actively want to make module changes :)

aaronfowles:
> This looks great! The only extra thing is you should add the shared module to the trigger patterns of your workspace:
> 
> https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/tfc-configuration/gcp-projects.tf#L105
> 
> This will mean the workspace does a plan (which will report its status to a PR) whenever a PR opens touching that path. It's especially important for shared modules so people don't unintentionally make changes which negatively impact some of the other things sharing the module. The big benefit though is you get to see the plans when you actively want to make module changes :)

Ah good shout! I definitely didn't think about that. I'll make those changes.

jfharden (approved):

PR #4120, opened 2026-04-27
GOV.UK Chat: add Cloudwatch alarms for new models

jackbot:
We're using 2 new models (Haiku 4.5 and Sonnet 4.5) in GOV.UK Chat, so
we want to configure Cloudwatch alarms for their usage.

We currently use 3 models on Chat, and each model has 2 alarms
configured (one for 50% rate limit usage and one for 100% usage).

We're now supporting 2 more models - which means 4 more alarms.

This code previously defined quite a long resource for each individual
alarm and would have got pretty unwieldy soon if we added 4 more.

So the first commit here refactors the code to use variables and a single
`resource` block. The second commit adds the new alarms for the new models.

### DIFF
--- a/terraform/deployments/chat/cloudwatch_alarms.tf
+++ b/terraform/deployments/chat/cloudwatch_alarms.tf
@@ -7,38 +7,38 @@ locals {
     claude_sonnet = {
       model_id    = "eu.anthropic.claude-sonnet-4-20250514-v1:0"
       token_limit = var.chat_token_limits_per_minute["claude_sonnet"]
-      expression  = "((m1 + m2 + (m3 * 5)) / TOKEN_LIMIT) * 100"
+      expression  = "((CacheWriteInputTokenCount + InputTokenCount + (OutputTokenCount * 5)) / TOKEN_LIMIT) * 100"
       sns_topic   = aws_sns_topic.chat_alerts_dublin.arn
     }
     claude_sonnet_4_5 = {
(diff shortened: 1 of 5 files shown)


terraform/deployments/chat/cloudwatch_alarms.tf
```diff
@@ -10,6 +10,18 @@ locals {
       expression  = "((m1 + m2 + (m3 * 5)) / TOKEN_LIMIT) * 100"
       sns_topic   = aws_sns_topic.chat_alerts_dublin.arn
     }
+    claude_sonnet_4_5 = {
+      model_id    = "eu.anthropic.claude-sonnet-4-5-20250929-v1:0"
+      token_limit = var.chat_token_limits_per_minute["claude_sonnet_4_5"]
```
AP-Hunt, line 15: In `variables.tf` you could change the type of `chat_token_limits_per_minute` to be [an object with a field for each model](https://developer.hashicorp.com/terraform/language/expressions/type-constraints#object). You'd get full support for the keys that are present on that object that way, and wouldn't be able to encounter a whole class of errors.

AP-Hunt (approved):
I'm approving on the grounds that the code looks technically sound and correct, but I would suggest taking a look at the change in variable type I've suggested.

terraform/deployments/chat/cloudwatch_alarms.tf
jackbot: Thanks @AP-Hunt, that makes it all much nicer; I didn't like the default values. Thanks for the review!

PR #4122, opened 2026-04-28
GOV.UK Chat: fix CloudWatch metric names

jackbot:
According to the CloudWatch docs:

> The valid characters are letters, numbers, and underscore. The first character must be a lowercase letter.

This fixes the metric names here so they start with a lowercase letter

See https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_MetricDataQuery.html#API_MetricDataQuery_Contents


dj-maisy (approved):
LGTM. 👍

PR #4148, opened 2026-04-30
pinning Google provider version

Nyzl:
Gov-Graph requires a Google provider version of 6.27.0 to support it's use of VM deployment via https://github.com/terraform-google-modules/terraform-google-container-vm

The longer term plan is to move off VM dependancy for data ingestion, but this reversion is needed to as current TF plans are failing


aaronfowles (approved):
LGTM! Let me know if you need a merge 👍

PR #4156, opened 2026-05-05
Rename alert destination following team name change

AgaDufrat:
The team is focusing on APIs so the name was changed to reflect that.


nicholsj (approved):
LGTM thanks!

PR #4181, opened 2026-05-08
App startups

JonathanHallam:
In a previous commit we added a startup script to asset manager to allow us to set the log driver so logs could be piped through from docker containers to google cloud logging. The good news is that it worked! This generifies the approach and does the same thing for the other apps


jfharden (approved):

aaronfowles (approved):
LGTM!

PR #4182, opened 2026-05-08
Remove infrastructure for queue search_api_to_be_indexed

emmalowe:
This queue was used as part of the process to update links in documents in the government index in elasticsearch. Now all formats have been migrated to the govuk index, and the government index is being decommissioned, we no longer need this queue. (Documents in the govuk index are update via different queues.)

The queue consumers were removed in search-api already [1]. This change is to remove the queue and related infrastructure.

[1] https://github.com/alphagov/search-api/pull/3589


samsimpson1 (approved):

koetsier (approved):
lgtm

PR #4192, opened 2026-05-12
Improve ephemeral cluster shutdown documentation

jasonBirchall:
## Summary

- Adds a prerequisites checklist for the shutdown process
- Documents what the shutdown script does at each stage (Argo apps, Helm charts, TFC destroy order)
- Explains how the script authenticates to Terraform Cloud
- Adds troubleshooting sections for three common failure modes:
  - MFA session expiring mid-run
  - Cluster API server already gone (partial destroy)
  - Pending or errored Terraform Cloud runs
- Documents the manual workspace destruction path with dependency order
- Reorders README sections to follow the cluster lifecycle (create → operate → destroy)
- Fixes a stray backtick in the Basic Usage section

## Why

Ran into all three failure modes while shutting down an ephemeral cluster
and the existing docs didn't cover recovery. The shutdown section was also
positioned between creation and operation, which was confusing.

### DIFF
--- a/terraform/deployments/ephemeral/README.md
+++ b/terraform/deployments/ephemeral/README.md
@@ -8,7 +8,7 @@ This module is used to provision ephemeral EKS clusters via Terraform Cloud.
 1. Do a `terraform init`
 1. Run an apply with your chosen ephemeral cluster ID (this isn't generated for you):
    ```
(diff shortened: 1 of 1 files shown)


jfharden (approved):
❤️ Love this huge improvement

PR #4195, opened 2026-05-12
Sch 2050 update documentation for new vertex name

emmalowe:
Changes references from "Vertex" to "Discovery Engine", since Google have rebranded "Vertex AI Search" to "Agent Search". Here we're using the API name "Discovery Engine" instead, because it's more stable and a more familiar name for developers.

At the same time, I've added some links to help with documentation and corrected some comments related to GA4 events.

I haven't touched anything to do with automated evaluations, since this is due to be retired. See https://gov-uk.atlassian.net/browse/SCH-490

Jira ticket (for this PR): https://gov-uk.atlassian.net/browse/SCH-2050

### DIFF
--- a/terraform/deployments/search-api-v2/events_ingestion.tf
+++ b/terraform/deployments/search-api-v2/events_ingestion.tf
@@ -47,7 +47,7 @@ resource "google_bigquery_table" "search_event" {
 
 }
 
-# ga4 'select_item' events get transformed and inserted into this time-partitioned search-event table defined with a Discovery Engine schema
+# ga4 'page_view' events get transformed and inserted into this time-partitioned search-event table defined with a Discovery Engine schema
 resource "google_bigquery_table" "view_item_event" {
   dataset_id          = google_bigquery_dataset.dataset.dataset_id
   table_id            = "view-item-event"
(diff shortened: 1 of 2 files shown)


jfharden (approved):

terraform/deployments/search-api-v2/events_ingestion.tf
```diff
@@ -47,7 +48,8 @@ resource "google_bigquery_table" "search_event" {
 
 }
 
-# ga4 'select_item' events get transformed and inserted into this time-partitioned search-event table defined with a vertex schema
+# ga4 'page_view' events get transformed and inserted into this time-partitioned view-item-event table defined with a
```
stringiest, line 51: Ooh this is interesting! So we're using `page_view` rather than `select_item` in the dataform queries for everything apart from external links. I guess that makes sense as otherwise pages might not appear in the results in the first place to get clicked on 🤔

stringiest (approved):
This is great! The extra comments really help with understanding the event tracking.

terraform/deployments/search-api-v2/events_ingestion.tf
emmalowe: I thought it was interesting as well! From what I guesstimate, for external links we need to use "select_item" because we're not tracking views of external links (since they're not part of gov.uk) but using "page_view" for internal links gives us guaranteed additional information, such as content_id. For searches, we use "view_item_list".

PR #4198, opened 2026-05-13
Add discoveryengine.documents.get/delete to search_admin role

koetsier:
This is needed so that the search-admin get can get and delete documents from Google Vertex

It allows us to run the rake tasks documented in
https://docs.publishing.service.gov.uk/manual/get-or-delete-a-document-from-VAIS.html


hannako (approved):

nimalank7 (approved):
LGTM!

PR #4207, opened 2026-05-14
Add a CKAN EFS

kentsanggds:
This will allow cronjobs to create outputs which will be used by other pods to make them available for viewing

https://cddodatamarketplace.atlassian.net/jira/software/c/projects/DGUK/boards/527?selectedIssue=DGUK-527

### DIFF
--- a/terraform/deployments/datagovuk-infrastructure/ckan_efs.tf
+++ b/terraform/deployments/datagovuk-infrastructure/ckan_efs.tf
@@ -1,0 +1,46 @@
+data "aws_route53_zone" "internal" {
+  name = "${var.govuk_environment}.govuk-internal.digital"
+  tags = {
+    repository = "govuk-infrastructure"
+  }
+  private_zone = true
+}
+
+resource "aws_security_group" "ckan_efs" {
+  name        = "ckan-efs"
+  vpc_id      = data.tfe_outputs.vpc.nonsensitive_values.id
+  description = "Security group for ckan EFS volume"
+}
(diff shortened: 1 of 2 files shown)


terraform/deployments/datagovuk-infrastructure/ckan_efs.tf
```diff
@@ -0,0 +1,46 @@
...
+resource "aws_efs_mount_target" "ckan_efs" {
+  for_each        = toset(data.tfe_outputs.cluster_infrastructure.nonsensitive_values.private_subnets)
+  file_system_id  = aws_efs_file_system.ckan_efs.id
+  subnet_id       = each.key
+  security_groups = [aws_security_group.ckan_efs.id]
+}
+
+resource "aws_route53_record" "ckan_efs" {
+  zone_id = data.aws_route53_zone.internal.zone_id
+  name    = "assets.${data.aws_route53_zone.internal.name}"
```
AP-Hunt: I think this name is going to conflict. `assets.production.govuk-internal.digital` exists in the private `production.govuk-internal.digital`zone, and you're adding this to the public one. 

I think you probably want to be adding this to the private zone (`private_zone = true` in the lookup), so you'd be best off making this a unique-to-DGU name

kentsanggds: yes, I'll update that, good spot!

AP-Hunt (approved):

PR #4218, opened 2026-05-18
Migrate Tempo Bucket to New S3 Module

### LINKED ISSUE #3752: Migrate terraform/deployments/cluster-services/tempo.tf to S3 Shared Module

> [!CAUTION]

### Acceptance Criteria

dj-maisy:
## What?
This migrates the Tempo S3 Bucket to use the shared module.

### Related

* Resolves https://github.com/alphagov/govuk-infrastructure/issues/3752


terraform/deployments/cluster-services/tempo.tf
```diff
@@ -4,36 +4,28 @@ locals {
...
-      variable = "aws:SecureTransport"
-    }
-  }
+moved {
+  from = aws_s3_bucket_policy.tempo_bucket_policy
+  to   = module.secure_s3_bucket_tempo.aws_s3_bucket_policy.bucket_policy
 }
 
 module "tempo_iam_role" {
+  depends_on = [module.secure_s3_bucket_tempo]
```
jfharden: You shouldn't need this explicit depends on (or the other) because this references aws_iam_policy.tempo.arn which in turn references data.aws_iam_policy_document.tempo.json which itself references the module, so terraform should build a correct dependency tree

terraform/deployments/cluster-services/tempo.tf
```diff
@@ -52,12 +44,15 @@ module "tempo_iam_role" {
 
 data "aws_iam_policy_document" "tempo" {
   statement {
+    sid       = "TempoAccessDataSources"
     actions   = ["s3:ListBucket", "s3:?*Object", "s3:?*ObjectTagging"]
-    resources = [aws_s3_bucket.tempo.arn, "${aws_s3_bucket.tempo.arn}/*"]
+    resources = ["${module.secure_s3_bucket_tempo.arn}", "${module.secure_s3_bucket_tempo.arn}/*"]
```
AP-Hunt: The first array element doesn't need to be interpolated

AP-Hunt (approved):
Approving on the grounds that the code changes look good and the Terraform plans are doing what you'd expect (and those that are doing more are doing it for reasons I know about).

The one comment I've left isn't a deal breaker, just a nice-to-fix. We should, however, sort out the commit history before merging. It's all gone a bit Pete Tong in there I think.

PR #4230, opened 2026-05-21
Add output bucket to store check links reports

kentsanggds:
Output bucket for the check-links reports but might also be used for other outputs.

https://cddodatamarketplace.atlassian.net/jira/software/c/projects/DGUK/boards/527?selectedIssue=DGUK-527


AP-Hunt (approved):
On the condition that a future commit moves to using IRSA to give the necessary pods an identity, and that identity is substituted for the node roles here.

PR #4233, opened 2026-05-21
Add required_status_checks and format to govuk-social-data

oliverroberts1-gds:
Added required_status_checks to govuk-social-data with additional_contexts as an empty array.

### DIFF
--- a/terraform/deployments/github/repos.yml
+++ b/terraform/deployments/github/repos.yml
@@ -471,11 +471,13 @@ repos:
   govuk-social-data:
     can_be_deployed: false
     visibility: "private"
-    standard_contexts: *standard_security_checks
     push_allowances: ["alphagov/gov-uk-data", "alphagov/gov-uk-production-admin", "alphagov/gov-uk-production-deploy"]
     required_pull_request_reviews:
       pull_request_bypassers:
         - "alphagov/gov-uk-data"
+    required_status_checks:
+      # standard security checks are disabled as not configured for a private repo
+      additional_contexts: []
 
   govuk_ab_testing:
     publishes_gem: true


terraform/deployments/github/import.tf
```diff
@@ -0,0 +1,4 @@
+import {
```
AP-Hunt: This is unnecessary. This repo is already managed by Terraform :+1:

AP-Hunt (changes requested):

terraform/deployments/github/import.tf
oliverroberts1-gds: Ah, thank you for that info. I'll remove it.

Just to check my understanding, If I were to add a folder https://github.com/alphagov/govuk-infrastructure/tree/main/terraform/deployments/govuk-social-data/ terraform cloud would then create the workspace automatically

oliverroberts1-gds: @AP-Hunt Done

oliverroberts1-gds: Ah I see in the documentation I have to approve the plan after merge

AP-Hunt: No, there's additional configuration you'll need for dedicated Terraform workspaces. See for example https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/tfc-configuration/chat.tf

terraform/deployments/github/import.tf
AP-Hunt, line 1: It'd be better to just rebase and drop the commit than add a second one reverting it

oliverroberts1-gds: Very true, I was being lazy

AP-Hunt (approved):

PR #4244, opened 2026-05-26
Use separate resources to manage github vulnerability alerts and dependabot security updates

jfharden:
vulnerability_alerts is a deprecated attribute


AP-Hunt (approved):
This looks like how I'd do it and the plan doesn't obviously do anything stupid

PR #4305, opened 2026-05-29
Update opensearch shared module

jfharden:
* As our agreed pattern change, alter the ai-accelerator opensearch module to source via filepath not git ref
* Add features to the opensearch-blue-green-deployment module to prepare for migrating the elasticsearch-green search-api cluster to use it:
    * Allow not enabling advanced_security_options
    * Allow prefixing the "colour" instead of suffixing (e.g. `green-elasticsearch-domain` instead of `elasticsearch-domain-green`). Note I've only done this for green, and when they next launch blue (soon) we should _not_ set this, and then remove the option to mean in future all are suffixed.
    * Allow setting provisioned IOPS in the ebs options

This should result in No changes to the existing ai-accelerator deployment (which already triggers on the shared module code), however a terraform quirk means the entire advanced_security_options attribute is now being treated as sensitive, whereas previously the `enabled` property was seen as nonsensitive, this means there _is_ a change, but that change is merely to change the value `enable = true` to be sensitive 

```
update
enabled :
true
change to
Sensitive value:
```

Screenshot 2026-05-29 at 17 19 49

So this isn't actually going to change anything, and you can see the fact the block exists and everything else is unchanged, and enabled is hardcoded to true, means there really isn't a change


jfharden:
Rebased onto main

samsimpson1 (approved):

PR #4306, opened 2026-06-01
Allow disabling bucket logging

jfharden:
* Allow disabling bucket logging for buckets in ephemeral clusters, and for the logging bucket itself
* Add a variable to cluster-services to allow the Tempo bucket to have logging disabled
* Disable the Tempo bucket logging for ephemeral clusters

Note: I've purposefully chosen the inverse "disable" case for the variable name, since it sends a clearer message that you have to opt into something bad.

Really we need to update the test account to have a logging bucket, we already have the work planned to migrate it and I've added cleanup of this into the [Issue for migrating the logging bucket to the "new" module](https://github.com/alphagov/govuk-infrastructure/issues/3888)


AP-Hunt (approved):
The validation in the S3 module is a very good idea :+1:

PR #4312, opened 2026-06-02
Mark govuk-rota-generator as private [TECH-29]

ChrisBAshton:
Set the visibility of govuk-rota-generator to private.

Jira: https://gov-uk.atlassian.net/browse/TECH-29


AgaDufrat (approved):
I can't see the visibility change in the plan but otherwise looks fine. It maybe the case that we have to run TF plan/apply twice. 

Screenshot 2026-06-02 at 12 49 18

dj-maisy (approved):

PR #4314, opened 2026-06-02
Address the broken ECR Deployment, move old repos.

dj-maisy:
## What?
Some Github Repos have been marked as archived or have had their deployable status set to false. This now means the TF is trying to delete the already-created ECR repos which are protected by a prevent_destroy lifecycle rule.

Hopefully this change will allow us to "move" the state of the affected repos so we can then just delete them in a follow-up PR.


jfharden (approved):
An unusual but creative solution :) Love it!

PR #4318, opened 2026-06-03
Make Replatform Test App Deployable for testing.

dj-maisy:
## What?
This sets can_be_deployed to "true" so I can deploy the Replatform Test App into Integration for testing Kyverno and Cosign.

My IDE also auto-formatted the `push_allowances` arrays, which I don't disagree with as it seems to look cleaner.

### DIFF
--- a/terraform/deployments/github/repos.yml
+++ b/terraform/deployments/github/repos.yml
@@ -308,11 +308,9 @@ repos:
         - Format
         - Type checks
     push_allowances:
-      [
-        "alphagov/gov-uk-data",
-        "alphagov/gov-uk-production-admin",
-        "alphagov/gov-uk-production-deploy",
-      ]
+      - "alphagov/gov-uk-data"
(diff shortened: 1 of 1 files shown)


terraform/deployments/github/repos.yml
```diff
@@ -307,7 +307,12 @@ repos:
         - Lint
         - Format
         - Type checks
-    push_allowances: ["alphagov/gov-uk-data", "alphagov/gov-uk-production-admin", "alphagov/gov-uk-production-deploy"]
+    push_allowances:
+      [
```
jfharden: If this is going to me made a multi-line array (and the others) we should really make it the same format as all others and use

```
name:
  - item 1
  - item 2
  - item 3
```

I'd also say this is literally the first time I've ever seen a multi-line array in yaml done c-style (I always knew it was possible, but I've never seen it in the wild)

jfharden (approved):

PR #4332, opened 2026-06-08
SPIKE: tag RDS instances with which app usess it in integration

AP-Hunt:
We're running a spike on Backstage which is reading real data from integration. In order for Backstage to link apps and databases together, the databases must be tagged with the name of the apps that use them.

These tags can be removed by reverting this commit once the spike is done.

### DIFF
--- a/terraform/deployments/rds/variables.tf
+++ b/terraform/deployments/rds/variables.tf
@@ -38,6 +38,8 @@ variable "databases" {
     replica_engine_version    = optional(string)
     replica_apply_immediately = optional(bool)
     replica_multi_az          = optional(bool)
+
+    tags = optional(map(string), {})
   }))
 
   validation {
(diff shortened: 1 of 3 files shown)


terraform/deployments/rds/variables.tf
```diff
@@ -38,6 +38,8 @@ variable "databases" {
     replica_engine_version    = optional(string)
     replica_apply_immediately = optional(bool)
     replica_multi_az          = optional(bool)
+
+    tags = optional(map(string))
```
jfharden: Instead of coalescing, just make this optional with a default


```suggestion
    tags = optional(map(string), {})
```

AP-Hunt: I didn't know you could do that! TIL!

jfharden (approved):

PR #4334, opened 2026-06-09
IA-2750 create workspace for govuk-user-feedback-dev and update trigger patterns

aaronfowles:
- Add govuk-user-feedback-dev workspace (missed in first PR)
- update trigger_patterns of all workspaces that use `gcp-project-init` module so they get updated on changes to that module


jaskaransarkaria (approved):
LGTM

PR #4374, opened 2026-06-17
[PP-7776] Turn off read replica for Publishing API in staging

AgaDufrat:
### Why
We’re removing Publishing API read replica in staging (and shortly in production) to reduce cost as we’re not actively using it at the moment.

#### Notes
The values defaults to false: https://github.com/alphagov/govuk-infrastructure/blob/4bd09a0cf0e958703f890931613a4fdb276e49f8/terraform/deployments/rds/variables.tf#L37

### Depends on:
- https://github.com/alphagov/govuk_content_item_loader/pull/36
- https://github.com/alphagov/frontend/pull/5614
- https://github.com/alphagov/collections/pull/4549
- https://github.com/alphagov/feedback/pull/2482
- https://github.com/alphagov/govuk-helm-charts/pull/4310
- https://github.com/alphagov/govuk-helm-charts/pull/4311

### DIFF
--- a/terraform/variables/staging/rds.tfvars
+++ b/terraform/variables/staging/rds.tfvars
@@ -336,7 +336,6 @@ databases = {
     instance_class               = "db.m7g.2xlarge"
     performance_insights_enabled = true
     project                      = "GOV.UK - Publishing"
-    has_read_replica             = true
   }
 
   release = {


brucebolt (approved):

jfharden (changes requested):
I don't actually need changes, but please don't merge this right now, Sam is in the middle of upgrading release MySQL and the plan for this will make bad changes because of that

jfharden:
You'll need to co-ordinate with @samsimpson1 because he is in the middle of some state shenanigans to upgrade release mysql to 8.4 in staging

samsimpson1 (approved):

jfharden (approved):

PR #4400, opened 2026-06-24
Bring search es cluster into module management

jfharden:
This PR uses the work in https://github.com/alphagov/govuk-infrastructure/pull/4399 to bring the existing elasticsearch ES cluster under control of the module.

You can see what a plan for this looks like in https://app.terraform.io/app/govuk/workspaces/elasticsearch-green-integration/runs/run-RkS7UxTbgDoft67v, or wait until the earlier PR is merged and this changes base to main, at which point terraform cloud will run a plan.

Of note from the plan, and why it's ok:

* `aws_iam_policy.can_configure_es_snapshots` is changing because it could be having an ARN added to it, the plan can't tell that for now it actually wont, so right now it will actually be no change
* `module.opensearch.aws_iam_policy.opensearch_snapshot` is being edited to break out `s3:PutObject` and `s3:DeleteObject` into it's own statement and restricting it to the account being deployed into, this is fine since put and delete is only done by the ES cluster in the account, `s3:GetObject` is now in it's own statement and is allowed from the accounts which need it for data syncs (integration has integration & staging, staging has staing & production, and production only has production, these match what was allowed access prior to the change which can be seen in the plan)
* `module.opensearch.module.green_domain[0].aws_cloudwatch_log_resource_policy.opensearch_log_resource_policy` is:
    * changing the resource allowed to specify all the specific log groups, instead of wildcarding with `*`
    * Adding conditions that say only the ES domains can write to it from the same account
* `module.opensearch.module.snapshot_bucket.aws_s3_bucket_policy.bucket_policy` is:
    * Breaking out `s3:ListObject` and `s3:GetObject` into their own statement and allowing all accounts (as it currently does), but restricting `s3:PutObjectAcl`, `s3:PutObject`, and `s3:DeleteObject` to the current account

The following things are being added:

* A security group & rule, this will remain unused until the blue cluster launches which will have this security group instead of the broad one imported from another workspace
* A secrets manager secret with the details for connecting to the ES domain, currently (since advanced security isn't enabled) this just has the URL to talk to the cluster on in it, but when a cluster is launched with advanced security, it will also include a username and password
* A route53 record, this is a generic CNAME which should be used and will change between blue and green when `current_live_domain` is changed. Right now the configured URLs use the clusters real endpoint, which isn't great for blue-green deploys.
* An IAM role & policy to do snapshots, this is essentially the same as the one already defined, given advanced security is not set up neither or these roles are actually in use, and in a future PR I will remove the older role.


jaskaransarkaria (approved):
LGTM! I like all of the warnings in the source and the deprecated ones in the plans too! ⚠️

PR #4439, opened 2026-07-02
Start clone on 3.6.0 to upgrade Snapshot

dj-maisy:
## What?
This allows us to launch the Staging Cluster on 3.6.0 as we have to import the existing 3.6.0 snapshot before it can be upgraded - we can't just import it straight into a 5.0 cluster.

This also adds some additional variables to allow us to shift the version and cluster configs per-environment without breaking what we've done in Integration.

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/licensify_documentdb_clone.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/licensify_documentdb_clone.tf
@@ -35,11 +35,15 @@ resource "aws_docdb_cluster_parameter_group" "licensify_clone_parameter_group" {
   }
 }
 
+locals {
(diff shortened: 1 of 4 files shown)


AP-Hunt (commented):
There's a failing HCL expression being reported by Terraform Cloud 


```
Error: Invalid index
on licensify_documentdb_clone.tf line 39, in locals:
  db_clone_cluster_parameter_group_name = var.licensify_documentdb_clone_engine_version == "3.6.0" ? aws_docdb_cluster_parameter_group.licensify_parameter_group.name : aws_docdb_cluster_parameter_group.licensify_clone_parameter_group[0].name
aws_docdb_cluster_parameter_group.licensify_clone_parameter_group is empty tuple
The given key does not identify an element in this collection value: the collection has no elements.
```

AP-Hunt (approved):

PR #4440, opened 2026-07-02
GitHub Terraform optimisations

samsimpson1:
* Make API requests to GitHub in parallel. By default the provider makes one request at a time. Performing them in parallel, even with a generous wait period between requests, speeds up plans dramatically
* Remove the `github_repository` data source for all repos that aren't in the process of being archived. This data source isn't actually necessary unless a repo is being archived (to check GH Pages status), so all references to it have been refactored to use data from other places.
* Default to `main` branch for the CodeCommit mirror repos. We don't have any repos that use a different default branch, so we can just assume `main` is being used. This saves on having to use the `github_repository` data source
* Fix a small security issue where repositories are given access to the Slack webhook URL by just having the `govuk` tag set and being in alphagov

Plan times are down from around 9-10 minutes to 4 minutes with these changes.

### DIFF
--- a/terraform/deployments/github/main.tf
+++ b/terraform/deployments/github/main.tf
@@ -35,6 +35,9 @@ provider "aws" {
 provider "github" {
   owner = "alphagov"
 
+  parallel_requests = true
+  read_delay_ms     = 1000
+
   app_auth {
     id              = var.github_app_id              # or `GITHUB_APP_ID`
     installation_id = var.github_app_installation_id # or `GITHUB_APP_INSTALLATION_ID`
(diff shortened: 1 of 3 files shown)


terraform/deployments/github/main.tf
```diff
@@ -36,7 +36,7 @@ provider "github" {
   owner = "alphagov"
 
   parallel_requests = true
-  read_delay_ms = 1000
+  read_delay_ms     = 1000
```
AP-Hunt, line 39: Squash these into the previous commit

AP-Hunt (commented):
This all looks very sensible and easy enough to follow. I'll approve it once you've squashed the last commit, to save me getting my approval dismissed.

AP-Hunt (approved):

PR #4442, opened 2026-07-02
IA-3009: add infra for rds-parquet-exporter job

aaronfowles:
The Insights & Analytics are making a job to export parquet files from RDS - see draft PR https://github.com/alphagov/govuk-helm-charts/pull/4316. We know we need a KMS key and an IAM role but we don't routinely use AWS so hoping to get some feedback on whether this is on the right track given @AP-Hunt 's feedback in that PR.

I've cribbed the KMS from the the [neptune implementation](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/neptune/kms.tf) and the IAM role from the [db-backup](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/govuk-publishing-infrastructure/db_backup_iam.tf). @srinivasadaggu-dotcom has provided the required "actions".

# Requirements

1. dedicated KMS key for the encyrption/decryption of files
2. an IAM role which can kick off the export job and outputs the files to a `parquet/` prefix on the database-backups bucket

# Questions

- Can anyone tell me which kubernetes service accoun the [proposed job](https://github.com/alphagov/govuk-helm-charts/pull/4316/changes) will run as - I need to provide that [here](https://github.com/alphagov/govuk-infrastructure/pull/4442/changes#diff-675df1e3b882f8f3fd575a1eecc5ee178b858e6bd133a47e035d69735430d18dR2) by the looks of it. or do I @srinivasadaggu-dotcom has run the k8s without a dedicated service account and it ran fine - maybe using a default?

# Assumptions
- We only need to run the `rds-parquet-exporter` in gov.uk integration and staging environments.

@srinivasadaggu-dotcom is implementing this so has context. He doesn't have TFC access yet so I'm raising this on his behalf.

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/rds_parquet_exporter_iam.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/rds_parquet_exporter_iam.tf
@@ -1,0 +1,105 @@
+locals {
+  rds_parquet_exporter_service_account_name = "rds-parquet-exporter"
+  parquet_files_bucket_prefix               = "parquet"
+}
+
+module "rds_parquet_exporter_job_starter_iam_role" {
+  source  = "terraform-aws-modules/iam/aws//modules/iam-role-for-service-accounts"
+  version = "~> 6.0"
+
+  name                 = "${local.rds_parquet_exporter_service_account_name}-job-starter-${data.tfe_outputs.cluster_infrastructure.nonsensitive_values.cluster_id}"
+  use_name_prefix      = false
+  description          = "Role for rds-parquet-exporter job to initiate Export. Corresponds to ${local.rds_parquet_exporter_service_account_name} k8s ServiceAccount."
+  max_session_duration = 28800
+
+  policies = {
+    "${aws_iam_policy.rds_parquet_exporter_job_starter.name}" = aws_iam_policy.rds_parquet_exporter_job_starter.arn
+  }
+  oidc_providers = {
+    main = {
+      provider_arn               = data.tfe_outputs.cluster_infrastructure.nonsensitive_values.cluster_oidc_provider_arn
+      namespace_service_accounts = ["apps:${local.rds_parquet_exporter_service_account_name}"]
+    }
+  }
+}
+
+data "aws_iam_policy_document" "rds_parquet_exporter_job_starter" {
+  statement {
+    sid = "RDSExportPermissions"
+    actions = [
+      "rds:DescribeDBSnapshots",
+      "rds:DescribeExportTasks",
+      "rds:StartExportTask"
+    ]
+    resources = ["*"]
+  }
+  statement {
+    actions = ["iam:PassRole"]
+    resources = [
+      aws_iam_role.rds_parquet_exporter_export_task.arn,
+    ]
+    condition {
+      test     = "StringEquals"
+      variable = "iam:PassedToService"
+      values   = ["export.rds.amazonaws.com"]
+    }
+  }
+}
+
+resource "aws_iam_policy" "rds_parquet_exporter_job_starter" {
+  name        = "rds_parquet_exporter_job_starter"
+  description = "Permissions for the RDS parquet exporter job to initiate exports."
+  policy      = data.aws_iam_policy_document.rds_parquet_exporter_job_starter.json
+}
+
+resource "aws_iam_role" "rds_parquet_exporter_export_task" {
+  name                 = "rds-parquet-exporter-export-task-${data.tfe_outputs.cluster_infrastructure.nonsensitive_values.cluster_id}"
+  description          = "Role for RDS to use when exporting parquet files to S3"
+  assume_role_policy   = data.aws_iam_policy_document.allow_rds_exporter_access.json
+  max_session_duration = 28800
+}
+
+data "aws_iam_policy_document" "rds_parquet_exporter_export_task" {
+  statement {
+    sid = "S3AccessForExport"
+    actions = [
+      "s3:PutObject",
+      "s3:ListBucket",
+      "s3:GetBucketLocation",
+      "s3:DeleteObject",
+      "s3:GetObject"
+    ]
+    resources = [
+      "arn:aws:s3:::govuk-${var.govuk_environment}-database-backups",
+      "arn:aws:s3:::govuk-${var.govuk_environment}-database-backups/${local.parquet_files_bucket_prefix}/*",
+    ]
+  }
+  statement {
+    sid = "AllowAccessToKMSKey"
+    actions = [
+      "kms:Encrypt",
+      "kms:Decrypt",
+      "kms:DescribeKey"
+    ]
+    resources = [aws_kms_key.rds_parquet_exporter.arn]
+  }
+}
+
+resource "aws_iam_policy" "rds_parquet_exporter_export_task" {
+  name        = "rds_parquet_exporter_export_task"
+  description = "Permissions for the RDS parquet exporter export task to write to s3."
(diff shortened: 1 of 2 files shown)


terraform/deployments/govuk-publishing-infrastructure/rds_parquet_exporter_iam.tf
```diff
@@ -0,0 +1,66 @@
...
+  description          = "Role for parquet export jobs. Corresponds to ${local.rds_parquet_exporter_service_account_name} k8s ServiceAccount."
+  max_session_duration = 28800
+
+  policies = {
+    "${aws_iam_policy.rds_parquet_exporter_s3.name}" = aws_iam_policy.rds_parquet_exporter_s3.arn
+  }
+  oidc_providers = {
+    main = {
+      provider_arn               = data.tfe_outputs.cluster_infrastructure.nonsensitive_values.cluster_oidc_provider_arn
+      namespace_service_accounts = ["apps:${local.rds_parquet_exporter_service_account_name}"]
```
aaronfowles, line 21: Can someone please clarify whether we need to specify a k8s service account here? @srinivasadaggu-dotcom has deployed the helm chart into integration without a dedicated service account and it seems to be working as expected there - perhaps using a default?

terraform/deployments/govuk-publishing-infrastructure/rds_parquet_exporter_kms.tf
```diff
@@ -0,0 +1,121 @@
+resource "aws_kms_key" "rds_parquet_exporter" {
+  description = "GOV.UK Key for Encrypting RDS parquet exports"
+  key_usage   = "ENCRYPT_DECRYPT"
+  policy      = data.aws_iam_policy_document.rds_parquet_exporter_kms.json
+}
+
+resource "aws_kms_alias" "rds_parquet_exporter" {
+  name          = "alias/govuk/rds_parquet_exporter"
```
AP-Hunt: Aliases should include the environment name so we can deploy multiple full environments in one account; `alias/govuk/${var.govuk_environment}_rds_parquest_exporter`

terraform/deployments/govuk-publishing-infrastructure/rds_parquet_exporter_iam.tf
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+}
+
+data "aws_iam_policy_document" "rds_parquet_exporter_s3" {
+  statement {
+    sid = "S3AccessForExport"
+    actions = [
+      "s3:PutObject",
+      "s3:AbortMultipartUpload",
```
AP-Hunt: You don't need `AbortMultipartUpload` if you're not also using [CreateMultipartUpload](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateMultipartUpload.html)

AP-Hunt (changes requested):
The role you're creating here is a blend of what should be two roles:

1. A role (and Kubernetes service account) which is used by the cron job to start the backups using `rds:StartExportTask`

2. A role that the AWS RDS exporter service (`export.rds.amazonaws.com`) can assume, which grants permission to write to the S3 buckets

What you have so far is majority number 1, and is ~correct for that purpose (once you remove the bits related to number 2).

For number 2, you can use the basic `aws_iam_role` and `aws_iam_policy` resources.

You will also need an additional permission for number 1 to allow it to "pass" role number 2 (in AWS parlance) to RDS:

```
  statement {
    actions = ["iam:PassRole"]
    resources = [
      "ARN_OF_ROLE_2",
    ]
    condition {
      test     = "StringEquals"
      variable = "iam:PassedToService"
      values   = ["export.rds.amazonaws.com"]
    }
  }
```

terraform/deployments/govuk-publishing-infrastructure/rds_parquet_exporter_iam.tf
jaskaransarkaria: Yes you should be specifying a service account here, that way you'll pick up the policies you've attached above.

I'm not sure why it would have worked in integration with the default service account as that won't have permissions to interact with aws. Maybe there was another service account being used? Do you have the job id so I can inspect?

srinivasadaggu-dotcom: @jaskaransarkaria 
Integration environment : [job name ](https://eu-west-1.console.aws.amazon.com/eks/clusters/govuk/jobs/rds-test1?namespace=db-parquet-exports&region=eu-west-1) 

For reference attaching the logs from the test run. 
[logs.txt](https://github.com/user-attachments/files/29631475/logs.txt)

AP-Hunt: The default service account worked because it inherits the permissions that the underlying EC2 instance has, and @srinivasadaggu-dotcom manually changed its IAM policies on June 26th to include the permissions needed.

terraform/deployments/govuk-publishing-infrastructure/rds_parquet_exporter_iam.tf
```diff
@@ -0,0 +1,92 @@
...
+resource "aws_iam_policy" "rds_parquet_exporter_job_starter" {
+  name        = "rds_parquet_exporter_job_starter"
+  description = "Permissions for the RDS parquet exporter job to initiate exports."
+  policy      = data.aws_iam_policy_document.rds_parquet_exporter_job_starter.json
+}
+
+resource "aws_iam_role" "rds_parquet_exporter_export_task" {
+  name                 = "rds-parquet-exporter-export-task-${data.tfe_outputs.cluster_infrastructure.nonsensitive_values.cluster_id}"
+  description          = "Role for RDS to use when exporting parquet files to S3"
+  assume_role_policy   = data.aws_iam_policy_document.rds_parquet_exporter_export_task.json
```
AP-Hunt: The assume role policy for a role should be a separate policy document which grants permission to specific things to use the role. 

```hcl
statement {
    sid = "AllowRDSExporterAccess"
    effect = "Allow"
         
    actions = ["sts:AssumeRole"]
    principals {
        type        = "Service"
        identifiers = ["export.rds.amazonaws.com"]
    }
}
```

AP-Hunt (changes requested):
This is a really good improvement, I particularly like the clear distinction in naming between the two roles. I think the only thing that needs changing is the assume role policy I've highlighted.

terraform/deployments/govuk-publishing-infrastructure/rds_parquet_exporter_iam.tf
aaronfowles: Aha got it - thanks for the clear explanation!

aaronfowles: I've added a new [aws_iam_policy_document](https://github.com/alphagov/govuk-infrastructure/pull/4442/changes#diff-675df1e3b882f8f3fd575a1eecc5ee178b858e6bd133a47e035d69735430d18dR94-R105) and pointed the IAM role's [assume_role_policy](https://github.com/alphagov/govuk-infrastructure/pull/4442/changes#diff-675df1e3b882f8f3fd575a1eecc5ee178b858e6bd133a47e035d69735430d18dR58) to that. My mental model of how this all hangs together has massive holes. Is there anything else that needs changing @AP-Hunt ?

AP-Hunt (approved):
As far as I can tell without seeing everything running, this is correct now

PR #4446, opened 2026-07-02
Apply "contains_pii" tag to databases which contain PII

AP-Hunt:
The need/desire for this tag was identified in an incident retrospective


nicholsj (approved):
LGTM thanks

PR #4447, opened 2026-07-02
Prepare for creating a second snapshot bucket with a better name

jfharden:
This PR gives the opensearch module the ability to not create a snapshot bucket, and also to create a second snapshot bucket with a specified name. It also removes the ability to have a custom suffix for the primary bucket name, this was only needed to import the search cluster.

It also uses this in the search elasticsearch workspace. 

After this PR is applied, I will do a second one to re-enable the primary bucket (I only need to disable it here to allow the moved block to function) with another change that will make it named better so it doesn't have `elasticsearch6` in the name


jfharden:
This PR is defunct, I've taken another approach

PR #4451, opened 2026-07-03
Switch Staging Licennsify DocDB types to Graviton

dj-maisy:
## What?
Now we have upgraded to a compatible version of DocDB, we can upgrade to the newer instance types.


jfharden (approved):
🚀

PR #4461, opened 2026-07-07
Rename elasticsearch-green workspace to search-opensearch

jfharden:
Now the `elasticsearch` workspace is gone, and blue/green deploys are handled by the `module.opensearch` shared module in this workspace, rename the `elasticsearch-green` workspace to `search-opensearch`

Note the plans for `elasticsearch-green-<env>` will all fail because the filepaths are changing, that's expected and ok, and after tfc-configuration is applied I'll run the 3 `search-opensearch-<env>` workspaces


dj-maisy (approved):
This looks good to me. 👍

PR #4484, opened 2026-07-08
IA-3009: allow google-s3-mirror role to decrypt rds-parquet-exporter files

aaronfowles:
> There is an existing role `google-s3-mirror` which is used to mirror database backups to GCP. It doesn't appear to have been defined in terraform and only seems to exist in the integration and staging environments.
> 
> The role doesn't have access to the KMS key used to encrypt parquet files output by `rds-parquet-exporter`. This PR adds a policy to allow access and attaches it to that role.
> 
> The production plan is failing as the role doesn't exist there. Not sure how to address that... Do we maybe import the `google-s3-mirror` role and existing policy?

☝️ resolved by https://github.com/alphagov/govuk-infrastructure/pull/4494

---

The aim of this PR is now to allow the `google-s3-mirror` role decrypt the files output by `rds-parquet-exporter`.

It must only create the required resources in the envs that the `google-s3-mirror` role exists in (i.e. integration and staging).

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/rds_parquet_exporter_iam.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/rds_parquet_exporter_iam.tf
@@ -110,3 +110,35 @@ data "aws_iam_policy_document" "allow_rds_exporter_access" {
     }
   }
 }
+
+resource "aws_iam_policy" "google_s3_mirror_kms_key_access" {
+  count = var.create_google_s3_mirror_role ? 1 : 0
+
+  name        = "google_s3_mirror_kms_key_access"
+  description = "Permissions for decrypting parquet files in s3."
+  policy      = data.aws_iam_policy_document.google_s3_mirror_kms_key_access[0].json
+}
+
+data "aws_iam_policy_document" "google_s3_mirror_kms_key_access" {
+  count = var.create_google_s3_mirror_role ? 1 : 0
+
+  statement {
+    sid = "AllowDecryptionAccessToKMSKey"
+
+    actions = [
+      "kms:Decrypt",
+      "kms:DescribeKey"
+    ]
+
+    resources = [
+      "arn:aws:kms:${data.aws_region.current.region}:${data.aws_caller_identity.current.account_id}:key/*"
+    ]
+  }
+}
+
+resource "aws_iam_role_policy_attachment" "google_s3_mirror_kms_key_access" {
+  count = var.create_google_s3_mirror_role ? 1 : 0
+
+  role       = aws_iam_role.google-s3-mirror[0].name
+  policy_arn = aws_iam_policy.google_s3_mirror_kms_key_access[0].arn
+}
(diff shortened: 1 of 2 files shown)


terraform/deployments/govuk-publishing-infrastructure/rds_parquet_exporter_iam.tf
```diff
@@ -110,3 +110,29 @@ data "aws_iam_policy_document" "allow_rds_exporter_access" {
     }
   }
 }
+
+data "aws_iam_role" "google_s3_mirror" {
```
AP-Hunt: If you rebase your PR on `main` you'll have access to the `aws_iam_role` resource which defines that role now, instead of needing to use a data lookup :+1:

aaronfowles: Thanks @AP-Hunt . I've rebased to use the imported resource now. Also, I think I've scoped down permissions on the KMS `AllowGCPTransferRoleToDecrypt` statement as pointed out by @jfharden.

aaronfowles: Hmmm, I introduced a circular dependency (sorry the commit is gone but here's the [error](https://github.com/alphagov/govuk-infrastructure/actions/runs/29251402643/job/86820604379?pr=4484))

I tried changing the [resource](https://github.com/alphagov/govuk-infrastructure/pull/4484/changes#diff-5742eab2e25462c7a084353f1214e8f6bc1585139de68e1da26cf0fd75fa0353R138) section to `resources = [aws_kms_key.rds_parquet_exporter.arn]` 🤔

AP-Hunt: Yeah that's a really normal problem in Terraform because AWS really does have a circular dependency caused by the role and the resource having their own policies.

The policy on the resource, in this case, has the final say on who can use it. The policy on the role dictates whether it can make the API call, not whether it will be successful.

I'd change the role's policy to this

```hcl
data "aws_iam_policy_document" "google_s3_mirror_kms_key_access" {
  count = var.create_google_s3_mirror_role ? 1 : 0

  statement {
    sid = "AllowDecryptionAccessToKMSKey"

    actions = [
      "kms:Decrypt",
      "kms:DescribeKey"
    ]

    resources = [
        "arn:aws:kms:${data.aws_region.current.region}:${data.aws_caller_identity.current.account_id}:key/*"
    ]
  }
}
```

That will give the role the permission to describe any KMS key in the current account and region, and to _attempt_ decryption. The policies on the keys will protect them from unauthorised usage.

IMO this is the lesser evil between this and making the key policy more widely scoped.

aaronfowles: Thanks, now updated @AP-Hunt

AP-Hunt (approved):

PR #4487, opened 2026-07-08
Make multi-az with standby enabled configurable with helpful validations

jfharden:
Previously I assumed if there were non-t instance types in use we should enable muilti_az_with_standby, but this is not correct, it has many implications, and only works on OpenSearch >= 1.3. So this PR makes it a configurable option with a detailed validation to tell the user that it can only be used on OpenSearch >= 1.3, and only with non-burstable instance types.

Sorry about the validation, but terraform doesn't have a builtin semver comparison so I had to parse the heck out of the engine_version.

Currently searech and ai-accelerator both don't have this turned on, being on is a good sensible default, so I've set the default to on, but explicitly set to off in ai-accelrator and search opensearch clusters)


terraform/shared-modules/opensearch-blue-green-deployment/variables.tf
```diff
@@ -83,6 +84,21 @@ variable "blue_cluster_options" {
     condition     = var.blue_cluster_options == null || var.blue_cluster_options.zone_awareness_enabled == false || var.blue_cluster_options.instance_count >= 3
     error_message = "If var.blue_cluster_options.zone_awareness_enabled is true then var.blue_cluster_options.instance_count must be 3 or more."
   }
+
+  validation {
+    condition = var.blue_cluster_options == null || var.blue_cluster_options.multi_az_with_standby_enabled == false || (
+      !startswith(var.blue_cluster_options.instance_type, "t")
```
AP-Hunt, line 90: This is actually remarkably easy to parse as a condition. Well done

AP-Hunt (approved):

PR #4490, opened 2026-07-08
Add role to gcp integration access group

emmalowe:
Add token creator role to the integration access group to allow service account impersonation of the search_admin service account for local development.

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/google_service_account_iam#google_service_account_iam_binding https://docs.cloud.google.com/iam/docs/service-account-permissions#token-creator-role

See equivalent PR for search-api-v2: https://github.com/alphagov/govuk-infrastructure/pull/3814

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-1968

### DIFF
--- a/terraform/deployments/search-api-v2/service_accounts.tf
+++ b/terraform/deployments/search-api-v2/service_accounts.tf
@@ -68,6 +68,17 @@ resource "google_service_account" "search_admin" {
   description  = "Service account to provide access to the search-admin Rails app"
 }
 
+resource "google_service_account_iam_binding" "search_admin_sa_token_creator_iam" {
+  # Only create this resource if the environment is integration.
+  count = terraform.workspace == "search-api-v2-integration" ? 1 : 0
+
+  service_account_id = google_service_account.search_admin.id
+  role               = "roles/iam.serviceAccountTokenCreator"
+  members = [
+    "group:govuk-gcp-access-integration@digital.cabinet-office.gov.uk"
+  ]
+}
+
 resource "google_service_account_key" "search_admin" {
   service_account_id = google_service_account.search_admin.id
 }


hannako (approved):
LGTM!

AP-Hunt (approved):
I was going to say you should make this into a feature flag rather than a magic string, but I've seen this Terraform root isn't set up in the same way as the rest.

I'm going to approve the change now, and I'll refactor it to match the rest later.

terraform/deployments/search-api-v2/service_accounts.tf
```diff
@@ -68,6 +68,17 @@ resource "google_service_account" "search_admin" {
   description  = "Service account to provide access to the search-admin Rails app"
 }
 
+resource "google_service_account_iam_binding" "search_admin_sa_token_creator_iam" {
+  # Only create this resource if the environment is integration.
+  count = terraform.workspace == "search-api-v2-integration" ? 1 : 0
```
jfharden, line 73: We are moving away from Terraform Cloud so this wont work in the "near" ®️  future.

The real ideal would be to add a variable that's something like:

```
variable "create_search_admin_sa_token_creator" {
  type = bool
  description = "...."
  default = false
}
```

Then make this count:
```
count = var.create_search_admin_sa_token_creator ? 1 : 0
```

Failing that just testhing the govuk_environment would be acceptable

```
count = var.govuk_environment == "integration" ? 1 : 0
```

jfharden: Oh by the time I finished writing this andy approved and it was merged 😄 

Hopefully the comment will help in the future anyway

emmalowe: Thanks @jfharden. Context and pointers are always helpful!

PR #4495, opened 2026-07-09
Allow creating opensearch vpc endpoints

jfharden:
PR best viewed with whitespace changes disabled: https://github.com/alphagov/govuk-infrastructure/pull/4495/changes?w=1

For clusters to use remote reindex, the source cluster needs to expose a VPC endpoint, even if the 2 clusters are within the same VPC (See https://aws.amazon.com/about-aws/whats-new/2023/01/amazon-opensearch-service-remote-reindex-vpc-domains/). This adds the ability to launch a vpc endpoint for either blue or green clusters.

Also they need to communicate on port 443 as well as 9200, so add that allowed in all directions and to self.


AP-Hunt (approved):
Some complicated conditions again, but easy enough to follow with a little attention

PR #4498, opened 2026-07-10
Migrate logging s3 bucket to s3 shared module

jfharden:
Migrate logging bucket to shared module.

Some notes:
* I had to add a bunch of options to the replication configuration in the shared module
* I had to fix the validation that dissallows disabling logging, it was special casing the wrong logging bucket name
* Logging is disabled since this bucket is the recipient of the logs (this is already the case)
* The ACL is being removed (the bucket policy has been updated to reflect a log delivery write acl), and bucket owner enforcement is being enabled.
* The bucket policy can't actually be seen, so I will apply in integration, review the bucket policy and ensure log delivery is still happening before I move on to apply other envs
* The replication configuration of the `govuk-<env>-mirror` buckets in `govuk-publishing-infrastructure-<env>` workspaces says it is changing, but shows no changes inside the plan for it, I think this is because I'm adding new fields which are being set in state to the values the API already has them as, so the state will be transformed "changes", but the result is identical "no changes inside the resource plan".


jfharden:

samsimpson1 (approved):

PR #4505, opened 2026-07-13
Refactor search-api-v2 Terraform Cloud workspaces

AP-Hunt:
## What

A while back, we changed how our Terraform Cloud workspaces are structured, to focus on usign tfvars files instead of Terraform Cloud variable sets.

The search-api-v2 workspaces were not working like this, making it harder to maintain. This commit refactors the workspaces to work like the others.

## How to review

1. Check the plans and make sure that
    1. The workspace is not going to be deleted
    1. The non-environmental variables are being removed 
    1. The environment variables are being deleted and re-created (it doesn't matter if they are, and it's harder work to move them)
2. Check the values of the environment variables match their old values in each workspace


jfharden (approved):
This looks like it should be fine.

Definitely worth doing a plan on each of the workspaces after you have applied tfc-configuration

jfharden (changes requested):
Actually hold off

terraform/variables/staging/search-api-v2.tfvars
```diff
@@ -0,0 +1,3 @@
+gcp_project_id     = "search-api-v2-production"
+gcp_project_number = "931453572747"
```
jfharden: This is the production gpc project number

terraform/variables/staging/search-api-v2.tfvars
```diff
@@ -0,0 +1,3 @@
+gcp_project_id     = "search-api-v2-production"
```
jfharden: This should be -staging

terraform/variables/production/search-api-v2.tfvars
```diff
@@ -0,0 +1,3 @@
+gcp_project_id     = "search-api-v2-staging"
```
jfharden: This should be -production

terraform/variables/production/search-api-v2.tfvars
```diff
@@ -0,0 +1,3 @@
...
+gcp_project_number = "773027887517"
```
jfharden: This is the staging project number but this is the production tfvars file

terraform/variables/staging/search-api-v2.tfvars
```diff
@@ -0,0 +1,3 @@
...
+gcp_project_number = "931453572747"
+gcp_env            = "production"
```
jfharden: This should be staging

jfharden (changes requested):
The tfvars files for staging and production are a mishmash of each others variables.

Staging should be:

```
gcp_project_id     = "search-api-v2-staging"
gcp_project_number = "773027887517"
gcp_env            = "staging"
```

Production should be:
```
gcp_project_id     = "search-api-v2-production"
gcp_project_number = "931453572747"
gcp_env            = "production"
```

jfharden (approved):
All good :)

PR #4507, opened 2026-07-13
Deploy logging workspace to test

jfharden:
* Make it possible to deploy the logging bucket with gcp logging bucket creation disabled
* Make a logging deployment for the test account (I've done this with a variable-set, not variables files, because all of the stuff in test is still using this mechanism, so better not to diverge.

### DIFF
--- a/terraform/deployments/logging/google_logging_bucket.tf
+++ b/terraform/deployments/logging/google_logging_bucket.tf
@@ -1,6 +1,10 @@
-data "google_project" "project" {}
+data "google_project" "project" {
+  count = var.create_google_logging ? 1 : 0
+}
 
 resource "google_storage_bucket" "google_logging" {
+  count = var.create_google_logging ? 1 : 0
+
   name          = "govuk-${var.govuk_environment}-gcp-logging"
   location      = "eu"
   storage_class = "MULTI_REGIONAL"
@@ -22,7 +26,7 @@ resource "google_storage_bucket" "google_logging" {
 }
 
 resource "google_storage_bucket_access_control" "google_logging" {
-  bucket = google_storage_bucket.google_logging.name
+  bucket = google_storage_bucket.google_logging[0].name
 
   role   = "WRITER"
   entity = "group-cloud-storage-analytics@google.com"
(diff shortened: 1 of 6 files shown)


terraform/deployments/tfc-configuration/logging.tf
```diff
@@ -122,3 +122,42 @@ module "logging-production" {
   ]
 }
 
+module "logging-test" {
+  source = "github.com/alphagov/terraform-govuk-tfe-workspacer"
+
+  organization      = var.organization
+  workspace_name    = "logging-test"
+  workspace_desc    = "This module manages the logging deploymentsthat are required by most other modules (VPC, DNS zones)"
```
samsimpson1:
```suggestion
  workspace_desc    = "This module manages the logging deployments that are required by most other modules (VPC, DNS zones)"
```

terraform/deployments/tfc-configuration/logging.tf
```diff
@@ -122,3 +122,42 @@ module "logging-production" {
...
+    oauth_token_id = data.tfe_oauth_client.github.oauth_token_id
+  }
+
+  team_access = {
+    "GOV.UK Non-Production (r/o)" = "write"
+    "GOV.UK Production"           = "write"
+  }
+
+  tfvars_files = [
+    "integration/logging.tfvars"
```
samsimpson1: It should use its own tfvars file, even if its identical to integration's (or empty)

terraform/deployments/logging/google_logging_bucket.tf
```diff
@@ -22,7 +26,7 @@ resource "google_storage_bucket" "google_logging" {
 }
 
 resource "google_storage_bucket_access_control" "google_logging" {
```
samsimpson1, line 28: Won't this error if `create_google_logging = false`? It'll try referencing a resource that doesn't exist

jfharden: Oh yeah you're right, I had meant to condition this too 🫠 Good catch!

terraform/deployments/tfc-configuration/logging.tf
jfharden: I didn't mean to leave this in, and in fact there are no tfvars files at all in this directory anymore, so I think I can just remove the whole setting

jfharden: We had a discussion in slack and I agreed to move to using the same tfvars variable pattern as other workspaces given this is the first "real" (non-ephemeral) deployment into test

samsimpson1 (approved):

jfharden:

PR #4512, opened 2026-07-14
Import test logging bucket

jfharden:
Turns out a bunch of the stuff for the logging deployment already exists in test, and some cannot be created. So:

* Make it possible to disable vpc flow logs log creation, and disable for test logging deployment
* Import the exist s3 buckets, some roles and policies

After apply in logging-test I'll need to do a follow on PR to remove the imports which will be unapplyable and unplanable in the non-test workspaces

It's making a few changes to the logging bucket, but they bring it inline with our current standards and setup


samsimpson1 (approved):

jfharden:

PR #4517, opened 2026-07-14
Migrate content_data_csvs s3 bucket to use the s3 module

jfharden:
This bucket is purposefully public, so I've included a public-read equivalent bucket policy, and set to disable the public access block.

Given we decided to enforce versioning, I've added an additional lifecycle rule to delete the non-current versions 7 days after they become non-current. This means if a second overwrites a first version the non-current will definitely live for the whole time of the lifecycle of the original current version

Deleting the ACL will not actually change the ACL as we uncovered previously, and it will fail to set bucket owner enforcement since the acl is still set. This is good because it means on first apply it will add the bucket policy which also grants get object and list bucket to the world, and then at the cli I can change the canned acl to private before applying the change to bucket owner enforcement.

Note in integration there are other changes happening which are coming from outside this PR, I think perhaps that stuff has been clickopsed for testing, but we have to apply so it will get wiped out, but it should be set via terraform anyway.

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/content_data_s3.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/content_data_s3.tf
@@ -35,11 +35,24 @@ module "content_data_csvs_s3_bucket" {
 
 data "aws_iam_policy_document" "content_data_csvs_s3_bucket_public_read" {
   statement {
-    sid = "AllowPublicRead"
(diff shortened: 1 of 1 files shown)


jfharden:
Relates to: https://github.com/alphagov/govuk-infrastructure/issues/3889

AP-Hunt (approved):

PR #4519, opened 2026-07-15
Block aggressive bot crawlers in CKAN WAF

amar-gds-ndl:
GPTBot and ClaudeBot were not blocked at any level in production and were hitting CKAN at over 100 rps combined. This was exhausting all gunicorn workers and causing liveness probe failures and production outages.

This change blocks bots at the WAF before requests reach the ALB, which removes the load entirely from the cluster. The nginx-level blocking in govuk-dgu-charts remains as a second line of defence but the WAF is the right place to handle this.

I have split the bots into two groups and added two new rules to the existing CKAN WAF Web ACL.

### DIFF
--- a/terraform/deployments/datagovuk-infrastructure/wafs.tf
+++ b/terraform/deployments/datagovuk-infrastructure/wafs.tf
@@ -372,7 +372,7 @@ resource "aws_wafv2_web_acl" "ckan" {
     }
     statement {
       rate_based_statement {
-        limit              = 100
+        limit              = 100 # AWS WAF rate windows are always 5 minutes, so this is 100 requests per 5-minute window per User-Agent string
         aggregate_key_type = "CUSTOM_KEYS"
         custom_key {
           header {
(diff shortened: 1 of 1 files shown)


kentsanggds:
Is there any reasoning behind splitting the bots into 2 groups, one for the hard block and the other for rate limiting?

terraform/deployments/datagovuk-infrastructure/wafs.tf
```diff
@@ -279,12 +326,88 @@ resource "aws_wafv2_web_acl" "ckan" {
...
+          response_header {
+            name  = "Retry-After"
+            value = "300"
+          }
+        }
+      }
+    }
+    statement {
+      rate_based_statement {
+        limit              = 100
```
kentsanggds: Is this a limit per 5 minutes?

amar-gds-ndl: Yes AWS WAF rate-based rules always use a 5-minute window which means this is 100 requests per 5 minutes per User-Agent string. 

I have added a comment to the code to make that clear.

kentsanggds (approved):
LGTM

AP-Hunt (approved):

PR #4521, opened 2026-07-15
Allow creating remote connections in opensearch module

jfharden:
Allow establishing a remote connection from blue to green or green to blue clusters which automatically creates a vpc endpoint when one is needed.

Also remove the explicit creation of vpc endpoints now they are not required for search openseasrch in integration.

Also authorize vpc endpoint use if needed


terraform/shared-modules/opensearch-blue-green-deployment/remote_connections.tf
```diff
@@ -0,0 +1,62 @@
+resource "aws_opensearch_outbound_connection" "to_green_from_blue" {
+  count = var.create_remote_connection_to_import_to_blue_from_green ? 1 : 0
```
samsimpson1, line 2: Is this the right var? Shouldn't it be var.create_remote_connection_to_import_to_green_from_blue?

samsimpson1 (approved):

terraform/shared-modules/opensearch-blue-green-deployment/remote_connections.tf
jfharden: No, the remote connection is established the opposite way around from the direction of data 🫠 

So to copy to blue from green, you establish an outbound connection from blue to green 😢

PR #4530, opened 2026-07-16
Fix CKAN WAF regex pattern set exceeding 10-pattern limit

amar-gds-ndl:
The regex pattern set had 12 individual patterns which exceeded the AWS WAFv2 limit of 10 patterns per set, causing a WAFLimitsExceededException on apply.

I have now consolidated the 12 patterns into 6 by combining pairs using regex alternation.


kentsanggds (approved):
LGTM

samsimpson1 (approved):

PR #4533, opened 2026-07-16
Use opensearch version 2.19 for the blue cluster

koetsier:
We want to try different versions to see if reindexing works. The blue cluster is currently unused.


jfharden (approved):
The "fix" to make it autmatically recreate the outbound connection isn't working :(

I'm on holiday for the next week and a half. In the meantime if you manually delete the outbound connection before you merge this PR it will create it again automatically when it recreates the cluster

PR #4559, opened 2026-07-24
chore: 🤖 migrate search_sitemaps bucket to module

### LINKED ISSUE #3891: Migrate terraform/deployments/govuk-publishing-infrastructure/search_sitemaps_s3.tf to S3 Shared Module

> [!CAUTION]

### Acceptance Criteria

jaskaransarkaria:
4 to create:

> \+ owenership controls (the same "ObjectWriter")
> \+ bucket policy (adds https only policy)
> \+ server side encryption (stays the same, on by default)
> \+ versioning stays the same (disabled)

1 to change:

> ~ add extra "Name" tag to bucket

closes: https://github.com/alphagov/govuk-infrastructure/issues/3891

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/search_sitemaps_s3.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/search_sitemaps_s3.tf
@@ -1,21 +1,49 @@
-resource "aws_s3_bucket" "search_sitemaps_bucket" {
-  bucket = "govuk-${var.govuk_environment}-sitemaps"
-}
+module "search_sitemaps_bucket" {
+  source = "../../shared-modules/s3"
 
-resource "aws_s3_bucket_logging" "search_sitemaps_bucket" {
-  bucket        = aws_s3_bucket.search_sitemaps_bucket.id
-  target_bucket = "govuk-${var.govuk_environment}-aws-logging"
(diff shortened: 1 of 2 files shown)


terraform/deployments/govuk-publishing-infrastructure/search_sitemaps_s3.tf
```diff
@@ -1,21 +1,49 @@
...
+module "search_sitemaps_bucket" {
+  source = "../../shared-modules/s3"
+
+  govuk_environment = var.govuk_environment
+  name              = "govuk-${var.govuk_environment}-sitemaps"
+
+  versioning_enabled         = false
+  enable_public_access_block = false
+
+  disable_bucket_logging = startswith(var.govuk_environment, "eph-") ? true : false
```
AP-Hunt, line 10: This is adding a new condition that isn't present in the original?

jaskaransarkaria: the extra conditional, you're right it is new, but I think it's sane too because these buckets / workspaces rarely if ever are deployed to eph. If they were for some reason deployed to eph you probably wouldn't want those logs going into the access bucket imo

AP-Hunt (approved):

PR #4560, opened 2026-07-24
chore: 🤖 migrate publishing asset bucket to module

### LINKED ISSUE #3890: Migrate terraform/deployments/govuk-publishing-infrastructure/asset_manager_s3.tf to S3 Shared Module

> [!CAUTION]

### Acceptance Criteria

jaskaransarkaria:
3 to create:

> \+ ownership controls (same as before ObjectWriter)
> \+ bucket policy (https only for s3 actions)
> \+ server side encryption (same as before default)

1 to change:

> ~ adds "Name" tag to bucket

closes: https://github.com/alphagov/govuk-infrastructure/issues/3890

### DIFF
--- a/terraform/deployments/govuk-publishing-infrastructure/asset_manager_s3.tf
+++ b/terraform/deployments/govuk-publishing-infrastructure/asset_manager_s3.tf
@@ -1,22 +1,36 @@
-resource "aws_s3_bucket" "assets" {
-  bucket = "govuk-assets-${var.govuk_environment}"
+locals {
+  assets_bucket_name = "govuk-assets-${var.govuk_environment}"
+  assets_bucket_arn  = "arn:aws:s3:::${local.assets_bucket_name}"
 }
 
-resource "aws_s3_bucket_versioning" "assets" {
-  bucket = aws_s3_bucket.assets.id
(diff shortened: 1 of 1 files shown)


terraform/deployments/govuk-publishing-infrastructure/asset_manager_s3.tf
```diff
@@ -1,22 +1,36 @@
...
+module "assets" {
+  source = "../../shared-modules/s3"
+
+  govuk_environment = var.govuk_environment
+  name              = local.assets_bucket_name
+
+  versioning_enabled         = true
+  enable_public_access_block = false
+
+  disable_bucket_logging = startswith(var.govuk_environment, "eph-") ? true : false
```
AP-Hunt, line 15: Adds a condition that isn't present in the original?

jaskaransarkaria: the extra conditional, you're right it is new, but I think it's sane too because these buckets / workspaces rarely if ever are deployed to eph. If they were for some reason deployed to eph you probably wouldn't want those logs going into the access bucket imo

AP-Hunt (approved):

PR #4564, opened 2026-07-27
Create IAM configuration for data.gov.uk db backups to sync across accounts

AP-Hunt:
## What

The new DGU team in the National Data Library want to synchronise their database backups from our account to their new account. For now this is to be done in the integration environment only, but it will expand to other environments later.

## How to reviwe

1. Code review 
2. Check the plans and make sure integration is the only one with changes


nimalank7 (approved):
LGTM!

PR #4567, opened 2026-07-27
Export RDS backups S3 bucket ARN

AP-Hunt:
The ARN will be needed in IAM policies in a future commit, but the change needs to be rolled out independently first.


nimalank7 (approved):
LGTM!

PR #4570, opened 2026-07-27
Block amazon-Quick and TalonGDP bots in CKAN WAF (DGUK-505)

amar-gds-ndl:
This Adds two additional bot User-Agent patterns to the CKAN WAF hard-block rule (priority 0, 403 response), identified during the Wednesday 22 July production outage investigation.

**New patterns added to `ckan_hard_block_bots` regex pattern set:**
- `amazon-Quick-on-behalf-of-*` — an Amazon scraper bot currently generating significant unblocked traffic on CKAN production (observed top UA in nginx logs, 550+ requests per 12-minute window)
- `TalonGDPvalAttachmentGroupExplorer/0.1` — unidentified public attachment discovery crawler observed in logs

Pattern set now uses 8 of the 10 allowed patterns.


kentsanggds (approved):
I'll approve this for now, I wonder if it would be easier to have an allowlist of User-Agents

AP-Hunt (approved):

PR #4572, opened 2026-07-28
fix: 🐛 log s3 access to the correct eph bucket

jaskaransarkaria:
Ephemeral cluster builds fail when deploying buckets using our shared s3 module because the are trying to log to a bucket that doesn't exist

Screenshot 2026-07-28 at 10 53 53

> var.govuk_environment points to the cluster name and not at "ephemeral"

### DIFF
--- a/terraform/shared-modules/s3/main.tf
+++ b/terraform/shared-modules/s3/main.tf
@@ -1,3 +1,9 @@
+data "aws_caller_identity" "current" {}
+
+locals {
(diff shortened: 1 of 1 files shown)


AP-Hunt (dismissed):

AP-Hunt:
Approved this now, but I'd like to refactor how ephemeral environment workspaces work so that we can more easily use feature flags for them

AP-Hunt (approved):

PR #4574, opened 2026-07-28
IA-3078: import new GovGraph datasets and set permissions

aaronfowles:
https://gov-uk.atlassian.net/browse/IA-3078


srinivasadaggu-dotcom (approved):
LGTM

PR #4583, opened 2026-07-29
Sch 2011 Remove quota overrides for search

emmalowe:
These quota overrides do not reflect our current project setup, and we don't want to change the current setup with respect to overrides, so these should be removed.
    
The overrides for discoveryengine.googleapis.com/documents and for discoveryengine.googleapis.com/search_requests are no longer listed under the [consumerQuotaMetrics endpoint](https://docs.cloud.google.com/service-usage/docs/reference/rest/v1beta1/services.consumerQuotaMetrics/list). They seem to have been replaced with regional equivalents. Our current settings for the regional equivalents match either the default values, or in some cases have been set higher than the default, so capping the quotas via overrides is not part of the desired setup.
    
However, we still want to keep the documentation for overrides in the README, in case it is useful at a later date.
    
The `terraform plan` before these changes used to show 6 resources related to quota overrides being added. After this change, there are none.

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-2011

### DIFF
--- a/terraform/deployments/gcp-search-api-v2/modules/search-api-v2/main.tf
+++ b/terraform/deployments/gcp-search-api-v2/modules/search-api-v2/main.tf
@@ -51,34 +51,6 @@ resource "google_project_service" "api_service" {
   disable_dependent_services = true
 }
 
-resource "google_service_usage_consumer_quota_override" "discoveryengine_search_requests" {
-  provider = google-beta
-  project  = google_project.environment_project.project_id
-
-  service = "discoveryengine.googleapis.com"
-  metric  = urlencode("discoveryengine.googleapis.com/search_requests")
-  force   = true
-
-  # limit is equivalent to `unit` field when making a GET request against the metric, but without
-  # leading `1/` and without curly braces
-  limit          = urlencode("/min/project")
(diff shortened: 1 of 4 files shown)


stringiest (approved):
This looks good!

nimalank7 (approved):
LGTM!

terraform/deployments/gcp-search-api-v2/modules/search-api-v2/main.tf
```diff
@@ -51,34 +46,6 @@ resource "google_project_service" "api_service" {
   disable_dependent_services = true
 }
 
-resource "google_service_usage_consumer_quota_override" "discoveryengine_search_requests" {
```
emmalowe, line 54: @hannako I had another look at this today. The `override_value` is a required integer argument (see [docs](https://registry.terraform.io/providers/hashicorp/google-beta/latest/docs/resources/service_usage_consumer_quota_override#argument-reference)) so we can't set this to null or leave it out. I ran the plan with the default limits as the override values and they still show up in the plan so I think the best thing to do if we want to leave this code in tact so we can easily implement limits in the future is to just comment it out.

hannako (approved):

PR #4584, opened 2026-07-29
Sch 2011 document terraform folder for search

emmalowe:
Updates READMEs related to search-api-v2 deployments to provide additional content on Terraform setup.

Also removes project_id from `gcp-search-api-v2` since this variable was unused but was required to be specified
when running Terraform, which was confusing.

Jira ticket: https://gov-uk.atlassian.net/browse/SCH-2011

### DIFF
--- a/terraform/deployments/gcp-search-api-v2/variables.tf
+++ b/terraform/deployments/gcp-search-api-v2/variables.tf
@@ -20,11 +20,6 @@ variable "google_cloud_billing_account" {
   description = "The ID of the Google Cloud billing account to associate projects with"
 }
 
-variable "project_id" {
-  type        = string
-  description = "The ID of the overarching terraform cloud project for all workspaces"
-}
-
 variable "tfe_project_name" {
   type        = string
   default     = "govuk-search-api-v2"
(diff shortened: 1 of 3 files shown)


stringiest (approved):
This makes much more sense now - thanks! ⭐

nimalank7 (approved):
LGTM!

terraform/deployments/gcp-search-api-v2/variables.tf
```diff
@@ -20,11 +20,6 @@ variable "google_cloud_billing_account" {
   description = "The ID of the Google Cloud billing account to associate projects with"
 }
 
-variable "project_id" {
```
hannako, line 23: You're deleting this because it's not referenced in https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/deployments/gcp-search-api-v2/main.tf - is that correct?

hannako (approved):

terraform/deployments/gcp-search-api-v2/variables.tf
emmalowe: Correct!

PR #4586, opened 2026-07-30
Allow ckan-db-backup-xaccount-sync-* role sufficient permissions for s3 sync

AP-Hunt:
## What

This role will be used to perform an `aws s3 sync` (without recursive or delete flags), and I missed some permissions for it.

## How to review

1. Do a bit of research of your own to see what permissions are actually required and whether I've missed any

### DIFF
--- a/terraform/deployments/datagovuk-infrastructure/ckan_db_cross_account_backup_iam.tf
+++ b/terraform/deployments/datagovuk-infrastructure/ckan_db_cross_account_backup_iam.tf
@@ -33,11 +33,16 @@ data "aws_iam_policy_document" "ckan_db_backup_xaccount_sync" {
   statement {
     sid     = "AllowReadingFromSource"
     effect  = "Allow"
-    actions = ["s3:GetObject", "s3:HeadObject", "s3:ListObjects"]
+    actions = ["s3:CopyObject", "s3:GetObject", "s3:HeadObject", "s3:ListObjects"]
 
-    resources = [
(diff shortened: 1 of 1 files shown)


terraform/deployments/datagovuk-infrastructure/ckan_db_cross_account_backup_iam.tf
```diff
@@ -33,11 +33,16 @@ data "aws_iam_policy_document" "ckan_db_backup_xaccount_sync" {
...
-    resources = [
-      "${data.tfe_outputs.rds.nonsensitive_values.database_dump_bucket_arn}:ckan-postgres/*"
-    ]
+    resources = ["${data.tfe_outputs.rds.nonsensitive_values.database_dump_bucket_arn}:ckan-postgres/*"]
+  }
+
+  statement {
+    sid       = "AllowListingSourceBucket"
+    effect    = "Allow"
+    actions   = ["s3:ListBucket"]
```
jfharden: When I was reading what perms sync needed when first looking how we might do what DGU needs I read it also needed the `s3:GetBucketLocation` permission, I'm not clear if it is needed on source, destination, or both (both seems likely)

jfharden: I assume you're not going to run with the `--delete` flag (I wouldn't), but if you are you also need DeleteObject

AP-Hunt: Good catch I'll add it

jfharden (approved):

PR #4587, opened 2026-07-30
feat: 🎸 add gk module

### LINKED ISSUE #4359: [Gatekeeper] - [Audit mode] - Install the gatekeeper terraform module into the cluster
**User Need**

**As a** platform engineer
**I want** to have a policy engine validating resources and their changes in our cluster
**so that** we can enforce rules about what resources are created and how they are changed

---

### Context

https://github.com/alphagov/govuk-infrastructure/issues/4208
https://github.com/alphagov/govuk-infrastructure/pull/4329

---

### What’s Needed

List anything the solution must do or be (behaviour, performance, security, UX, etc.).

- [ ] add the [terraform module](https://github.com/alphagov/govuk-terraform-gatekeeper) into the infra repo (cluster-services feels like a good place)
- [ ] add it into it's own namespace
- [ ] make sure there are no mutations deployed unless confirmed by the rest of the team as ok
- [ ] do not include any rules when deploying, we will decide and turn those on in a subsequent ticket
- [ ] deploy this into integration and monitor the log volume for a couple of days in logit before deploying this to production
---

### Acceptance Criteria

Clear, measurable conditions to verify the story is complete and valuable.

- [ ] gatekeeper is deployed successfully in audit only mode and isn't affecting resource creation and updates

---

### Risks & Mitigation (optional)

- Affects resources across the cluster
- Mutates pods and affects their behaviour

In audit only mode we avoid any potential issues with resources

---

jaskaransarkaria:
This PR adds gatekeeper as a module in infra and deploys it into the `gatekeeper-system` namespace via `cluster-services` terraform workspace. It is deployed into an audit only mode so it will not block any resource CRUD actions in the cluster.

> [!NOTE]  
> ~~There are a lot of files in this pr but a lot of them are gatekeeper tests, you probably only need to view one of them. The same is true for constraints and constraint_templates under the `module/gatekeeper/resources/` dir~~ There were a lot of files included in the pr because I thought the constraints would make good reference material, on second thoughts you can access those constraint example here https://github.com/alphagov/govuk-terraform-gatekeeper

When testing the module, I've verified gatekeeper can be:

- installed successfully into a pre-existing cluster
- installed successfully into a fresh cluster
- whilst in audit mode does not interfere with the deployment of resources
- can view [audited](https://open-policy-agent.github.io/gatekeeper/website/docs/audit/) resources

## Reviewing the gatekeeper module

I've included lots of policies (and their tests) that we may choose to delete at a later date but I think keeping them there for reference as a sort of library is useful whilst the module is new

Because we have these policies in dry run mode there are a few resources we need to sync into gatekeeper's cache via the `Config` resource, you can find them in `terraform/deployments/cluster-services/modules/gatekeeper/main.tf`

## Reviewing the terraform calling the module

Other than removing `kyverno` and adding the `kubectl` provider to `cluster-services`, I've mainly had to move around the `depends_on` hierarchy. We want gatekeeper to come in as soon as possible so it can enforce the policies we choose on everything coming into the cluster (including other cluster services). It has a dependency on the `aws_lb_controller` so gatekeeper needs to depend on that but then everything else that previously depended on `aws_lb_controller` is moved down the hierarchy and now depends upon `module.gatekeeper`

```mermaid
flowchart TB
    A["helm_release.aws_lb_controller"] -- <br> --> B("module.gatekeeper")
    B --> n1["helm_release.external_secrets"]
    A --> n2["helm_release.dex"]
    n1 --> n3["helm_release.cluster_secrets"]
    n3 --> n2
    n2 --> n4["helm_release.argo_workflows"]

    n1@{ shape: rounded}
    n2@{ shape: rounded}
    n3@{ shape: rounded}
    n4@{ shape: rounded}
```

closes https://github.com/alphagov/govuk-infrastructure/issues/4359


terraform/deployments/cluster-services/modules/gatekeeper/resources/constraint_templates/block_ingresses.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+  targets:
+    - target: admission.k8s.gatekeeper.sh
+      rego: |
+        package k8sblockingresses
+
+        violation[{"msg": msg}] {
+          input.review.kind.kind == "Ingress"
+          not input.review.object.metadata.namespace == "jaskaran-dev"
+          msg := sprintf("\nIngress actions are temporarily blocked whilst the Cloud Platform goes under essential maintenance. The maintenance window is on Tuesday 4th March between 1900 to 2300. For more info see https://mojdt.slack.com/archives/CH6D099DF/p1740668486144479. Please raise any other queries in #ask-cloud-platform.", [])
```
AP-Hunt: We should review all the policies and make sure we remove anything that straight up isn't relevant, or just names the wrong things

terraform/deployments/cluster-services/gatekeeper.tf
```diff
@@ -0,0 +1,39 @@
...
+    ingress_internal_class_domain      = true,
+    immutable_configmap                = true
+  }
+
+  is_production                        = terraform.workspace == "production"
+  environment_name                     = var.govuk_environment
+  cluster_domain_name                  = local.external_dns_zone_name
+  constraint_violations_max_to_display = 25
+
+  controller_mem_limit = terraform.workspace == "production" ? "8Gi" : "1Gi"
```
AP-Hunt: Let's make these part of the tfvars for `cluster-services` instead of relying on magic strings

terraform/deployments/cluster-services/modules/gatekeeper/constraint_templates/templates.tf
```diff
@@ -0,0 +1,7 @@
+# remember any kind used by a constraint template must also be added to the sync config at the end of this file
```
AP-Hunt: There's no sync config at the end of the file

terraform/deployments/cluster-services/modules/gatekeeper/constraints/constraints.tf
```diff
@@ -0,0 +1,7 @@
+resource "kubectl_manifest" "constraints" {
+  for_each = local.constraint_map
+
+  yaml_body = yamlencode("${each.value}")
```
AP-Hunt: This can be `yamlencode(each.value)` I think

terraform/deployments/cluster-services/modules/gatekeeper/constraints/locals.tf
```diff
@@ -0,0 +1,46 @@
+locals {
+  service_type_yaml                       = yamldecode(file("${path.module}/../resources/constraints/service_type.yaml"))
```
AP-Hunt: You could reduce the repetition here by capturing `${path.module}/../resources/constraints/` as a local and reusing it

terraform/deployments/cluster-services/modules/gatekeeper/constraints/locals.tf
```diff
@@ -0,0 +1,46 @@
...
+  allow_duplicate_hostname_yaml           = yamldecode(file("${path.module}/../resources/constraints/ingress_allow_duplicate_hostname.yaml"))
+  block_ingresses                         = yamldecode(file("${path.module}/../resources/constraints/block_ingresses.yaml"))
+  ingress_valid_classname_yaml            = yamldecode(file("${path.module}/../resources/constraints/ingress_valid_classname.yaml"))
+  ingress_internal_class_domain_yaml      = yamldecode(file("${path.module}/../resources/constraints/ingress_internal_class_domain.yaml"))
+  immutable_configmap_yaml                = yamldecode(file("${path.module}/../resources/constraints/immutable_configmap.yaml"))
+
+  # For each constraint, a value needs to be in the constraint map. This bloc allows us to set values on constraints which enables us to toggle the configuration of the constraints. -- we merge in the spec separately to avoid overwriting entire spec key
+  constraint_map = {
+    service_type                       = merge(local.service_type_yaml, { "spec" : merge(local.service_type_yaml["spec"], { "enforcementAction" : var.dryrun_map.service_type ? "dryrun" : "deny" }) })
```
AP-Hunt: Is there a better way to do this at all? It's quite hard to read. Could it be done with a template maybe?

terraform/deployments/cluster-services/modules/gatekeeper/constraints/variables.tf
```diff
@@ -0,0 +1,36 @@
+variable "cluster_domain_name" {
+  description = "The cluster domain used for externalDNS annotations and certmanager"
```
AP-Hunt: Let's add types to these vars and any validations we can make

terraform/deployments/cluster-services/modules/gatekeeper/constraints/variables.tf
```diff
@@ -0,0 +1,36 @@
...
+}
+
+variable "integration_test_zone" {
```
AP-Hunt: What's a zone in this context?

terraform/deployments/cluster-services/modules/gatekeeper/test/suite/block_ingresses.yaml
```diff
@@ -0,0 +1,16 @@
+kind: Suite
```
AP-Hunt: I'm sad these all have to be big YAML files. I was hoping they'd look [more like testing Rego files](https://www.openpolicyagent.org/docs/policy-testing)

terraform/deployments/cluster-services/modules/gatekeeper/resources/constraint_templates/block_ingresses.yaml
```diff
@@ -0,0 +1,24 @@
...
+spec:
+  crd:
+    spec:
+      names:
+        kind: k8sblockingresses
+
+  targets:
+    - target: admission.k8s.gatekeeper.sh
+      rego: |
+        package k8sblockingresses
```
AP-Hunt: 🤔 This might not be an in-this-PR issue, but could we extract the Rego code to their own files? That way we can get better IDE support for them if we want.

terraform/deployments/cluster-services/modules/gatekeeper/main.tf
```diff
@@ -0,0 +1,93 @@
...
+    audit_mem_limit                      = var.audit_mem_limit
+    audit_mem_req                        = var.audit_mem_req
+  })]
+
+  lifecycle {
+    ignore_changes = [keyring]
+  }
+}
+
+resource "time_sleep" "wait_30_seconds" {
```
AP-Hunt, line 36: Are these sleeps to account for the failings of the Helm provider?

AP-Hunt (changes requested):

AP-Hunt:
Ooh, also! Could we get some more detail in the commit message about what exactly is in the commit, please?

terraform/deployments/cluster-services/modules/gatekeeper/resources/constraint_templates/block_ingresses.yaml
jaskaransarkaria: I initially kept all the constraints in the module because I thought they'd be a useful reference for the first handful of new constraints we write. Thinking about it though I think I'll blast all of them and if people need a reference the original repo with the policies and tests in can still be used as a reference

https://github.com/alphagov/govuk-terraform-gatekeeper

terraform/deployments/cluster-services/modules/gatekeeper/constraints/locals.tf
jaskaransarkaria: updated

terraform/deployments/cluster-services/modules/gatekeeper/constraints/variables.tf
jaskaransarkaria: Since I've removed a the previously mentioned reference constraints, there's no longer a need for this var

terraform/deployments/cluster-services/modules/gatekeeper/main.tf
jaskaransarkaria: It's to do with when the destroy happens, it's really important everything gets deleted in order or you can leave the cluster in an irreparable state

terraform/deployments/cluster-services/modules/gatekeeper/constraints/locals.tf
jaskaransarkaria: I didn't explore templates this time around, but when I created the module initially I agonised over this part for quite a while. Whilst it's difficult to read, iirc it's the best method to ensure that the yaml file you test against is _same one_ you pass onto terraform to be deployed.

That's why a template wouldn't work here because the yaml file you are toggling the dryrun on isn't a template, it's valid yaml needed for the tests, I think that trade off is worth it.

Other wise you end up writing something that passes the tests and something that looks different that gets deployed

terraform/deployments/cluster-services/modules/gatekeeper/test/suite/block_ingresses.yaml
jaskaransarkaria: see above comment

terraform/deployments/cluster-services/modules/gatekeeper/constraints/variables.tf
jaskaransarkaria: I've removed these vars since we removed a lot of the reference constraints which were needing them

AP-Hunt (approved):

PR #4595, opened 2026-07-30
Correct the source bucket for CKAN database backups

AP-Hunt:
Backups are stored in the bucket created by the publishing infrastructure, not the bucket created in the RDS root.


nimalank7 (approved):
LGTM!

PR #4625, opened 2026-08-06
Bump kubectl provider to 3.0.0-beta3

jfharden:
kubectl provider less than 3 has a serious bug where if the apply errors the state is updated anyway and now it thinks the change has been applied and will never apply it.

This is fixed in the beta for 3.0, given we are actively hitting the bug in 2 so it's broken for us, I'm taking the unusual step to use the beta version of the provider


nimalank7 (approved):
LGTM!

Issue #2938, opened 2025-10-07
🗑️ Deprecate x86 (AMD64) Builds of GOV.UK Images


dj-maisy:
**User Need**

**As a** Platform Engineer
**I want** to stop making x86 architecture builds of GOV.UK Application Images
**so that** we make our Build Pipelines more efficient and reliable

---

### Context

This was previously suggested after the initial completion of the migration work from x86 to AWS Graviton. We spend more time in GitHub Runner Minutes producing x86 versions of our images that are likely almost never used and the multi-architecture build process has multiple stages including an extra "consolidation and manifest" stage at the end which only runs if all builds are successful. It is complex and hard to maintain.

We put off making this change until the department rolled out new Apple Silicon-based machines to everyone. That is now complete and we can progress this work.

--- 

### What’s Needed

List anything the solution must do or be (behaviour, performance, security, UX, etc.).

- [ ] Create an ARM-only image building GitHub Actions workflow.
- - [ ]  Ensure it works with Docker >= 29
- [ ] Switch the Deploy/Release actions of each individual GOV.UK application repository to use the new build action (consider if https://github.com/alphagov/bulk-changer can be used for this)
- [ ] Switch the workflow with builds images as part of PRs
- [ ] Retire the multi-architecture build process

---

### Acceptance Criteria

Clear, measurable conditions to verify the story is complete and valuable.

- [ ] Applications are building and deploying correctly (i.e. they notice no difference)
- [ ] Developers are notified of the change and the rationale
- [ ] The old workflow has been retired.

---

### Assumptions (optional)

- All Developers are on Apple Silicon MacBooks

jasonBirchall:
[pre-planning] we've decided it should be investigated (user research) before we carry out this work. This needs to go through refinement.

nicholsj:
Suggestion is to wait until next dev laptop rollout to be finalised - but also need to account for contractor/MSP laptops etc depending on policy

AP-Hunt:
Pushing this back until DSIT laptops have been issued to everyone

AP-Hunt:
We all have new ARM-based laptops now, we can make it high priority

AP-Hunt:
I've reworked this a little ahead of refinement

dj-maisy:
The Reusable workflow is "done" - it now needs to be consumed by apps. Starting with Ruby Base Images.

dj-maisy:
### Updates

* Release has been tested successfully.
[...]

PR #4626, opened 2026-08-06
Create ARM-only reusable build workflow.

dj-maisy:
## What?
This is a new reusable workflow file to deprecate the x86 and multi-arch build workflows.

## Key changes
Compared to the previous builds, this PR:

* Provides some more inputs, for more flexibility, but with sensible defaults and with more useful descriptions.
* Sets the Docker Engine to version 29, removing support for some deprecated Docker features.
* Removes the need for the "combine manifests" job, as the manifest will be pushed automatically by docker buildx.
* Uses a shallow fetch to speed up the Git checkout

## Why?
The existing build workflows:

* Are more complex (they use a build matrix to run each architecture (x86 and ARM) in parallel
* Takes longer - more steps to run, consolidation job must be run, and this must wait for all parallel jobs to complete before finishing.
* Doubles the number of build minutes used.

This made sense when we were first exploring the move to ARM, but as we are now exclusively using Apple Silicon and AWS Graviton, we no longer need to tolerate the trade-offs that come with supporting multiple architectures.

### Related Resources

* https://github.com/alphagov/govuk-infrastructure/issues/2938


.github/workflows/build-push-arm-image.yml
```diff
@@ -0,0 +1,208 @@
...
+          labels: |
+            org.opencontainers.image.description=Container image for ${{ github.repository }}
+            org.opencontainers.image.revision=${{ steps.local-head.outputs.sha }}
+            org.opencontainers.image.title=${{ inputs.imageName }}
+            org.opencontainers.image.vendor=GDS
+          annotations: |
+            org.opencontainers.image.authors=Government Digital Service
+            org.opencontainers.image.source=${{ github.server_url }}/${{ github.repository }}
+          tags: |
+            type=raw,value=latest,enable=${{ inputs.pushLatest && github.ref_name == github.event.repository.default_branch }}
```
samsimpson1: Shouldn't this be `input.gitRef == github.event.repository.default_branch`? Otherwise it'd be checking the branch the workflow was run from (usually main if ran manually)

.github/workflows/build-push-arm-image.yml
```diff
@@ -0,0 +1,208 @@
...
+        description: 'Git ref, tag, or SHA to checkout before building.'
+        required: false
+        type: string
+        default: ${{ github.sha }}
+      runnerName:
+        description: 'Runner label used for native ARM64 build.'
+        required: false
+        type: string
+        default: ubuntu-24.04-arm
+      registry:
```
samsimpson1: Is there much point in paramatarising this given the rest of the repository URL is constructed using components that only make sense when using GHCR?

.github/workflows/build-push-arm-image.yml
```diff
@@ -0,0 +1,208 @@
...
+        description: 'Container registry hostname.'
+        required: false
+        type: string
+        default: ghcr.io
+      pushLatest:
```
samsimpson1: build-and-push-multiarch-image.yml always pushes `latest` on a `main` branch build, so shouldn't this default to `true` to preserve current behaviour?

samsimpson1 (approved):

PR #4647, opened 2026-08-12
cloud run job & schedular  deployment code changes

srinivasadaggu-dotcom:
Deploy the GCP Cloud Run Job & Cloud Schedular for rds-parquet-bq-loader & docdb-bq-loader ingestion pipelines.

### DIFF
--- a/terraform/deployments/gcp-gov-graph/docdb-bq-loader.tf
+++ b/terraform/deployments/gcp-gov-graph/docdb-bq-loader.tf
@@ -18,3 +18,82 @@ resource "google_service_account_iam_policy" "service_account_user_docdb_bq_load
   service_account_id = google_service_account.docdb_bq_loader.name
   policy_data        = data.google_iam_policy.service_account_user_docdb_bq_loader.policy_data
 }
+
+# ==============================================================================
+# CLOUD RUN JOB DEPLOYMENT
+# ==============================================================================
+
+resource "google_cloud_run_v2_job" "docdb_bq_loader" {
+  name     = "docdb-bq-loader"
+  location = "europe-west2"
+
+  deletion_protection = true
+
+  # Ensure the Service Account and its IAM policies are fully established
+  # before the Cloud Run Job is deployed.
+  depends_on = [
+    google_service_account_iam_policy.service_account_user_docdb_bq_loader
+  ]
+
+  template {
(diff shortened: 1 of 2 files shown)


terraform/deployments/gcp-gov-graph/docdb-bq-loader.tf
```diff
@@ -18,3 +18,82 @@ resource "google_service_account_iam_policy" "service_account_user_docdb_bq_load
...
+        resources {
+          limits = {
+            cpu    = "2"
+            memory = "8Gi"
+          }
+        }
+
+        env {
+          name  = "CONFIG_PATH"
+          value = "config/live/config.json"
```
aaronfowles: This will need to vary by env presumably?

You could create and parametrise the value of the config file name [here](https://github.com/alphagov/govuk-infrastructure/blob/main/terraform/variables/integration/gcp-gov-graph.tfvars).

aaronfowles (changes requested):
Looks good but we need to account for the fact that different envs use different configs. Happy to discuss.

srinivasadaggu-dotcom:
@aaronfowles
The environment name is now parameterised (var.environment). Thanks

terraform/deployments/gcp-gov-graph/rds-parquet-bq-loader.tf
```diff
@@ -18,3 +18,88 @@ resource "google_service_account_iam_policy" "service_account_user_rds_parquet_b
...
+  deletion_protection = true
+
+  # Ensure the Service Account and its IAM policies are fully established
+  # before the Cloud Run Job is deployed.
+  depends_on = [
+    google_service_account_iam_policy.service_account_user_rds_parquet_bq_loader
+  ]
+
+  template {
+    task_count  = 5
```
aaronfowles, line 39: Is this correct? I was under the impression a single job would handle the ingestion for all sources and should therefore be set to 1. I may have misunderstood!

srinivasadaggu-dotcom: We are still using a single Cloud Run Job. However, we are configuring it to split the workload into 5 tasks that run in 3 way parallel.

aaronfowles (approved):

Issue #4407, opened 2026-06-25
Operator: validate the approver and requester names


AP-Hunt:
**User Need**

**As a**  platform engineer
**I want** prevent a job request system user from approving their own request
**so that** we are mitigating an obvious threat

---

### Context

The requester and approvers names are set on the object by an external process, so don't worry about that in this story. When the operator sees an approval come in, it should validate that the target request's submitter name isn't the same as the approvers name. If they are, it should reject the approval and set an appropriate state.

---

### What’s Needed

List anything the solution must do or be (behaviour, performance, security, UX, etc.).

- [x] Validate the names involved in request and approvals
- [x] In the CLI, prevent a user approving their own requests before it gets sent
- [x] Accept and reject as appropriate 

---

### Acceptance Criteria

Clear, measurable conditions to verify the story is complete and valuable.

- [x] When I try to approve my own request, it is rejected
- [x] When I approve somebody else's request, it is accepted.

---

### Assumptions (optional)

- The approver and requester names are set on the object by a different process

AP-Hunt:
[refinement] not refining this for now because it's blocked by other things

jfharden:
This is blocked by #4606 because we don't actually have a field in the CRD to store the requestor/approver, and that ticket will add it

AP-Hunt:
Dropped into todo because it has a blocker assigned and it's an active piece of work that's being blocked

jfharden:
This is now implemented:

* [In the operator](https://github.com/alphagov/govuk-job-request-operator/pull/97)
* [In gatekeeper policies](https://github.com/alphagov/govuk-infrastructure/pull/4650)
* [In govuk-cli](https://github.com/alphagov/govuk-cli/pull/43)

PR #4650, opened 2026-08-13
Add gatekeeper rules to deny reviewing own jobrequest

jfharden:
NOTE: This PR looks massive, but it's almost entirely yaml fixture files, the actual rego to review is not that big

Add gatekeeper rules to:

* Parse the requested-by and reviewed-by ARNs to usernames, and fail the rule if they cannot be parsed as expected
* Deny creating a JobRequestReview which tries to review a JobRequest created by the same user

I've written it as a rego v0 rule in this PR, in [this follow on PR I will convert all the rules in a single PR to rego v1](https://github.com/alphagov/govuk-infrastructure/pull/4651/changes?w=1)


samsimpson1 (approved):

jfharden:
Relates to https://github.com/alphagov/govuk-infrastructure/issues/4407

PR #4651, opened 2026-08-13
Convert all our gatekeeper rules to rego v1

jfharden:
NOTE: This PR has a huge amount of extra indentation, you should [view it with whitespace ignored](https://github.com/alphagov/govuk-infrastructure/pull/4651/changes?w=1)

* Convert all our existing gatekeeper rego rules to rego v1

rego v0 is long deprecated, all the docs are now for rego v1, and the rego playground defaults to v0, it's supported in gatekeeper, so lets convert while we have hardly any rules

Note if you are using regols as an LSP for rego, you should switch to regal which defaults to v1. I believe regols has no v1 support.


AP-Hunt (approved):
Yeah v1 is massively easier to comprehend. Approved.

PR #4664, opened 2026-08-18
[MAIN-7503] Add RDS instance for fact-check-manager in production

young-rob:
[Jira ticket](https://gov-uk.atlassian.net/jira/software/c/projects/MAIN/boards/1555?selectedIssue=MAIN-7503)

### Changes
Add an RDS instance for fact-check-manager in production and snapshot identifers to the existing staging and integration tfvars blocks.

### DIFF
--- a/terraform/variables/production/rds.tfvars
+++ b/terraform/variables/production/rds.tfvars
@@ -242,6 +242,28 @@ databases = {
     }
   }
 
+  fact_check_manager = {
+    engine         = "postgres"
+    engine_version = "17"
+    engine_params = {
+      log_min_duration_statement = { value = 10000 }
+      log_statement              = { value = "ddl" }
+      deadlock_timeout           = { value = 2500 }
+      log_lock_waits             = { value = 1 }
+    }
+    engine_params_family         = "postgres17"
+    name                         = "fact-check-manager"
+    allocated_storage            = 100
+    instance_class               = "db.t4g.medium"
+    performance_insights_enabled = true
+    project                      = "GOV.UK - Publishing"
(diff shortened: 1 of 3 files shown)


terraform/variables/production/rds.tfvars
```diff
@@ -242,6 +242,26 @@ databases = {
...
+      log_statement              = { value = "ddl" }
+      deadlock_timeout           = { value = 2500 }
+      log_lock_waits             = { value = 1 }
+    }
+    engine_params_family         = "postgres17"
+    name                         = "fact-check-manager"
+    allocated_storage            = 100
+    instance_class               = "db.t4g.medium"
+    performance_insights_enabled = true
+    project                      = "GOV.UK - Publishing"
```
nimalank7, line 259: Can you set a `maintenance_window` and `snapshot_identifier`?

nimalank7 (changes requested):

terraform/variables/production/rds.tfvars
young-rob: Thanks @nimalank7, I've added both of those now. Two questions:

We don't have them set for fact-check-manager in `staging/rds.tfvars` and `integration/rds.tfvars` - do we need to add them?

I've set `maintenance_window` to be the same as for link_checker_api - is this ok?

nimalank7: No need to set the `maintenance_window` for integration and staging but if you can set `snapshot_identifier` for integration and staging as well.

Yep maintenance window is fine to be the same as `link_checker_api`

nimalank7 (approved):
LGTM!

PR #4674, opened 2026-08-20
cloud run job timeout reduced

srinivasadaggu-dotcom:
1) The Cloud Run job timeout was reduced from two hours to 30 minutes to prevent long wait times, as the ingestion process usually completes in about 10 minutes
2) Audit table ddl schema included


aaronfowles (approved):
LGTM

PR #4676, opened 2026-08-21
Address permanent drift in search-api-v2

emmalowe:
Adjust boostAmount of 0
    
This is to avoid permanent drift in the Terraform configuration, which is caused by explicitly setting the boostAmount to 0. When the boostAmount is set to 0, it is removed from the response by the Discovery Engine API, causing a permanent drift.
    
One way to remove the drift would be to remove the set boostAmount so that our Terraform configuration matches what's in the API response. However, if we do that, we have no way of confirming what the boostAmount actually is, since it is not visible in the GCP console (UI), or the API response. To remove this uncertainty as well as the drift, we are setting the boostAmount to a near-zero value instead.

### DIFF
--- a/terraform/deployments/search-api-v2/serving_config_global_variant.tf
+++ b/terraform/deployments/search-api-v2/serving_config_global_variant.tf
@@ -44,8 +44,8 @@ module "control_global_boost_freshness_general" {
         interpolationType = "LINEAR",
         # Control points explained
         # 0.2 (decaying to 0.05), for the first 7 days
-        # 0.05 (decaying to 0), for 8 days to 90 days
-        # 0 (decaying to -0.5) for 91 days to 365 days
+        # 0.05 (decaying to ~0), for 8 days to 90 days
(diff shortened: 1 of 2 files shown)


terraform/deployments/search-api-v2/serving_config_global_variant.tf
```diff
@@ -58,7 +58,10 @@ module "control_global_boost_freshness_general" {
           },
           {
             attributeValue = "90D",
-            boostAmount    = 0.0
+            boostAmount    = 0.00000001
```
emmalowe, line 61: I'm not sure if this is also going to give us drift because the API response gives "1e-08" back as a response. If it does, I can also try setting it as "1e-08" (which the Google API accepts), or failing that, just go with the smallest value that doesn't get converted into scientific notation (0.0001).

AP-Hunt (approved):

stringiest (approved):

PR #4682, opened 2026-08-21
[SCH-2261] Add search v2 quota preferences

stringiest:
We've agreed certain quota overrides with Google via the console. Adding them to the terraform deployment to document the agreed quotas.

See [Jira ticket SCH-2261](https://gov-uk.atlassian.net/browse/SCH-2261)

N.B. Do not merge until PR #4737 has been merged and the terraform manually applied. Marking as draft for now.

### DIFF
--- a/terraform/deployments/search-api-v2/README.md
+++ b/terraform/deployments/search-api-v2/README.md
@@ -32,6 +32,10 @@ This module provisions the following resources into the Google Cloud Platform pr
 - AWS Secrets Manager secrets to be consumed by the API service application in the corresponding
   environment on the Kubernetes platform
 
+## Additional information
+### GCP quota preferences
+[Quota preferences][quota_preference] are used to request changes to quota values and to document changes already agreed with Google. Note that they cannot be used to alter system limits. See [Quotas and system limits][quotas-and-system-limits] for more information.
+
 > **Note**
 > The Discovery Engine resources are managed through the [RestAPI provider][restapi_provider_docs]
 > due to the Google provider not offering first party Terraform resources yet (as of October 2023).
@@ -52,3 +56,5 @@ This module provisions the following resources into the Google Cloud Platform pr
 [publishing-deployment]: ../govuk-publishing-infrastructure/README.md
 [variables]: ./variables.tf
 [variables-folder]: ../../variables
+[quota_preference]: https://registry.terraform.io/providers/hashicorp/google-beta/latest/docs/resources/cloud_quotas_quota_preference
+[quotas-and-system-limits]: https://docs.cloud.google.com/docs/quotas/quotas
--- a/terraform/deployments/search-api-v2/README.md
+++ b/terraform/deployments/search-api-v2/README.md
@@ -33,8 +33,41 @@ This module provisions the following resources into the Google Cloud Platform pr
   environment on the Kubernetes platform
 
 ## Additional information
-### GCP quota preferences
-[Quota preferences][quota_preference] are used to request changes to quota values and to document changes already agreed with Google. Note that they cannot be used to alter system limits. See [Quotas and system limits][quotas-and-system-limits] for more information.
+### Quotas and system limits
+Our usage of the various Google APIs is controlled through [quotas and system limits][quotas-and-system-limits]. Quota changes can be agreed with Google through [quota preferences](#quota-preferences) or capped using [quota overrides](#quota-overrides), while system limits are fixed. 
+
+Despite its fixed nature we have agreed an increase to the system limit `SearchRequestsPerMinutePerProjectPerRegion` on `production`. Quotas and system limits can be found in the GCP console under "IAM and Admin > Quotas and system limits".
+
+### Quota preferences
+[Quota preferences][quota_preference] are used to request changes to quota values and to document changes already agreed with Google. Note that they cannot be used to alter system limits.
+
+### Quota overrides
+Quota overrides are used to cap limits below values that have been set by default or by admin/producer overrides.
+On GCP, these are somewhat complex to set up and use inconsistent terminology between the
+console UI, the REST API, and the (beta) Terraform provider. In particular, it can be somewhat
+confusing to figure out the `limit` value for the `google_service_usage_consumer_quota_override`
+resource (which actually corresponds to the `unit` field in the API but with different syntax), and
+to find the internal (not display) name of quotas.
+
+If you need to set up a new `google_service_usage_consumer_quota_override` resource for a Discovery
+Engine project, the best way of finding out these values is to make a GET request to the
+`consumerQuotaMetrics` endpoint like so:
+
+```bash
+curl -H "Authorization: Bearer $(gcloud auth print-access-token)" \
+-H "Content-Type: application/json" \
+"https://serviceusage.googleapis.com/v1beta1/projects/${GCP_PROJECT}/services/discoveryengine.googleapis.com/consumerQuotaMetrics" \
+| jq -r '.metrics[] | "\(.displayName): \(.consumerQuotaLimits[0].metric) (\(.consumerQuotaLimits[0].unit | gsub("[1\\{\\}]";"")))"' \
+| sort
+```
+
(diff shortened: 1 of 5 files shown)


terraform/deployments/search-api-v2/cloud_quotas.tf
```diff
@@ -0,0 +1,30 @@
...
+  parent   = "projects/${var.gcp_project_id}"
+  name     = "discoveryengine_complete_query_requests"
+  service  = "discoveryengine.googleapis.com"
+  quota_id = "CompleteQueryRequestsPerMinutePerProject"
+  quota_config {
+    preferred_value = var.complete_query_requests
+  }
+}
+
+resource "google_cloud_quotas_quota_preference" "search_requests_regional_preference" {
```
emmalowe: I wasn't sure if this would work because it's a system limit. Did you try it on integration?

terraform/deployments/search-api-v2/variables.tf
```diff
@@ -50,3 +50,9 @@ variable "evaluation_create_requests" {
   description = "The maximum number of evaluation create requests per day for Discovery Engine"
   default     = 100
 }
+
+variable "complete_query_requests" {
+  type        = number
+  description = "The maximum number of complete query requests per minute for Discovery Engine"
+  default     = 300
```
emmalowe, line 57: I think it's worth mentioning in the commit message that this is the current default value in GCP.

terraform/deployments/search-api-v2/variables.tf
```diff
@@ -56,3 +56,9 @@ variable "complete_query_requests" {
   description = "The maximum number of complete query requests per minute for Discovery Engine"
   default     = 300
 }
+
+variable "search_requests_regional" {
+  type        = number
+  description = "The maximum number of search requests per minute per region for Discovery Engine"
+  default     = 300
```
emmalowe: I'm a bit confused by our current settings for this. It looks like the default in GCP is 300, but for individual regions it's 100? Do you know what's going on here?

terraform/deployments/search-api-v2/cloud_quotas.tf
```diff
@@ -17,3 +17,14 @@ resource "google_cloud_quotas_quota_preference" "complete_query_requests_prefere
     preferred_value = var.complete_query_requests
   }
 }
+
+resource "google_cloud_quotas_quota_preference" "search_requests_regional_preference" {
+  parent     = "projects/${var.gcp_project_id}"
+  name       = "discoveryengine_search_requests_regional"
+  dimensions = { region = "global" }
```
emmalowe: Looking in the GCP console, it looks like we have a higher limit set for eu and us regions in production too. Since we're not using those regions (as far as I know), I don't think we necessarily need that, but I was wondering if that was a conscious decision?

emmalowe:
The structure of this (and quota IDs) look good to me - I just had some questions on the details!

terraform/deployments/search-api-v2/cloud_quotas.tf
stringiest: Just tried on integration and it did not work. I'd hoped that it would as it was returned in the list of QuotaInfos. For our future reference the Bad RequestException message said `Could not change quota value because 'is_fixed' field is set for quota id 'SearchRequestsPerMinutePerProjectPerRegion'.` I'll remove this commit.

terraform/deployments/search-api-v2/variables.tf
stringiest: This will be moot as we can't use quota preferences to set system limits.

terraform/deployments/search-api-v2/cloud_quotas.tf
stringiest: It wasn't a conscious decision, no. This is also moot as we can't use quota preferences to set system limits.

terraform/deployments/search-api-v2/variables.tf
stringiest: Good idea! Now done :)

terraform/deployments/search-api-v2/cloud_quotas.tf
emmalowe: Oh that's a shame. I wondered if it's worth adding that as a comment here, or documenting it in some way? I'm also not sure how we would have increased that previously - can we send a request in the UI for that?

stringiest: I'm not sure how we *did* manage to increase a system limit. Everything in the docs says that it's not possible! It doesn't show in the requested increases tab. 

I've added a paragraph to the README to say that system limits can't be changed using `quota_preferences`, only quotas.

terraform/deployments/search-api-v2/README.md
```diff
@@ -32,11 +32,13 @@ This module provisions the following resources into the Google Cloud Platform pr
 - AWS Secrets Manager secrets to be consumed by the API service application in the corresponding
   environment on the Kubernetes platform
 
+## Additional information
+### GCP quota preferences
+[Quota preferences][quota_preference] are used to request changes to quota values and to document changes already agreed with Google. Note that they cannot be used to alter system limits. See [Quotas and system limits][quotas-and-system-limits] for more information.
```
emmalowe: I wonder if it's worth mentioning specifically that we have an override for regional search requests per minute.

terraform/deployments/search-api-v2/README.md
```diff
@@ -32,11 +32,13 @@ This module provisions the following resources into the Google Cloud Platform pr
 - AWS Secrets Manager secrets to be consumed by the API service application in the corresponding
   environment on the Kubernetes platform
 
+## Additional information
```
emmalowe, line 35: I think we should move the info on quota limit overrides from the [gcp-search-api-v2 README](https://github.com/alphagov/govuk-infrastructure/tree/main/terraform/deployments/gcp-search-api-v2) to here. It would make sense for them to be together and if we were to apply those again in the future, I would put them in this folder.

emmalowe (commented):
Infra changes look good to me - just a couple of comments on the documentation.

terraform/deployments/search-api-v2/README.md
stringiest: I've added a comment to explain that we've got an increase for the limit, signposting to GCP to see the limit rather than including it here.

terraform/deployments/search-api-v2/README.md
stringiest: Now consolidated.

terraform/deployments/search-api-v2/README.md
```diff
@@ -33,8 +33,41 @@ This module provisions the following resources into the Google Cloud Platform pr
   environment on the Kubernetes platform
 
 ## Additional information
-### GCP quota preferences
-[Quota preferences][quota_preference] are used to request changes to quota values and to document changes already agreed with Google. Note that they cannot be used to alter system limits. See [Quotas and system limits][quotas-and-system-limits] for more information.
+### Quotas and system limits
+Our usage of the various Google APIs is controlled through [quotas and system limits][quotas-and-system-limits]. Quota changes can be agreed with Google through [quota preferences](#quota-preferences) or capped using [quota overrides](#quota-overrides), while system limits are fixed. 
+
+Despite its fixed nature we have agreed an increase to the system limit `SearchRequestsPerMinutePerProjectPerRegion` on `production`. Quotas and system limits can be found in the GCP console under "IAM and Admin > Quotas and system limits".
```
emmalowe, line 39: This should be "APIs and services > Quotas and system limits"

stringiest: Interesting! Are you sure? I can only access it via IAM and Admin - it doesn't show in the list of options under APIs and services for me.

emmalowe: We discovered you can use both routes! But @stringiest's is more direct, so we'll keep that here!

emmalowe (approved):

AP-Hunt (approved):

PR #4689, opened 2026-08-24
setup cronjob only for production

srinivasadaggu-dotcom:
Setup cronjob only for production


srinivasadaggu-dotcom:
After discussing with the team , it was agreed that this code change is not required, closing the PR.

PR #4690, opened 2026-08-24
Add govuk-user-reviewer to sentry projects

jfharden:
Import govuk-user-reviewer project which was clickopsed


jfharden:
I jumped the gun wanting to bring this into terraform, the terraform is to configure projects for continually running apps managed by GOV.UK Developers, so this isn't required

PR #4703, opened 2026-08-26
Only build arm64 images on PRs

jfharden:
All PR builds are broken for a lot of projects because we no longer have amd64 base images, it is intentional not to have them anymore, but on PRs we are still trying to build for both architectures.

This PR takes a big bang approach and changes the existing workflow so we unblock everyone, we can always restore this from git history if we need to.

### DIFF
--- a/.github/workflows/build-image-from-pr.yml
+++ b/.github/workflows/build-image-from-pr.yml
@@ -26,6 +26,8 @@ on:
         required: false
         type: string
         default: ubuntu-24.04-arm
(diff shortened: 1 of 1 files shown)


nimalank7 (approved):
LGTM!

jfharden:
[This PR in hmrc-manuals-api](https://github.com/alphagov/hmrc-manuals-api/pull/1347) was opened pointing at this branch to test this changed workflow, it successfully built the image on the PR

PR #4704, opened 2026-08-26
Add MySQL databases for Specialist Publisher [WHIT-3578]

ChrisBAshton:
Specialist Publisher is moving from Mongo (shared DocumentDB instance) to MySQL. The vast majority of its data actually lives in Publishing API: Specialist Publisher only requires a database in the first place for Signon integration. The gds-sso gem mandates that there is a database with a Users table.

There are around 200-300 users that visit Specialist Publisher every month, across around 1k unique sessions. The amount of data actually stored is literally bytes.

I've opted for the smallest possible MySQL instance and floor of 20GB storage.

MySQL was chosen over Postgres because our long term plans are to consolidate Specialist Publisher into Whitehall, which itself uses MySQL.

Jira: https://gov-uk.atlassian.net/browse/WHIT-3578

### DIFF
--- a/terraform/variables/integration/rds.tfvars
+++ b/terraform/variables/integration/rds.tfvars
@@ -515,6 +515,25 @@ databases = {
     }
   }
 
+  specialist_publisher = {
+    engine                      = "mysql"
+    engine_version              = "8.4"
+    allow_major_version_upgrade = true
+    engine_params = {
+      max_allowed_packet = { value = 1073741824 }
+    }
+    engine_params_family         = "mysql8.4"
+    name                         = "specialist-publisher"
+    allocated_storage            = 20
+    instance_class               = "db.t4g.micro"
+    performance_insights_enabled = true
+    project                      = "GOV.UK - Publishing"
+    snapshot_identifier          = "specialist-publisher-mysql-post-encryption"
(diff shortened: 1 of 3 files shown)


terraform/variables/integration/rds.tfvars
```diff
@@ -515,6 +515,25 @@ databases = {
...
+    engine_version              = "8.4"
+    allow_major_version_upgrade = true
+    engine_params = {
+      max_allowed_packet = { value = 1073741824 }
+    }
+    engine_params_family         = "mysql8.4"
+    name                         = "specialist-publisher"
+    allocated_storage            = 20
+    instance_class               = "db.t4g.micro"
+    performance_insights_enabled = true
```
jfharden, line 529: It's worth checking in on the DB to make sure it's ok cpu & memory wise, performance insights runs extra process(es) in the RDS instances which consume some memory and cpu, just with the instance being so small it's worth keeping an eye on it for a little bit

jfharden (approved):

terraform/variables/integration/rds.tfvars
ChrisBAshton: Now disabled in #4706.

PR #4729, opened 2026-08-27
Adjust boostAmount close to zero

emmalowe:
Adjusting the freshness boostAmount at 90 days to the smallest number we can before Terraform and the Google API start converting it into scientific notation, which also causes permanent drift (potentially because Terraform and
Google have a slightly different way of writing scientific notation).

I have tried working with the scientific notation by specifying the boostAmount as the following values (none of which worked):

- 1e-8
- 1e-08
- format("%e", 0.00000001)
- format("%#v", 0.00000001)
- "1e-8"


stringiest (approved):
🚀

AP-Hunt (approved):

Issue #4063, opened 2026-04-10
Remove cruft from Logstash configuration


nimalank7:
**User Need**

**As a** Platform Engineer
**I want** to remove unnecessary configuration from our logstash configuration
**so that** our logstash configuration is readable and up to date

---

### Context

Our Logstash configuration hasn't been touched in a while and contains unnecessary configuration (e.g. IIS)

---

### What’s Needed

List anything the solution must do or be (behaviour, performance, security, UX, etc.).

- [x] Take a copy of the existing configuration and replacing the existing on in Git
- [x] Start with the version in staging, because it's been formatted correctly and is readable
- [x] Remove any unnecessary configuration from Logstash

---

### Acceptance Criteria

Clear, measurable conditions to verify the story is complete and valuable.

- [x] Logstash continues to work as before
- [x] The Logstash configuration file is smaller and easier to understand

---

nimalank7:
Currently all environments are using the same logstash pipeline. Remaining PRs:

- https://github.com/alphagov/govuk-infrastructure/pull/4730
- https://github.com/alphagov/govuk-infrastructure/pull/4731

Once this is merged in and applied to Logit I'll write up some docs on the Logstash pipeline.

```
# If error is a string and not an object (fixes mirror?)
  if [error] and ![error][message] and ![error][type] {
    mutate {
      rename => { "error" => "error_message" }
    }
  }
```

This is the only part I'm unsure whether Filebeat fills this out and if it's necessary.

nimalank7:
PR is now merged in and a README is created.

PR #4730, opened 2026-08-27
Refactor `logjson` processing in Logstash

nimalank7:
Description:
- https://github.com/alphagov/govuk_sidekiq/blob/c5d2466558fa0e2cad97e5833dd0016f2ad97578/lib/govuk_sidekiq/govuk_json_formatter.rb#L7 doesn't expose `[ts]` so this part can be removed
- Message is already parsed into JSON so `if [message] and [message][args]` can be removed
- `Extract human readable message` is covered by `if [logjosn][message]` so that seciton can be removed
- `[logjson][jid]` doesn't already exist in `logjson_message` so the `remove_field` section can be removed
- Filebeat doesn't annotate `useragent` so the field has to be extracted from nginx logs. This is why the `useragent` filter has to occur after `logjson` transformation happens
- Use [ecs_compatibility](https://www.elastic.co/docs/reference/logstash/plugins/plugins-filters-useragent#plugins-filters-useragent-ecs_compatibility) for `useragent` filter instead of manually creating fields. It will automatically populate and add the `useragent` field
- https://github.com/alphagov/govuk-infrastructure/issues/4063

### DIFF
--- a/logstash/logstash.conf
+++ b/logstash/logstash.conf
@@ -5,167 +5,43 @@ filter {
       target => "logjson"
       skip_on_invalid_json => true
     }
-  }
 
-  useragent {
-    source => "[user_agent][original]"
-    target => "ua_tmp"
-    add_field => {
-      "[user_agent][device][name]" => "%{[ua_tmp][device]}"
-      "[user_agent][os][name]" => "%{[ua_tmp][os_name]}"
+    useragent {
+      source => "[logjson][http_user_agent]"
+      ecs_compatibility => "v8"
     }
-  }
 
-  # OS version ECS compatibility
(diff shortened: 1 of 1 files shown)


AP-Hunt:
I think all of the mutations that follow from the existence of `logjson` should sit inside the same block the JSON parsing is happening in. `logjson` doesn't exist without it.

nimalank7:
> I think all of the mutations that follow from the existence of `logjson` should sit inside the same block the JSON parsing is happening in. `logjson` doesn't exist without it.

Cool I've refactored it inside that block now

AP-Hunt (approved):

PR #4790, opened 2026-09-09
Remove temporary MySQL 8.4 upgrade PGs

samsimpson1:
These were left over from the MySQL 8.4 upgrade work, and are no longer in use


nimalank7 (approved):
LGTM!

PR #4805, opened 2026-09-11
feat: 🎸 github custom property add team_other if val not in existing valid values list

jaskaransarkaria:
We set the "other" field for EE's custom property on a github repo when a value we want to set the custom property to is not in the existing valid values list set at the organisation level by EE.

- set the "team_other" custom property to the new vustom property value (team name / slack channel) when the "other" field is set

### DIFF
--- a/terraform/deployments/github/custom_properties.tf
+++ b/terraform/deployments/github/custom_properties.tf
@@ -20,3 +20,14 @@ resource "github_repository_custom_property" "team_identifier" {
   property_value = [contains(data.github_organization_custom_properties.team_identifier.allowed_values, each.value) ? each.value : "other"]
 }
 
+resource "github_repository_custom_property" "team_other" {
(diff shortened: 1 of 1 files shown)


terraform/deployments/github/custom_properties.tf
```diff
@@ -3,6 +3,12 @@ data "github_organization_custom_properties" "team_identifier" {
   value_type    = "single_select"
 }
 
+
+data "github_organization_custom_properties" "team_other" {
```
jfharden: Not sure you need to do the data lookup for this, it's just free text and the only outputs you're using from this are the 2 params you're setting below.

Doesn't really matter though, it's only 1 api request of ONE BILLION 😆

jfharden (dismissed):

jfharden (approved):

PR #4809, opened 2026-09-14
Decommission green es clusters and clean up

jfharden:
* Deccomission the elasticsearch green clusters
* Had to do some state moves to deal with the s3 buckets, but you can see this is safe since ai-accelerator is not deleting its S3 bucket, and search-opensearch is only deleting the old one (govuk-<env>-elasticsearch6-manual-snapshots) which is confirmed ok to do by @koetsier 

The following things are being deleted as part of tearing green clusters down:

* `aws_route53_record.service_record`: The old CNAME which isn't in use anymore (and hasn't been for a long time before this)
* `module.opensearch.aws_iam_role.elasticsearch_snapshot[0]`: IAM role `arn:aws:iam::<account_id>:role/green-elasticsearch6-manual-snapshot-role` This was the IAM role for the old clusters, has not been used since it was recreated on September 8 during [this revert PR](https://github.com/alphagov/govuk-infrastructure/pull/4778), the new one should be using the alternate role govuk-integration-search-domain-opensearch-snapshot, in this case that role was last used 5 hours ago suggesting it is the in-use role, and the 
* `module.opensearch.aws_opensearch_authorize_vpc_endpoint_access.green[0]`: The vpc endpoint which was part of the madness of trying to get the remote reimport to work
* The old security group rules (many of them) which were applied to the old security group, the old security group is not even applied to the new blue clusters, so we know that isn't in use
* All the old cloudwatch log groups and the log resource policy
* The old snapshot s3 buckets


jfharden:
Once applied I will do a follow on PR to rename the `elasticsearch.tf` file in the `search-opensearch` deployment to `opensearch.tf`, and remove the imports

AP-Hunt:
The S3 buckets that are being deleted  have `force_destroy = false`. Have you cleared them out already?

jfharden:
@AP-Hunt no, I'll do it when I'm about to apply. My intention is to merge, then clear the s3 bucket in integration, and thenapply in integration and make sure things are all happy there before I do other envs (It will mean we can't apply staging and prod, but the deployments are scope limited to the open search clusters so I think that's fine for a few hours)

AP-Hunt:
Ok, happy with that plan.

AP-Hunt (approved):

Issue #4116, opened 2026-04-23
Create runbook for upgrading PostgreSQL 14 to 18


AP-Hunt:
**User Need**

**As a** platform engineer
**I want** to develop a runbook for upgrading a PostgreSQL 14 database to version 18
**so that** development teams can follow the runbook and upgrade their databases in a consistent way

---

### Context

We did an upgrade from PostgreSQL 13 to 14 in the summer of last year. You may be able to start from [the runbook we created](https://docs.google.com/document/d/1unUfeeoW_AellZFtrmjO2NIhbE4dWuXdO8djytxc_N8/edit?tab=t.0#heading=h.735gxihe2frj) back then.

In the MySQL run book we created, we made some useful tweaks to the process that we can mirror in this one.

---

### What’s Needed

List anything the solution must do or be (behaviour, performance, security, UX, etc.).

- [ ] Test performing an upgrade to see if we can even go directly from 14 to 18
- [ ] Work out how to upgrade `pact-broker` in Heroku
- [ ] Write up a runbook
- [ ] Test the runbook by performing a test upgrade again

---

### Acceptance Criteria

Clear, measurable conditions to verify the story is complete and valuable.

- [ ] The runbook I've written results in as little downtime as possible
- [ ] I've tested my runbook and I'm confident that it is, at least, mostly right

nicholsj:
NB there are two teams outside GOV.UK impacted by this: Chat and DGU (CKAN database). I've messaged both as well as posting to GOV.UK DMs/PMs and TLs

nimalank7:
https://github.com/alphagov/govuk-infrastructure/pull/4829 to test out staging `support-api`

PR #4810, opened 2026-09-14
Set apply_method correctly to ckan_sams_test RDS PG params

samsimpson1:
Default apply_method is immediate, which is not supported by these params

https://github.com/alphagov/govuk-infrastructure/issues/4116


nimalank7 (approved):
LGTM!

PR #4827, opened 2026-09-17
Alter s3 bucket replication configuration

jfharden:
We need to support rule priorities in order to import the replication rules for the assets bucket

In this PR

* Remove old imports
* Add rule priority support to the s3 bucket module 
* Bring the db_backup_s3.tf bucket replication rules into the module instead of outside


terraform/deployments/govuk-publishing-infrastructure/db_backup_s3.tf
```diff
@@ -64,6 +64,25 @@ module "secure_s3_bucket_db_backup_main" {
...
+    rules = [{
+      id                        = "replicate-db-backups-out-of-region"
+      priority                  = 10
+      status                    = var.govuk_environment == "production" ? "Enabled" : "Disabled"
+      delete_marker_replication = { status = "Disabled" }
+      destination = {
+        bucket        = module.secure_s3_bucket_db_backup_replica.arn
+        storage_class = "STANDARD_IA"
+      }
+    }]
```
jaskaransarkaria, line 79: In the deleted old config there's an empty `filter {}` underneath this rule, do we still need it? Or is it covered by the new dynamic block?

jaskaransarkaria (approved):

terraform/deployments/govuk-publishing-infrastructure/db_backup_s3.tf
jfharden: It's covered by the new dynamic block

PR #4831, opened 2026-09-17
Import s3 replication rules for assets bucket

jfharden:
Turns out our replication config for the assets bucket in prod is not in terraform. This is a first pass at importing it.

I've changed the type of the replication config rules from a set to a list so we can maintain the order of the rules when we import them

There is a small change as a result of this, but it's just to rename the rules from `prod-....` to `production-....`


samsimpson1 (approved):

terraform/deployments/govuk-publishing-infrastructure/asset_manager_s3.tf
```diff
@@ -25,6 +54,41 @@ module "assets" {
       },
     ]
   }
+
+  # There is a whole rule missing here, we also have a rule prod-assets-to-prod-backup
+  # but that uses a different role and terraform doesn't support setting different roles
+  # for different rules. We will need to consolidate the roles into one and then import it
+  replication_config = length(var.replicate_assets_to_environments) == 0 ? null : {
+    role = data.aws_iam_role.assets_s3_replication[0].arn
```
jaskaransarkaria, line 62: If we aren't in "production" will this throw an error?

terraform/shared-modules/govuk-aws-accounts/README.md
```diff
@@ -0,0 +1,47 @@
+# GOV.UK AWS Accounts
+
+A simple module which provides easy lookup of account names and IDs
```
jaskaransarkaria, line 3: 😍

terraform/deployments/govuk-publishing-infrastructure/asset_manager_s3.tf
jfharden: No, because all of the other envs don't have replicate_assets_to_environments set

PR #4839, opened 2026-09-18
add publishing taxononmy service account as a publishing api member.

AlexFaria93:
Add new service account, taxonomy-pipeline@govuk-publishing.iam.gserviceaccount.com to as viewer member for publishing-api.


.DS_Store
aaronfowles, line 1: @AlexFaria93 would you mind removing this file from the commit and then I'll approve. Thanks!

aaronfowles (changes requested):

.DS_Store
AlexFaria93: File deleted

aaronfowles (approved):

PR #4850, opened 2026-09-22
BQ table partition included

srinivasadaggu-dotcom:
As part of the feedback processing pipeline replacement with DataForm , BigQuery table partition is included for better query performance.


aaronfowles (approved):
LGTM. There was a a deployment issue with this table this morning but the plan looks OK now. We can merge this and then see if we need to take further action in a follow-up PR. I suspect we may need to reimport the resource but let's see...

